PR #575, opened 2014-09-21
Better circular dependency handling

kendallb:
Better circular dependency handling so property injection ALWAYS works with circular dependencies and property injected classes are always properly injected prior to being passed to the constructor of a constructor injection. Hence you can have classes that use constructor injection that can use any instances with circular dependencies passed to them as long as they are using property injection.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -153,3 +153,4 @@ $RECYCLE.BIN/
 
 # Mac crap
 .DS_Store
+copy_debug.bat
--- a/Autofac.sln
+++ b/Autofac.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 2013
-VisualStudioVersion = 12.0.30501.0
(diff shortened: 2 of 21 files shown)


tillig:
Given this has been open a while and a lot has changed with the .NET Core support and v4.0, I'm going to close it and we will use the great work here as a jumping-off point for potential improvements to the updated code base.

I've started an issue at #798 to get some discussion and ideas going around ways to improve circular dependency.

kendallb:
Ok, it is extremely annoying that my changes got ignored for an entire decade, especially because this was never actually addresses and fixed. We just got bitten by this same bug as we changed to using official Autofac some time ago and I didn't do enough testing to see if this was resolved, and it is not.

I have started a new tree and resurrected my original test and it crashes the same way it used to a decade ago. I will take a look at how Autofac is structured now to better handle circular dependencies when using property injection as clearly none of the changes over the last decade have actually addressed this. Hopefully I can figure it out and we will have to go back to building and publishing out own version of Autofac again until this can be resolved upstream.

Issue #477, opened 2014-01-22
Circular dependency resolution does not fully resolve classes passed to constructor injection


alexmg:
_From [kendallb...@gmail.com](https://code.google.com/u/108912633848931461930/) on December 07, 2013 13:44:15_

I have noticed an anomaly in the way that Autofac handlers wiring of property injection compared to how I was used it being done in Ninject. For a lot of our stuff we use property injection because of circular references, as there is no way to resolve those without using property injection (I have tried!). So as a rule most of our primary data access, business logic and service classes use property injection, and they need to be wired up such that dependency can be circular. Now sometimes we have classes that need to be constructed on the fly, and in Ninjet we used the factory extension, but for Autofac it is conveniently built in using Func<IMyInterace> members. A lot of those classes were originally designed to only use constructor injection and relied on the interfaces injected into it being wired up when the constructor is called. However with Autofac that is NOT the case if those interfaces were wired internally with property injection.

Basically what would happen is I would have a constructor like this:

public MyClass(string arg1, int arg2, ISomeDepdency dep1, IAnothterDep dep2)
{
    // Do something with dep1 and dep2
}

But if dep1 and dep2 are set up to use property injection (in my case also using circular dependencies), dep1 and dep2 may not be actually wired up yet, so if I make calls in the constructor to dep1 or dep2 methods that internally use property injected dependencies, they will crash. This appears to happen if MyClass is the first class to consume dep1 and dep2 in the current scope, so they had to be created as part of creating the instance of MyClass. If they already existed, then it works correctly.

This is clearly a leaky abstraction layer because now MyClass has to care about HOW dep1 and dep2 were wired up, and hence I cannot do anything useful in the constructor. All I can do is save off all the parameters and then move the business logic that would have occurred in the constructor into a Load() function that is called whenever internal methods or properties are accessed.

In reality Autofac should be intelligent enough to know that it needs to COMPLETELY resolve dep1 and dep2 before passing them to MyClass. Unless there is a circular dependency to be found, this should be completely doable.

To illustrate the issue I have built a project that demonstrates it clearly, so you can compile and run it to see the problem.

**Attachment:** [CircularCrashTest.zip](http://code.google.com/p/autofac/issues/detail?id=477)

_Original issue: http://code.google.com/p/autofac/issues/detail?id=477_

tillig:
This may or may not tie into the newly opened design-based issue #798.

tillig:
[...]

PR #579, opened 2014-09-29
Bug in CircularDependency tracking

Yshayy:
Error in InstanceLookup constructor may result in Exception when attempting to find scope, such an exception cause inconsistent _callDepth value. 
This can cause a CircularDependency exception by reaching the threshold of the max depth.

[Review on Reviewable](https://reviewable.io/reviews/autofac/autofac/579)


tillig:
I believe we're still looking for a unit test on this one before we can pull it in, yes?

tillig:
This has been sitting for a while and is now not able to be taken due to changes in the branching structure. We also are still unclear on what specific use case this fixes since there is no test.

I'm going to link this to issue #477 as a place to improve circular dependency handling.

tillig:
Correction: follow this on #798.

PR #602, opened 2014-11-28
Add DisposeContainerOnAppDisposing method to OwinExtensions.

### LINKED ISSUE #565: Should UseAutofacMiddleware dispose container on OWIN shutdown?
I currently have this method in my OWIN Startup class:

```c#
protected virtual void RegisterShutdownCallback(IAppBuilder app, IContainer container)
{
    var context = new OwinContext(app.Properties);
    var token = context.Get<CancellationToken>("host.OnAppDisposing");

    if (token != CancellationToken.None)
    {
        token.Register(container.Dispose);
    }
}
```

My question is should UseAutofacMiddleware do this for me?

On the one hand, the `IContainer` instance is supplied externally, so there may be cases where the container should not be disposed simply because the OWIN app is shutting down.

On the other hand, the default behavior means that containers won't be disposed properly when a self-hosted app or IIS app shuts down. This can cause problems for containers that have services that need to clean up before the process exits.

chriseldredge:
Fixes #565 on branch develop.

### DIFF
--- a/Core/Source/Autofac.Integration.Owin/AutofacAppBuilderExtensions.cs
+++ b/Core/Source/Autofac.Integration.Owin/AutofacAppBuilderExtensions.cs
@@ -27,6 +27,7 @@ using System;
 using System.ComponentModel;
 using System.Linq;
 using System.Security;
+using System.Threading;
 using Autofac;
 using Autofac.Core;
 using Autofac.Core.Lifetime;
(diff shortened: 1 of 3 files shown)


chriseldredge:
Replaced by https://github.com/autofac/Autofac.WebApi.Owin/pull/2

Issue #272, opened 2014-01-22
PreserveExistingDefaults doesn't work in nested lifetime scopes


alexmg:
_From [mrdont@mail.ru](https://code.google.com/u/mrdont@mail.ru/) on November 19, 2010 12:06:33_

What steps will reproduce the problem? 1. Register one implementation of a service in root container.
2. Begin new lifetime scope and register another implementation of the service, but preserving existing defaults.
3. Resolve service in the created scope.

What is the expected output? First implementation of the service.
What do you see instead? Another implementation of the service.

Service registered in root container is also default implementation in nested scope, until another implementation is registered. But when I specify PreserveExistingDefaults, I expect that new registration won't override existing registrations of the same service.

The following code would reproduce the problem:

        public interface IService {}
        public class Service1: IService {}
        public class Service2: IService {} 

        [Test]
        public void PreserveExternalOverrides()
        {
            var builder = new ContainerBuilder();
            builder.RegisterType<Service1>().As<IService>();
            var context = builder.Build();

            var newContext = context.BeginLifetimeScope(b => b.RegisterType<Service2>().As<IService>().PreserveExistingDefaults());

            Assert.IsInstanceOf<Service1>(context.Resolve<IService>());
            Assert.IsInstanceOf<Service1>(newContext.Resolve<IService>()); // fails

        }

_Original issue: http://code.google.com/p/autofac/issues/detail?id=272_

tillig:
NICE! This had been open for almost six years!

alexmg:
I manually merged across the PR from @ilya-g as the starting point so props to them. :+1:

PR #606, opened 2014-12-17
Preserve correct order of service registrations in the nested component registry

ilya-g:
This pull request provides the solution for issue #272.

**Requirements**
The key to resolve this issue is to ensure the correct order of the service implementations in `ServiceRegistrationInfo` class, since the first one of them is treated as default one.
To figure out which order is correct I was using the following assumptions:
- *default* registration always override all registrations made before, including those from parent scope
- *preserve-default* registration is always overridden by registrations made before. It's used as default only if there are no registrations made before.

So the correct order is one where in the beginning there are default explicit registrations made in nested scope, then registrations coming from sources, such as parent registry, and then in the end explicit preserve-default registrations.
These implementations in the middle from the parent registries must also be in the correct order, so that in case if the nested scope doesn't override the default implementation, the default one must be same as in the parent registry. The latter requirement was hard to implement due to the way parent registries were queried for registrations. 
A parent registry is added as the `ExternalRegistrySource` to the nested one. The source takes all non-adapting registrations from the registry and wraps them to coerce some of their properties. But this means that on the next level of hierarchy these wrapped registrations are excluded as adapting-ones when being queried by `ExternalRegistrySource` (probably, to prevent double wrapping). Therefore to get the all service implementations you have to traverse up the whole hierarchy and query every parent source for registrations. It becomes very hard to reconstruct the correct order of implementations when adding each provided implementation one by one.

**Changes**
So I've decided to change the way `ExternalRegistrySource` works. Now it doesn't exclude implementations wrapped by the parent registry. To distinguish these wrapped implementation from the adapting ones I've subtyped `ComponentRegistration` class. This also prevents double registration wrapping, 'cause such subtyped `ExternalComponentRegistartion` class can be shared between all the nested registries down to hierarchy.

**Questions**
One thing left unclear to me is the relation between *preserve-defaults* registrations and the registrations originated from sources. There are two mutually exclusive options:
1. Preserve-default registrations are always superseded by the other registrations, even those from sources.
2. The sources are consulted for registration only if there is no any explicit registration made to the registry, including preserve-default one. This means preserve-default registration takes precedence over that which was originated from registration source.
Can someone provide the answer (would be better if it's backed up with the reference to some existing docs)?

In the latter case we'll need to distinguish somehow `ExternalRegistrySource` from the other source types when initializing `ServiceRegistrationInfo`.

**TODOs**
I've introduced a new requirement for `IComponentRegistry.RegistrationsFor` method implementation. It must return the sequence, starting with the default implementation of the service. This requirement was fulfilled de-facto, but I think some new unit tests or assertions in existing tests should be made to ensure the default implementation (`ComponentRegistry`) behavior is correct.

### DIFF
--- a/Core/Source/Autofac/Core/IComponentRegistry.cs
+++ b/Core/Source/Autofac/Core/IComponentRegistry.cs
@@ -73,7 +73,10 @@ namespace Autofac.Core
         /// have been invoked.
         /// </summary>
         /// <param name="service">The service for which registrations are sought.</param>
-        /// <returns>Registrations supporting <paramref name="service"/>.</returns>
+        /// <returns>
+        ///  Registrations supporting <paramref name="service"/>.
+        ///  Default registration for the service must be first in the returned sequence.
+        /// </returns>
         IEnumerable<IComponentRegistration> RegistrationsFor(Service service);
 
         /// <summary>
--- a/Core/Source/Autofac/Core/Lifetime/LifetimeScope.cs
+++ b/Core/Source/Autofac/Core/Lifetime/LifetimeScope.cs
@@ -194,13 +194,18 @@ namespace Autofac.Core.Lifetime
                 .Where(src => src.IsAdapterForIndividualComponents))
                 builder.RegisterSource(source);
 
-            var parents = Traverse.Across<ISharingLifetimeScope>(this, s => s.ParentLifetimeScope)
+
+            // Issue #272: Only the most nested parent registry with HasLocalComponents is registered as an external source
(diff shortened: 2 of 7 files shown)


alexmg:
I have manually merged this PR in as part of #272 as the starting point for additional ordering work. Just wanted to let you know that your efforts have been valuable and appreciated, even if it did take a long time for us to get around to tackling this issue. Unfortunately, we have a whole bunch of new and shitty requirements layered on top of this thanks to the .NET Core conforming container specification, but this is a good starting point to make sure the preserve existing defaults continues to work in nested lifetime scopes while I refactor further.

PR #613, opened 2015-01-12
update autofac version to 3.5.2

hoffmanic:
We need this as the masstransit autofac integration has updated to this version of autofac. It appears there is no nuget package for this. can you please kindly use this to update the nuget packages?


alexmg:
The latest package on NuGet is 3.5.2. It was released Tuesday, August 12 2014 and has 87498 downloads.

https://www.nuget.org/packages/autofac

Is the official NuGet repository missing from your configured package sources?

hoffmanic:
Hi Alex,

The packages I would like updated are autofac.web, autofac.mvc5, autofac.wcf and autofac.configuration there are a heap more, but they are the ones we use specifically.

These dependencies use nuget to reference autofac itself. what happened? Autofac had an assembly version change which broke the strongname from 3.4 to 3.5, now when use use one of the above with autofac 3.5.2 we get an assembly not found exception Autofac, Version=3.4.0.0, Culture=neutral, PublicKeyToken=17863af14b0044da

We could add an assembly binding redirect to solve this problem, but to me this is hiding a problem that does not need to exist as all of these dependencies come from the same solution.

Can you please accept the pull request and rev the dependent packages?

tillig:
Have you tried adding an assembly redirect to your web.config/app.config file to redirect requests for the older assembly to the new version? [This is pretty standard stuff when dealing with new assembly versions and strong naming.](http://msdn.microsoft.com/en-us/library/7wd6ex19%28v=vs.110%29.aspx) Usually NuGet puts that in for you, but sometimes misses it.

```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
    <runtime>
        <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
            <dependentAssembly>
                <assemblyIdentity name="Autofac"
                                  publicKeyToken="17863af14b0044da"
                                  culture="neutral" />
                <bindingRedirect oldVersion="0.0.0.0-3.5.0.0"
                                 newVersion="3.5.0.0" />
            </dependentAssembly>
        </assemblyBinding>
    </runtime>
</configuration>
```

Internally we reference Autofac via NuGet to ensure we are maintaining backwards compatibility with older versions where possible. The assembly binding redirect takes care of things for projects that choose to upgrade core Autofac and doesn't require us to issue new integration packages tied to every time.

hoffmanic:
"We could add an assembly binding redirect to solve this problem, but to me this is hiding a problem that does not need to exist as all of these dependencies come from the same solution."

I did say that I could add this, again this leads to other issues down the road especially if there small breaking changes in the API, which is probably why you rev'd the assembly version.

The change I have submitted a pull request for, revs the autofac version for you, all you should have to do is accept the pull request and let your CI do the magic.

Accepting the request will really help us out, and does not affect anybody else, so it's a win/win at the end of the day.

tillig:
Oops, sorry. Missed that somehow. **#readfail on my part.** :)

The .NET standard way of handling this is through binding redirects. From a semantic versioning perspective, updating all of these things simply to take a new version of core Autofac means we issue 0.0.1 increments that people will be inclined to take (because they'll assume something is fixed in the release) and will indirectly force them to also take the latest Autofac... which they may or may not want.

Issuing a new package actually does affect a lot of people.

And, yeah, while it's currently all in the same repo, we're in the process of splitting the repos up to be separately built, separately maintained, separately tagged. From a legacy perspective we're in one repo, but you have to treat each package as though it's all independently built, versioned, and released. We're not "masking a problem," we're maintaining backwards compatibility for folks and following semantic versioning patterns on a per-package basis. Each project is effectively its own product, it just happened to be easier to manage historically in one repo.

Using the binding redirect is going to be the solution to this one.

hoffmanic:
If you look at it from a wider perspective, we got in this situation because newtonsoft updated a major version for for us from 5-6, we then upgraded masstransit and then it's masstransit autofac integrations, the only missing piece of the puzzle is autofac 1st party integrations. 

The point I am trying to make is because of the way these dependencies are being consumed the op really has no choice, autofac was already updated by external dependencies.

Perhaps from a semantic versioning perspective this is a minor release, to prevent inclination to update when ops don't necessarily need to?

A binding redirect is not a solution when an issue is isolated to a single authority (autofac). It would be nice for autofac to actually support their own ecosystem.

tillig:
I recognize the challenge here, but given that binding redirects are available and have been the standard solution to this since the start of .NET, there's no real value to releasing a whole new set of packages.

When ASP.NET vNext comes out, [strong naming goes away](http://davidfowl.com/asp-net-vnext/) and so does this headache. We're working on converting Autofac to that now. Until then, we need to leverage the .NET standard fusion/assembly location mechanisms, and that includes binding redirects as needed.

Issue #620, opened 2015-01-31
Use attributes to mark properties that should/should not be injected


twsouthwick:
I'm using AutoFac to resolve properties in a PowerShell cmdlet, but find that the extension method ResolutionExtensions.InjectUnsetProperties does not provide the behavior I need.  I want to mark certain properties as 'Do Not Inject' as they are optional parameters for cmdlet.  An easy work around would be to add an attribute that causes the resolution logic to skip that specific property.

dazinator:
+1

MpDzik:
+1

cbfrank:
+1

tillig:
@twsouthwick Just to close the loop - you submitted PR #622 (which looks good)... would that work _instead_ of specifically adding attributes?

I'm also thinking about custom attributes vs. reuse of existing attributes. What if we added support in the Autofac.Mef library to support `[Import]` in combination with PR #622 - so you would mark the properties you want Autofac to inject with an `[Import]` attribute. That would avoid tying Autofac core to a custom attribute and would look something like

`builder.RegisterType<Foo>().InjectImportProperties()`

which would then call the delegate filter in PR #622 to attach whitelisting to properties marked with `[Import]`.

What do you think?

cbfrank:
Can autofac support to register types with attribute either? for example:
[Reg(AsType, Name/Key...)]
public class Service:IService{
so that autofac can register all type defined in assmebly automatically

Thanks

tillig:
No, there is not currently attribute-based type registration.

cbfrank:
Thanks, will autofac support this kind of registration in the future?

tillig:
Possibly through an extension or community contribution, but generally core Autofac has tried to be more unobtrusive - register by assembly scanning and _your own_ custom attributes rather than littering your code with Autofac-specific attributes.

cbfrank:
I see. Thank you!

alexmg:
Perhaps we could add a `FindPropertiesWith` method similar to the `FindConstructorsWith` method, that allows the user to provide there own `IPropertyFinder` implementation similar to to `IConstructorFinder`. This would only be called with a list of properties that had already been checked as settable, and we would ship a standard implementation that could be replaced. That way we aren't limiting how the properties are discovered, and the user could choose attributes or any other convention they liked, and keep the logic encapsulated in their own classes. Food for thought.

twsouthwick:
[...]

PR #622, opened 2015-02-07
Add delegate to filter which properties are injected

twsouthwick:
This adds an overload to ResolutionExtensions.InjectProperties that takes in a delegate that allows custom filtering of properties that will be injected beyond if they are just set or not.

This resolves the issue described in #620


alexmg:
Thanks for the PR. We are in the middle of adding support for Core CLR in the develop branch. This will be reviewed again for the 4.0 release. It might get lost in the transition if we pull it now.

twsouthwick:
Closing as better implementation is being done in #714

PR #623, opened 2015-02-14
Fixed KeyNotFoundException in AggregateService resolving open generic via a method call

kcuzner:
I was using an AggregateService the other day with a method that was generic with the intention of using it to resolve open generics on the fly. Calling this method resulted in a `KeyNotFoundException`.

```csharp
public interface IOpenGenericAggregate
{
	IOpenGeneric<T> GetOpenGeneric<T>();
}

public class OpenGeneric<T> : IOpenGeneric<T>
{
}
```

Test which fails:

```csharp
builder.RegisterAggregateService<IOpenGenericAggregate>();
builder.RegisterGeneric(typeof(OpenGeneric<>))
	.As(typeof(IOpenGeneric<>));

var aggregateService = container.Resolve<IOpenGenericAggregate>();

var generic = aggregateService.GetOpenGeneric<object>(); //KeyNotFoundException here
Assert.That(generic, Is.Not.Null);
```

I made a couple changes to fix the issue. If this isn't an intended use of the AggregateService I'm ok with these not being included.

The problem was that the existing interceptor stored the generic method definition in the invocation map, but the filled-in generic method (is there a term for that? Closed generic?) was used in the lookup. My changes do the following:

 * If the invoked method is a generic method, the generic method definition is used in the invocation map lookup instead of directly using the invocation method. This prevents the `KeyNotFoundException`.
 * The `IInvocation.Method.ReturnType` is always used as the argument in the `Resolve()` call since the `returnType` found in the `SetupInvocationMap` method might not match the return type of the method when it is called. This handles the case when the method is called to resolve a concrete type based on an open generic.

All of the tests pass, including the new one I added for testing the resolution of open generics. If there are any issues with the changes I'm more than willing to fix them.

Thanks.

### DIFF
--- a/Extras/Source/Autofac.Extras.AggregateService/ResolvingInterceptor.cs
+++ b/Extras/Source/Autofac.Extras.AggregateService/ResolvingInterceptor.cs
@@ -63,7 +63,12 @@ namespace Autofac.Extras.AggregateService
             {
                 throw new ArgumentNullException("invocation");
             }
-            var invocationHandler = _invocationMap[invocation.Method];
+
+            //if the method is generic, we use the definition in the invocation map
+            var method = invocation.Method.IsGenericMethod ?
+                invocation.Method.GetGenericMethodDefinition() : invocation.Method;
+
+            var invocationHandler = _invocationMap[method];
             invocationHandler(invocation);
         }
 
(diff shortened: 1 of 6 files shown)


alexmg:
Thanks for the PR. The code looks solid.

As part of preparing for ASP.NET vNext and Core CLR we split out all the integrations and extras into separate repositories. When we merge back from develop to master all those will disappear.

It would be great if you could submit the PR to the new repository so that the change doesn't get lost. The repository below is now the master for AggregateService.

https://github.com/autofac/Autofac.Extras.AggregateService

Hopefully once we merge the develop branch back it will be more obvious that these projects are longer in the main repository.

PR #640, opened 2015-05-05
Change source of LimitType in log4net doco

piers7:
Using e.Component.**Target**.Activator.LimitType not e.Component.Activator.LimitType (note: extra '.Target') appears to be preferable for log injection when dealing with registrations that use delegate activators, like Func\<Owned\<T\>\>, or Lazy\<T\>
With e.Component.Activator.LimitType you get a logger for Func\<Owned\<**T**\>\>
With e.Component.**Target**.Activator.LimitType you get a logger for **T**


piers7:
I've clearly done something wrong with my pull request haven't I? How did it bring in those other commits?

tillig:
No worries. I figured it out and updated with your change. Thanks for the PR!

Issue #641, opened 2015-06-08
Add 'WhereTypeClosesOpenGenericInterface' registration builder extension


matthewcanty:
I've been using the extension method described in this post:

[Registering open generic interface types in Autofac - Alex Meyer-Gleaves (2009)](http://alexmg.com/registering-open-generic-interface-types-in-autofac/)

Seems to me like it should be an extension method available in the AutoFac features - `OpenGenericRegistrationExtensions`?.

Willing to hear views and submit a pull request - *being sure to cite Alex as the author of course!*

tillig:
Per the blog entry:

> UPDATE (22 December 2009): I have submitted a patch to Nick and this feature has now been added to the Autofac V2 codebase. While including the patch Nick added support for open generic classes and renamed the extension method to `AsClosedTypesOf`. This post will be left in its current form and remains a valid example of extending Autofac.

Have you tried using `AsClosedTypesOf`? We have documentation about it and other scanning methods [on the Autofac doc site](http://autofac.readthedocs.org/en/latest/register/scanning.html).

matthewcanty:
I am using the `AsClosedTypesOf` but we still need this call to discover them all:

```
builder
    .RegisterAssemblyTypes(AppDomain.CurrentDomain.GetAssemblies())
    .WhereTypeClosesOpenGenericInterface(typeof(IMessageHandler<>))
    .AsClosedTypesOf(typeof(IMessageHandler<>));
```

tillig:
If you look at the source of the `AsClosedTypesOf` method, [it appears to already do the filtering for you.](https://github.com/autofac/Autofac/blob/c9b56b7737df387afa667a6833b9d6100f6100be/src/Autofac/Features/Scanning/ScanningRegistrationExtensions.cs#L117) Like, it literally looks nearly identical to what you posted. That's why I ask.

    builder.RegisterAssemblyTypes(assemblies).AsClosedTypesOf(typeof(IMessageHandler<>));

We have [some unit tests showing this working](https://github.com/autofac/Autofac/blob/c9b56b7737df387afa667a6833b9d6100f6100be/test/Autofac.Test/Features/Scanning/ScanningRegistrationTests.cs#L175). There is a whole separate [assembly with lots of different types in it](https://github.com/autofac/Autofac/tree/c9b56b7737df387afa667a6833b9d6100f6100be/test/Autofac.Test.Scenarios.ScannedAssembly) that we use to verify the discovery and filtering.

[...]

PR #642, opened 2015-06-08
Add 'Where Type Closes Open Generic Interface' extension

matthewcanty:
Addresses [#641 - *Add 'WhereTypeClosesOpenGenericInterface' registration builder extension*](https://github.com/autofac/Autofac/issues/641)

### DIFF
--- a/src/Autofac/Features/Scanning/ScanningRegistrationExtensions.cs
+++ b/src/Autofac/Features/Scanning/ScanningRegistrationExtensions.cs
@@ -168,5 +168,39 @@ namespace Autofac.Features.Scanning
 
             return registration;
         }
+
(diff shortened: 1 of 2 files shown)


tillig:
Per the blog entry:

> UPDATE (22 December 2009): I have submitted a patch to Nick and this feature has now been added to the Autofac V2 codebase. While including the patch Nick added support for open generic classes and renamed the extension method to `AsClosedTypesOf`. This post will be left in its current form and remains a valid example of extending Autofac.

Have you tried using `AsClosedTypesOf`? We have documentation about it and other scanning methods [on the Autofac doc site](http://autofac.readthedocs.org/en/latest/register/scanning.html).

tillig:
I'm going to close this for now since `AsClosedTypesOf` appears to do the same thing. Let me know if we need to reopen.

PR #643, opened 2015-06-09
Providing the instance type being property injected onto

sheastrickland:
Which allows things like closing generics for `Serilog.ForContext(type)` and `LogManager.GetLogger(type)` by accessing the type parameter during preparation or activation. (legacy WebForms in this case)


sheastrickland:
Didn't have access to latest VS, will amend tests when I can get my hands on it later.

-Thanks

sheastrickland:
Morning @alexmg and @tillig, looking to get this bad boy in when possible. Any chance for a quick review? Happy to make any changes to help get this little fella over the line. As always, love your work.

tillig:
Sorry, we've been sort of busy trying to get DNX support stable. Given that we're in the middle of that, even if we pull this in, it'll end up in a CI build on MyGet until VS support for DNX beta 5 is figured out. We've already merged a 4.0 beta to the master/release line, so there's not really any going back. :)

We do appreciate the PR and it looks like it shouldn't be a problem to bring in.

In the meantime, can you...

* Make the pull request off the "develop" branch instead of the "master" branch?
* Add a unit test to demonstrate how this works?

nblumhardt:
One other thought - since the presence of a `TypedParameter` might alter constructor selection in existing cases (i.e. if an alternative constructor accepting `Type` appears on the target, it may now be accidentally chosen) is it possible to pass this as a `NamedParameter` instead, and supply a constant for the parameter name?

Cheers!

sheastrickland:
Yeah, that's what I thought and why I added the comment to the code, I'll go down that path.

Issue #644, opened 2015-06-11
Logging Injector


piers7:
This isn't really an issue, but I put together a generic version of the logging injection module that's in the docs. Since it's generic, it no longer has any dependencies on log4net (or any logging framework), it's just a way of automatically resolving dependencies via a Type->dependency factory delegate. So my thoughts are that it might be handy in core Autofac library (given it's quite a common pattern).

https://gist.github.com/piers7/81724d51a7ca158d721e

Let me know if you are interested in me pushing a PR

tillig:
Ooooh, I like that, it's pretty slick. It might be good to put it in an Extras package, though, rather than in core, so we can update it/add to it as needed without affecting core Autofac. Do you think that would make sense? Like an Autofac.Extras.Logging or something?

@alexmg, what's your take?

alexmg:
I like it. Yeah, probably not a core feature but could be useful in an Extras package. I'm looking for an alternative to this PR too.

https://github.com/autofac/Autofac/pull/645

sheastrickland:
Thanks @piers7, looks great, and like we were on a similar journey :)
We had similar (referenced PR), except with some added hooks for PropertyInjection in a legacy codebase.

piers7:
@tillig *Personally* I'm not sold on creating additional packages (and hence nugets) without a compelling reason. Projects of any size already drown in a nuget soup, and I'd rather it didn't get worse for the sake of a handful of classes. Avoiding host-specific dependencies (Autofac.MVC etc...) for sure, but (for example) I don't see why Autofac.Config has to be separate, and I'd make the same argument here. Just for one thing, discoverability really suffers.
I totally understand the reluctance to go down paths that may lead to breaking changes in future, and avoid creeping bloat, but I look at my projects today and think: wow, that's a *lot* of assemblies.
Just my 2c.

piers7:
@sheastrickland I did see your PR, but to be honest I didn't really grok what it was doing at the time. I'll take another look.

sheastrickland:
@piers7, yeah I've added a lot more information and a gist to help.

piers7:
@sheastrickland Right. I tried to extend my gist to handle log injection into properties also, and _finally_ understood what your PR is all about, and what you've had to do, and why. And that will be really helpful for use cases like this. I've made some other comments on the PR itself.

piers7:
[...]

PR #645, opened 2015-06-13
Provide the instance type being property injected onto via NamedParameter

sheastrickland:
Moved here via suggestion from @tillig in PR #643 

//cc @nblumhardt

### DIFF
--- a/src/Autofac/Autofac.csproj
+++ b/src/Autofac/Autofac.csproj
@@ -388,7 +388,6 @@
       <DependentUpon>ReflectionExtensionsResources.resx</DependentUpon>
     </Compile>
     <Compile Include="Util\ReleaseAction.cs" />
-    <Compile Include="Util\ResourceManagerHack.cs" />
     <Compile Include="Util\SafeDictionary.cs" />
     <Compile Include="Util\SequenceExtensions.cs" />
     <Compile Include="Util\Traverse.cs" />
--- a/src/Autofac/Core/Activators/Reflection/AutowiringPropertyInjector.cs
+++ b/src/Autofac/Core/Activators/Reflection/AutowiringPropertyInjector.cs
@@ -32,6 +32,8 @@ namespace Autofac.Core.Activators.Reflection
 {
     class AutowiringPropertyInjector
     {
+        public const string InstanceTypeNamedParameter = "Autofac.AutowiringPropertyInjector.InstanceType";
+
         public static void InjectProperties(IComponentContext context, object instance, bool overrideSetValues)
         {
             if (context == null) throw new ArgumentNullException(nameof(context));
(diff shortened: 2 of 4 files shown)


alexmg:
I have merged this but it isn't sitting quite right with me at the moment.

Can you post an example of using this property with Serilog so that we can think about some other approaches?

How does your objective line up with the sample below?

http://docs.autofac.org/en/latest/examples/log4net.html

sheastrickland:
Thanks @alexmg, the main objective is injecting properties on an existing instance. The particular case where this arose, was using the `Autofac.Web` package and utilising the Http Modules which perform property injection on existing UserControls and Page instances. 

The log4net examples all work when the property is injected while constructing the instance containing the property to inject, however; when an existing instance of the containing type is injected onto after, it doesn't fire the Activated event.

The Http Modules call [ResolutionExtensions.InjectProperties](https://github.com/autofac/Autofac/blob/master/src/Autofac/ResolutionExtensions.cs#L51). Which proxies down to [AutowiringPropertyInjector.InjectProperties](https://github.com/autofac/Autofac/blob/master/src/Autofac/Core/Activators/Reflection/AutowiringPropertyInjector.cs#L60), so unless the TLogger is registered in the container it will short circuit here.

So, after registering the TLogger factory, which will pass the previous short circuit, when it gets to the [Resolve](https://github.com/autofac/Autofac/blob/master/src/Autofac/Core/Activators/Reflection/AutowiringPropertyInjector.cs#L71), it's missing the instanceType context.

Enter the `NamedParameter` :)

It's hard to articulate, though in short: Injecting onto an existing type with `InjectProperties` and it's kin. Doesn't provide the same hooks as what `AttachToComponentRegistration` and `IComponentRegistration` provide.

sheastrickland:
So, similar to PR #644, we have this:

https://gist.github.com/sheastrickland/e2899cd7abe61e20de69

With the additional hooks for PropertyInjection when the containing type is resolved, as well as when an existing instance it property injected onto, after the fact.

piers7:
Nitpicking, but shouldn't InstanceTypeNamedParameter be `public static readonly`, not `public const`, to avoid binary versioning issues if you ever change the magic string?

piers7:
So now I want to achieve the same as @sheastrickland I had a bit of a poke around, and now finally understand (and generally concur with) the summary above.

I think the root cause here is there just aren't the same hooks for controlling Property injection that exist for Parameters. Were it possible to extend a component's _Properties_ during Preparing/Activating (just as you can currently extend their Parameters) then we wouldn't need to do this - a log injection module could just add that dependency explicitly.

The approach used in the log4net doco sample waits until after the component has been Activated, then re-does half of the work that the activator has just done (scanning the properties, matching by type, checking it's not alread set). I'm currently using that approach, but I've not had to deal with @sheastrickland's issue with Activated not firing in some cases. Eitherway, it feels like a workaround to something that's missing.

Another potential option would be if the property injection engine (the InjectUnsetProperties) method were itself to be able to be re-used here, by explicitly passing parameters (or object instances) to it, and it auto-injecting the residual. Can't say I fancy that much, seems like too big a change.

I can't say I massively like the magic-string-named-parameter, but it does seem to me to be a neat, discrete fix for something that otherwise needs a fair bit of effort to unravel.

PR #698, opened 2015-11-18
Updated to work with RC1

shawnwildermuth:


src/Autofac/project.json
```diff
@@ -21,54 +21,54 @@
...
-                "System.Diagnostics.Tools": "4.0.1-beta-*",
-                "System.Globalization": "4.0.11-beta-*",
-                "System.Linq": "4.0.1-beta-*",
-                "System.Linq.Expressions": "4.0.11-beta-*",
-                "System.Reflection": "4.1.0-beta-*",
-                "System.Reflection.Extensions": "4.0.1-beta-*",
-                "System.Resources.ResourceManager": "4.0.1-beta-*",
-                "System.Runtime": "4.0.21-beta-*",
-                "System.Threading.Tasks": "4.0.11-beta-*"
+                "System.Collections": "4.0.11-rc-*",
```
tillig: Should this be `rc1`?

tillig:
Looks like this is going to take a bit more work to get going. Unfortunately, they didn't actually release rc1 of every package - for example, [Microsoft.Extensions.DependencyInjection got an rc1 release](https://www.nuget.org/packages/Microsoft.Extensions.DependencyInjection/1.0.0-rc1-final) but [System.Collections is still beta](https://www.nuget.org/packages/System.Collections/4.0.11-beta-23516). Switching all the deps to `-rc1-*` actually ends up with `-rc2-*` dependencies because on the core packages it appears they skipped straight from beta to rc2.

Sigh.

It's never easy. :frowning:

tillig:
Thanks for the kickstart, though, it's appreciated.

PR #714, opened 2016-02-09
Provide a way to register property injection

### LINKED ISSUE #620: Use attributes to mark properties that should/should not be injected

twsouthwick:
There are times when property injection is needed and there needs to be a way to control which properties are available. This follows the pattern of FindConstructorsWith and IConstructorFinder by creating a FindPropertiesWith method and a IPropertyFinder that can be used to determine which properties should be injected.

Fixes #620

### DIFF
--- a/src/Autofac/Core/Activators/Reflection/ReflectionActivator.cs
+++ b/src/Autofac/Core/Activators/Reflection/ReflectionActivator.cs
@@ -55,18 +55,21 @@ namespace Autofac.Core.Activators.Reflection
             Type implementationType,
             IConstructorFinder constructorFinder,
             IConstructorSelector constructorSelector,
+            IPropertyFinder propertyFinder,
             IEnumerable<Parameter> configuredParameters,
             IEnumerable<Parameter> configuredProperties)
             : base(implementationType)
         {
             if (constructorFinder == null) throw new ArgumentNullException(nameof(constructorFinder));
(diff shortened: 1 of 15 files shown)


src/Autofac/Core/Activators/Reflection/ReflectionActivator.cs
```diff
@@ -55,18 +55,21 @@ public class ReflectionActivator : InstanceActivator, IInstanceActivator
             Type implementationType,
             IConstructorFinder constructorFinder,
             IConstructorSelector constructorSelector,
+            IPropertyFinder propertyFinder,
```
twsouthwick: I wasn't sure if I should add this here or as another overload. Since this is a public constructor on a public type, it will break anything that has taken a dependency on it. I added here for now (since 4.0 hasn't been finalized), but can create an overload if preferred.

twsouthwick:
Didn't open against `develop` branch

twsouthwick:
Opened #715 instead

PR #715, opened 2016-02-09
Add property injection control

### LINKED ISSUE #620: Use attributes to mark properties that should/should not be injected

twsouthwick:
There are times when property injection is needed and there needs to be
a way to control which properties are available. This follows the
pattern of FindConstructorsWith and IConstructorFinder by creating a
FindPropertiesWith method and a IPropertyFinder that can be used to
determine which properties should be injected.

Fixes #620

### DIFF
--- a/src/Autofac/Builder/RegistrationBuilder{TLimit,TActivatorData,TRegistrationStyle}.cs
+++ b/src/Autofac/Builder/RegistrationBuilder{TLimit,TActivatorData,TRegistrationStyle}.cs
@@ -363,12 +363,10 @@ namespace Autofac.Builder
         /// container and follow specific criteria will be wired to instances of the appropriate service.
         /// </summary>
         /// <param name="propertySelector">Selector to determine which properties should be injected</param>
-        /// <param name="wiringFlags">Set wiring options such as circular dependency wiring support.</param>
+        /// <param name="allowCircularDependencies">Determine if circular dependencies should be allowed or not.</param>
         /// <returns>A registration builder allowing further configuration of the component.</returns>
-        public IRegistrationBuilder<TLimit, TActivatorData, TRegistrationStyle> PropertiesAutowired(IPropertySelector propertySelector, PropertyWiringOptions wiringFlags)
+        public IRegistrationBuilder<TLimit, TActivatorData, TRegistrationStyle> PropertiesAutowired(IPropertySelector propertySelector, bool allowCircularDependencies)
         {
-            var allowCircularDependencies = (int)(wiringFlags & PropertyWiringOptions.AllowCircularDependencies) != 0;
-
             if (allowCircularDependencies)
                 RegistrationData.ActivatedHandlers.Add((s, e) => AutowiringPropertyInjector.InjectProperties(e.Context, e.Instance, propertySelector));
             else
--- a/src/Autofac/Builder/RegistrationBuilder{TLimit,TActivatorData,TRegistrationStyle}.cs
+++ b/src/Autofac/Builder/RegistrationBuilder{TLimit,TActivatorData,TRegistrationStyle}.cs
@@ -360,21 +360,6 @@ namespace Autofac.Builder
 
         /// <summary>
         /// Configure the component so that any properties whose types are registered in the
-        /// container will be wired to instances of the appropriate service.
-        /// </summary>
-        /// <param name="wiringFlags">Set wiring options such as circular dependency wiring support.</param>
-        /// <returns>A registration builder allowing further configuration of the component.</returns>
-        public IRegistrationBuilder<TLimit, TActivatorData, TRegistrationStyle> PropertiesAutowired(PropertyWiringOptions wiringFlags)
-        {
-            var preserveSetValues = (int)(wiringFlags & PropertyWiringOptions.PreserveSetValues) != 0;
-
-            PropertiesAutowired(new DefaultPropertySelector(preserveSetValues), wiringFlags);
-
-            return this;
-        }
-
-        /// <summary>
-        /// Configure the component so that any properties whose types are registered in the
         /// container and follow specific criteria will be wired to instances of the appropriate service.
         /// </summary>
         /// <param name="propertySelector">Selector to determine which properties should be injected</param>
--- a/src/Autofac/Builder/RegistrationBuilder{TLimit,TActivatorData,TRegistrationStyle}.cs
+++ b/src/Autofac/Builder/RegistrationBuilder{TLimit,TActivatorData,TRegistrationStyle}.cs
@@ -366,13 +366,28 @@ namespace Autofac.Builder
         /// <returns>A registration builder allowing further configuration of the component.</returns>
         public IRegistrationBuilder<TLimit, TActivatorData, TRegistrationStyle> PropertiesAutowired(PropertyWiringOptions wiringFlags)
         {
-            var allowCircularDependencies = (int)(wiringFlags & PropertyWiringOptions.AllowCircularDependencies) != 0;
             var preserveSetValues = (int)(wiringFlags & PropertyWiringOptions.PreserveSetValues) != 0;
 
+            PropertiesAutowired(new DefaultPropertySelector(preserveSetValues), wiringFlags);
+
+            return this;
+        }
+
+        /// <summary>
+        /// Configure the component so that any properties whose types are registered in the
+        /// container and follow specific criteria will be wired to instances of the appropriate service.
+        /// </summary>
+        /// <param name="propertySelector">Selector to determine which properties should be injected</param>
+        /// <param name="wiringFlags">Set wiring options such as circular dependency wiring support.</param>
+        /// <returns>A registration builder allowing further configuration of the component.</returns>
+        public IRegistrationBuilder<TLimit, TActivatorData, TRegistrationStyle> PropertiesAutowired(IPropertySelector propertySelector, PropertyWiringOptions wiringFlags)
+        {
+            var allowCircularDependencies = (int)(wiringFlags & PropertyWiringOptions.AllowCircularDependencies) != 0;
+
             if (allowCircularDependencies)
-                RegistrationData.ActivatedHandlers.Add((s, e) => AutowiringPropertyInjector.InjectProperties(e.Context, e.Instance, !preserveSetValues));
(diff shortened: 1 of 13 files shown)


src/Autofac/Core/Activators/Reflection/ReflectionActivator.cs
```diff
@@ -56,18 +56,21 @@ public class ReflectionActivator : InstanceActivator, IInstanceActivator
             Type implementationType,
             IConstructorFinder constructorFinder,
             IConstructorSelector constructorSelector,
+            IPropertyFinder propertyFinder,
```
twsouthwick: I wasn't sure if I should add this here or as another overload. Since this is a public constructor on a public type, it will break anything that has taken a dependency on it. I added here for now since 4.0 hasn't been finalized, but can create an overload if preferred.

alexmg:
Thanks for the PR @twsouthwick. This is definitely heading in the right direction though there are a few things that need to be considered.

The `InjectProperties` method in the `ReflectionActivator` is meant to set properties for configured properties on the registration, or property parameters provided at the time of resolution. Any property that is autowired should be set in the `AutowiringPropertyInjector`. This has a lot more validation about the types of properties that can be resolved and whether they can be overridden.

The `ConstructorFinder` is specific to the `ReflectionActivator` as this is needed to find the constructor to create the instance with, which is only relevant to services activated via reflection. Obviously a delegate or provided instance doesn't require a mechanism to find the appropriate constructor, as in the former the lambda has defined the constructor call, and in the later it is irrelevant.

In the case of the `PropertyFinder` the same doesn't hold true. Even though you can `Resolve` a property value in the lambda directly, you can also choose to apply `PropertiesAutowired()` as well. Both `builder.Register(c => new A { B = c.Resolve<B>() })` and `builder.Register(c => new A()).PropertiesAutowired()` are legitimate registrations that will result in the `B` property being injected. In fact, even though it really shouldn't be allowed, you can even inject properties of a provided instance `builder.RegisterInstance(new A()).PropertiesAutowired()`. Yeah, we should really raise a bug for that one, even though I've never heard of anyone doing that.

Anyway, the limiting of properties that should be injected is more broadly applicable than just to reflection based activations. That means that the `PropertyFinder` should be available on the `SimpleActivatorData` as well, and unfortunately there isn't currently a common base class with `ReflectionActivatorData`, and the interface `IConcreteActivatorData` isn't applied until `ConcreteReflectionActivatorData`. It would be good to find a clean way of resolving these type discrepancies without repeating code as `PropertiesAutowired` is currently open to all `TActivatorData`. The type inference in the builder API can get a bit tricky so need to be careful about this.

ffMathy:
What is the status on this?

twsouthwick:
@alexmg How about this approach? I added another overload of `PropertiesAutowired()` to IRegistrationBuilder. It actually is less code churn as well.

twsouthwick:
There's discussion on a commit that's been squashed away: https://github.com/autofac/Autofac/commit/c0813dc70c24e2bf5988aeeb8a57e36caefa4363

tillig:
The latest `dotnet` cli seems to have broken - the build failure on 5/31/2016 isn't due to this branch, it's a cli problem.

twsouthwick:
From @tillig at https://github.com/autofac/Autofac/commit/c0813dc70c24e2bf5988aeeb8a57e36caefa4363

> I don't see any breaking changes, but I do worry that someone might shoot themselves in the foot with a custom implementation. For example, our current filtering mechanism filters out things like indexed properties which we can't really set because we don't have an index to pass. It also filters out value types, which helps in situations like #749. Without the default filtering in place it's very flexible but implies potential support overhead as people ask questions like, "How come I marked this indexed property with my attribute and it's not injecting?"

> I'm curious if the additional property selector should be in addition to the default filtering. For example, run the default filtering, then also run the custom selector.

> In that case, the custom selector could be a Func<bool, PropertyInfo> rather than an interface. (I don't feel strongly about that. Just saying.)

> Given that, perhaps the default filtering may need to be relaxed a bit. Like, it currently requires a public setter. Maybe that's something that doesn't need to be checked, or, if we keep the IPropertySelector interface, that could be part of the DefaultPropertySelector. I could see a custom IPropertySelector may want to allow internal or private properties to be injected.

The main reason I went with the interface design was in case a filter was something that required some sort of more complicated state that would be better kept in a class implementation. I'm not tied to an interface vs func implementation. If there is some sort of default filtering that cannot be customized, I think it should only be for things that are technically impossible (like setting property indexers) and not for style (such as only setting public properties).

src/Autofac/Builder/RegistrationBuilder{TLimit,TActivatorData,TRegistrationStyle}.cs
```diff
@@ -366,13 +366,28 @@ public RegistrationBuilder(Service defaultService, TActivatorData activatorData,
         /// <returns>A registration builder allowing further configuration of the component.</returns>
         public IRegistrationBuilder<TLimit, TActivatorData, TRegistrationStyle> PropertiesAutowired(PropertyWiringOptions wiringFlags)
```
twsouthwick: This method is now somewhat redundant - is there any desired to move this as an extension method instead of defined on the interface itself?

src/Autofac/Builder/RegistrationBuilder{TLimit,TActivatorData,TRegistrationStyle}.cs
```diff
@@ -366,13 +366,28 @@ public RegistrationBuilder(Service defaultService, TActivatorData activatorData,
...
+        }
+
+        /// <summary>
+        /// Configure the component so that any properties whose types are registered in the
+        /// container and follow specific criteria will be wired to instances of the appropriate service.
+        /// </summary>
+        /// <param name="propertySelector">Selector to determine which properties should be injected</param>
+        /// <param name="wiringFlags">Set wiring options such as circular dependency wiring support.</param>
+        /// <returns>A registration builder allowing further configuration of the component.</returns>
+        public IRegistrationBuilder<TLimit, TActivatorData, TRegistrationStyle> PropertiesAutowired(IPropertySelector propertySelector, PropertyWiringOptions wiringFlags)
```
twsouthwick: if `wiringFlags` has the flag `PropertyWiringOptions.PreserveSetValues`, it won't do anything here. Is there any issue with that? Should it be in the summary or remarks?

src/Autofac/Builder/RegistrationBuilder{TLimit,TActivatorData,TRegistrationStyle}.cs
tillig: I'd be OK with that. It is a 4.0 release and that seems like a legit breaking change. Not sure it'd actually break anyone since the number of folks implementing their own registration builder should be pretty low.

src/Autofac/Builder/RegistrationBuilder{TLimit,TActivatorData,TRegistrationStyle}.cs
tillig: Seems like PropertyWiringOptions itself is somewhat obsolete with the new design. I wonder if it's just "allowCircularDependencies" as a bool now. @alexmg?

tillig: Oh, nevermind. We do still need it, for the above one where we use the default selector.

tillig:
I agree we wouldn't want to be overzealous in the always-implemented filtering. Just want to get people into that "pit of success," as it were.

What if we always filtered out value types and indexed properties (the first four "if" checks in there) and left the checks for public / writable / preserve set values in the `DefaultPropertySelector`? I think that would stop people from accidentally cornering themselves while still allowing for private/write-only properties, attributes, or whatever other behavior.

twsouthwick:
That's a good compromise. I wondered about that while refactoring it, but wasn't sure what was technically impossible vs "best practices". I'll update the PR

src/Autofac/Builder/RegistrationBuilder{TLimit,TActivatorData,TRegistrationStyle}.cs
twsouthwick: It will be a breaking change AFAIK at the binary level (it'll break anything compiled against the old version since the signature of the method will change), but not as far as API is concerned (the compiler will resolve the new method without issue). Since it's still pre-release of a major semantic version change (4.0), seems fine, but just wanted to make sure before refactoring.

src/Autofac/Builder/RegistrationBuilder{TLimit,TActivatorData,TRegistrationStyle}.cs
twsouthwick: Should I just have an arg of `bool allowCircularDependencies`?

tillig: I feel like that's probably more appropriate, at least here. The bool could be calculated above as needed when the `DefaultPropertySelector` is used.

src/Autofac/Builder/RegistrationBuilder{TLimit,TActivatorData,TRegistrationStyle}.cs
twsouthwick: I've updated the PR to show what this is like (it's an independent commit for now so it can be removed)

src/Autofac/Builder/RegistrationBuilder{TLimit,TActivatorData,TRegistrationStyle}.cs
twsouthwick: Done.

tillig:
I think it's good. Let @alexmg wake up and get another look now that the changes are in and we can pull if there's nothing left. Thanks!

tillig:
I'm going to go ahead and pull this in. If there are any additional changes we can just refactor on the back end. Thanks for the great work and the patience - it's greatly appreciated!

PR #719, opened 2016-02-29
Ignore compiler generated classes on autoregistration

CosminLazar:
When dealing with deferred IEnumerables (using yield return/break) the compiler generates a helper class to handle the deferred execution. Autoregistration discovers this class and registers it as a provider for IEnumerable<T> thus preventing the user from being able to resolve all implementations of T.

### DIFF
--- a/test/Autofac.Test/Features/Scanning/ScanningRegistrationTests.cs
+++ b/test/Autofac.Test/Features/Scanning/ScanningRegistrationTests.cs
@@ -438,7 +438,7 @@ namespace Autofac.Test.Features.Scanning
         }
 
         [Fact]
-        public void DefferedEnumerableHelperClassDoesNotGetRegistered()
+        public void DeferredEnumerableHelperClassDoesNotGetRegistered()
         {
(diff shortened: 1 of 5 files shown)


test/Autofac.Test/Features/Scanning/ScanningRegistrationTests.cs
```diff
@@ -436,5 +436,15 @@ public void ScanningKeyedRegistrationsFilterByAssignabilityBeforeMappingKey()
             }));
             Assert.True(c.IsRegisteredWithKey<IAService>(k));
         }
+
+        [Fact]
+        public void DefferedEnumerableHelperClassDoesNotGetRegistered()
```
tillig: Minor typo - two 'f's in the test name "Defferred" => "Deferred"

tillig:
Nice catch! I think there's one minor typo to fix up and we can pull it. Thanks!

CosminLazar:
Fixed the typo. Glad that I could help!

tillig:
Pulled. Thanks again!

Issue #718, opened 2016-02-15
Instance Creation Issue with Property injection and lazy instantiation


blsmithers:
Hi, I stumbled on an issue and think it might be similar to the issue described in https://github.com/autofac/Autofac/issues/666.  The below is failing for me with the error message "An item with the same key has already been added.".  I was hoping this case would be possible, any help you could give me would be appreciated!

```
static void Main(string[] args) {
            var builder = new ContainerBuilder();

            builder.RegisterType<A>().SingleInstance();
            builder.RegisterType<B>().SingleInstance().PropertiesAutowired(PropertyWiringOptions.AllowCircularDependencies);

            using (var container = builder.Build())
            {
                var myA = container.Resolve<A>();
            }
        }

        public class A
        {
            public A(Lazy<B> b)
            {
                var myB = b.Value;
            }
        }

        public class B 
        {
            public A A { get; set; }
        }
```

tillig:
Verified. It does seem to be around the use of `Lazy<T>` since removing `Lazy<T>` and using `T` directly doesn't encounter this. We'll have to see if we can track this down. Thanks!

tillig:
While the exception message is different now, this is still an issue. The issue is getting caught in `LifetimeScope.GetOrCreateAndShare` and it seems the use of `Lazy<T>` doesn't carry with it the circular property dependency information. I've added a failing unit test (skipped) to the solution.

Note we also have #798 where we're trying to figure out better circular dependency handling. The use of `Lazy<T>` needs to be considered there.

tillig:
This appears to have been resolved with the overhaul to the internals in v6. The [skipped/failing test for this issue](https://github.com/autofac/Autofac/blob/d20dc415cfeb9b623df2195d8abb9f977591ea68/test/Autofac.Test/Features/LazyDependencies/LazyRegistrationSourceTests.cs#L50) can be marked to run and it now passes.

Fix will be released with v6.

Issue #688, opened 2015-10-26
Resolving services registered as open generics with implemented interfaces throws ArgumentException


idursun:
``` csharp
interface IDouble<T,R> { }
internal interface ISingle<T> :  IDouble<T, int> { }
class CDerivedSingle<T>: ISingle<T> { }
class Program
{
    static void Main(string[] args)
    {
        var builder = new ContainerBuilder();
        builder.RegisterGeneric(typeof (CDerivedSingle<>)).AsImplementedInterfaces();
        var container = builder.Build();
        var resolved = container.Resolve<ISingle<double>>();
    }
}
```

Resolving service throws the following exception:

```
ArgumentException: The number of generic arguments provided doesn't equal the arity of the generic type definition.
```

tillig:
Thanks! I've added the example [as a unit test](https://github.com/autofac/Autofac/blob/ba576abc1a1c2f3a4287ef5b53647c48d5927af9/test/Autofac.Test/Features/OpenGenerics/ComplexGenericsTests.cs#L186) so we can start figuring out what's up.

idursun:
Thanks. Exception happens at [OpenGenericServiceBinder.cs](https://github.com/autofac/Autofac/blob/master/src/Autofac/Features/OpenGenerics/OpenGenericServiceBinder.cs#L63) 

I have tried to solve it but couldn't find any way of comparing generic parameter count to serviceGenericArguments count because it seems GetGenericArguments or any similar method is not supported in DNX Core.

tillig:
It only took a year and a half but I figured it out. We're now filtering the list of generic interfaces that we _try_ to fulfill by the number of generic parameters provided. Which is to say, if you only provide a single generic parameter (`ISingle<double>` -> `double`) then we don't try to fill in an interface that has two parameters (`IDouble<T, U>`) because, well, we only have one parameter to work with.

Issue #594, opened 2014-11-05
Autofac 4.0 and ASP.NET Core / .NET Core Tasks


alexmg:
This issue contains a list of (potential) tasks that need to be performed for Autofac 4.0 and adding support ASP.NET vNext.
- [x] Create a develop branch following Gitflow.
- [x] Make sure MyGet can build from the develop branch.
- [x] Update the Autofac project to Profile259 to make the PCL compatible with aspnetcore50 framework target.
- [x] Add .kproj files for Autofac and Autofac.AspNet.
- [x] Add sample application for ASP.NET vNext.
- [x] Remove NuGet dependencies on Autofac package while changes being made.
- [x] Split out non-core projects to separate repositories.
- [x] ~~Use MSBuild target with .nuspec in each project to create NuGet packages.~~ Package build is now done as part of AppVeyor build.
- [x] ~~Fix the assembly, file and package version numbers.~~ Versioning is now controlled through appveyor.yml file.
- [x] Tag the master branch with 3.5 prior to merging 4.0 into master (part of Gitflow anyway, but tags are required for doc site versions).
- [x] Tag the master branch with 4.0 after bringing the develop code in (also part of Gitflow, also required for doc site).
- [x] Update Autofac.Extensions.DependencyInjection to be strong named when [the strong named version of Microsoft.Extensions.DependencyInjection is released](https://github.com/aspnet/DependencyInjection/blob/1e21385ab12efacf1c8456809fbae87f85fa98eb/src/Microsoft.Extensions.DependencyInjection/project.json#L10).
- [x] Update to netstandard from dotnet TFM when released.
- [x] [Add support for `IGetRequiredServiceAware`](https://github.com/aspnet/DependencyInjection/issues/378) in Autofac.Extensions.DepedencyInjection
- [x] Add release notes to project.json/nuspec files to point to the wiki release note pages for each package. (Moved to #590)
- [x] Update the package icon URL in project.json/nuspec to point to the GitHub URL for the icon rather than Google Code.
- [x] Create a new version of the documentation site for 4.0.
- [x] Finalize the 4.0 behavior based on [changes that are being looked at for ASP.NET Core integration](https://github.com/aspnet/DependencyInjection/issues/433).
- [x] Generate an updated version of the API documentation for 4.0.
- [x] Update the autofac.org web site for 4.0. (Moved to #823)
- [x] Finish porting [the list of integration libraries](https://github.com/autofac/Autofac/issues/594#issuecomment-139563831)
[...]

PR #744, opened 2016-04-29
set nameof instead of string for ArgumentNullException parameters

GeertVL:


tillig:
Perfecto! Again, thanks, and sorry for the previous confusion. We do appreciate the help and the patience!

GeertVL:
no problem. I would like to help more. Could you point me to an 'easy' issue to start with?

tillig:
Great question. We'd love the help!

Many of the easy items we take care of pretty quickly, but some ideas:

- We've got a couple of issues on the list that we've had time to create failing unit tests for but haven't had a chance to solve. #688 is an example of that.
- There are some issues we haven't had the ability to track down the root cause of. For example, #718 - I verified that the issue exists but I haven't got any failing unit tests in place (and marked ignore so the build doesn't fail) nor do we know the root cause.
- We have a lot of other repos for integrations that have some simpler issues than the core repo has. For example, we've had requests for [an MvvmCross 4.0 compatible integration](https://github.com/autofac/Autofac/issues/688) and there's even a pull request for it, though the PR there has conflicts so we can't just take it.
- Any help at all on getting the core repo or others converted to RC2 and reliably working, running code analysis, XML doc, etc. would be more than welcome. We're trying to track that work on #594.

Mostly the core Autofac stuff we're trying to get done right now is a reliable RC2 build with all the stuff we used to have, so issues with new features are probably less interesting than issues with bug fixes or getting things RC2 compatible across all the repos.

Thanks again!

Issue #747, opened 2016-05-18
Modules are registered more than once when scanning assemblies


caioproiete:
When calling `RegisterAssemblyModules` to register all modules in one or more assemblies, any module that have already previously been registered, is registered again.

Example:

```
builder.RegisterModule(new MyCustomModule());

var assembliesInAppDomain = AppDomain.CurrentDomain.GetAssemblies().ToArray();
builder.RegisterAssemblyModules(assembliesInAppDomain);
```

As you can see in the snippet above, I'm already explicitly registering `MyCustomModule`, however it will get registered again by the `RegisterAssemblyModules` call, given that `MyCustomModule` is definitely somewhere in one of the assemblies loaded in the AppDomain.

I believe we should change `RegisterAssemblyModules` to skip modules already registered.

Thoughts?

tillig:
This is also sort of "functions as designed." Since modules can be registered with parameters, there's no way to tell if you're doing this...

```csharp
// Oops - duplicate!
builder.RegisterModule<MyModule>();
builder.RegisterModule<MyModule>();
```

Or intentionally doing this...

```csharp
// Same module acts differently based on parameters
builder.RegisterModule(new MyModule("param1"));
builder.RegisterModule(new MyModule("param2"));
```

Given we can't key off of type alone, there's not much we can key off of to determine if something is truly a duplicate.

I could see potentially adding LINQ support to assembly module scanning/registration so it has support like `RegisterAssemblyTypes`, something like this:

```csharp
builder.RegisterAssemblyModules(assemblies).Where(t => t != typeof(MyModule));
```

However, we don't really have that right now. Happy to take pull requests, though.

I'll leave this issue open as an enhancement to remind us that this may be an interesting feature we could add.

In the meantime, there's not really a workaround or fix for you that I can see, other than to maybe use things like `PreserveDefaults` to avoid overriding things that have already been set, or to be more selective about the assembly scanning that happens so you don't double-register.

tillig:
As mentioned in #748, one option might be to have an attribute-based approach for skipping module registration for a specific module, but I think the LINQ approach would offer more flexibility (and could technically allow for attribute-based handling later on).

Another workaround here might be that instead of using `RegisterAssemblyModules`, you could also just do a foreach over the set of appropriate assembly types and register them yourself. Off the top of my head (no compiler) it'd look like...

```
var modules = assemblies
    .SelectMany(a => a.GetTypes())
[...]

PR #752, opened 2016-05-24
Prevent duplicate registration of Modules by checking equality

caioproiete:


tillig:
Given the discussion in #747 and the desire to not break the module interface I'm going to close this, though I appreciate the submission. I'm not sure what the right solution is, but we can continue discussion over in #747 to see if there's a way we can still solve the issue.

PR #764, opened 2016-06-23
Removed redundant pinning comment

kierenj:
Just browsing some recent commits and noticed the pinning was undone, but comment remains.  Pedantic yes, but maybe helpful!


tillig:
Nice catch - thanks!

PR #765, opened 2016-06-23
Fix parameter name warning

twsouthwick:
There's a warning now due to a variable name I had in a recent PR that I hadn't seen. This fixes the warning.


tillig:
Whoops! Thanks!

Issue #755, opened 2016-06-07
IServiceProvider resolve differences


PleasantD:
I noticed a discrepancy between using the built-in `IServiceProvider` implementation for ASP.NET Core (RC2) and the autofac implementation using `builder.Populate(services)`.

Say I have a collection of service descriptions which contain multiple entries for `IStartupFilter` (which by default you will if you use IIS integration). The order of these is in the `IServiceCollection` is: `StartupA`, `StartupB`. 

Using the default `IServiceProvider` implementation, these are resolved in the same order.

Using `builder.Populate(services)` registers all of the services in the collection into the `BuilderContainer` in the same order. This is important so that later registrations hide earlier ones when resolving a single service. However when resolving all services it inverts the list. Resolving `IEnumerable<IStartupFilter>` now returns `StartupB`, `StartupA`.

Since the order of services matters for `IStartupFilter`, the middleware registered by these filters are running out of order. I can foresee this causing problems since `StartupA` is `AutoRequestServicesStartupFilter` which registers the middleware that initializes the per-request scope.

tillig:
We saw something similar with #695. I'm not sure if there's anything we can do about it, but it's worth a second look.

PleasantD:
A simple solution would be to have the `IServiceProvider` implementation reverse any resolution of `IEnumerable`. This at least makes autofac behave as expected by ASP.NET Core. However, this could still cause issues if registrations are made on the `ContainerBuilder` _before_ calling `builder.Populate(services)`.

I think a better pattern would be to create an extension method for `IServiceCollection` that returns a `ContainerBuilder`. Then the usage pattern shifts to:

```
public IServiceProvider ConfigureServices(IServiceCollection services)
{
    // Add services to IServiceCollection
    services.AddMvc();

    // Create autofac builder
    var builder = services.CreateAutofacBuilder();
      
    // Register against autofac
    builder.Register<MyService>().AsSelf();

    // could be replaced by something like --> return builder.BuildServiceProvider();
    var container = containerBuilder.Build();
    return container.Resolve<IServiceProvider>();
}

tillig:
[...]

PR #768, opened 2016-07-02
DI order spec conformance _lite_

aL3891:
Allright, here is a rough draft of what I was talking about in the discussion over at 
aspnet/DependencyInjection#416 I don't expect it to be merged in its current state but its ment to illustrate what I ment over in that other issue, as well as serve as a basis for discussion

So just to reiterate, This is my interpretations of @halter73 comments: 

> "Registration order" is probably a bit of a misnomer. It's really the order the service definitions occur in the IEnumerable<IServiceDescriptor> passed to container's Populate method. Populate is free to do whatever it wants with this metadata including calling registration methods in whatever order it wants.

With some assumptions on my part that  someone can hopefully confirm or deny:

> Does that mean that actual order given for resolving an IEnumeable<T> is given by the IEnumerable<IServiceDescriptor> passed to Populate?
Also, does it mean that order only has to be tracked for components added by the abstraction/Populate, not for components added via container specific apis?

In this PR I've assumed that the answers to those questions are "yes" and so this adapter would (barring bugs)

* Maintain order of components registered with the DI abstractions
* _Not_ maintain order of components registered with the autofac apis
* Have _unpredictable_ order (or possibly crash right now) for components whos implementations have been registered with a mix of autofac and DI abstraction apis.

So the question then is, would this satisfy the current spec? (or would this be an acceptable future spec?)

Again, though this works with a basic application, I havent tested it very much, its more to illustrate the idea at the moment. (i'll continue to test it, I just wanted to show you what I ment)

cc @tillig @alexmg @halter73 @davidfowl

### DIFF
--- a/src/Autofac.Extensions.DependencyInjection/AutofacServiceProvider.cs
+++ b/src/Autofac.Extensions.DependencyInjection/AutofacServiceProvider.cs
@@ -24,6 +24,10 @@
 // OTHER DEALINGS IN THE SOFTWARE.
 
 using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace Autofac.Extensions.DependencyInjection
@@ -31,20 +35,65 @@ namespace Autofac.Extensions.DependencyInjection
     internal class AutofacServiceProvider : IServiceProvider, ISupportRequiredService
     {
         private readonly IComponentContext _componentContext;
+        private List<OrderedServiceDescriptor> _originalDescriptors;
+        private HashSet<Type> _serviceTypes;
+        private MethodInfo _castMethodInfo;
 
-        public AutofacServiceProvider(IComponentContext componentContext)
+        public AutofacServiceProvider(IComponentContext componentContext, List<ServiceDescriptor> originalDescriptors)
         {
             _componentContext = componentContext;
+            _originalDescriptors = originalDescriptors.Select((d, i) => new OrderedServiceDescriptor
+            {
+                Index = i,
+                ImplementationInstance = d.ImplementationInstance,
+                ImplementationType = d.ImplementationType
+            }).ToList();
+            _serviceTypes = new HashSet<Type>(originalDescriptors.Select(d => d.ServiceType));
+            _castMethodInfo = typeof(Enumerable).GetTypeInfo().GetDeclaredMethod("Cast");
         }
 
         public object GetService(Type serviceType)
         {
-            return _componentContext.ResolveOptional(serviceType);
+            return GetOrdered(serviceType, _componentContext.ResolveOptional(serviceType));
         }
 
         public object GetRequiredService(Type serviceType)
         {
-            return _componentContext.Resolve(serviceType);
+            return GetOrdered(serviceType, _componentContext.Resolve(serviceType));
         }
+
+        private object GetOrdered(Type serviceType, object resolutionResult)
+        {
+            if (serviceType.Name == "IEnumerable`1" && _serviceTypes.Contains(serviceType.GenericTypeArguments[0]))
+            {
+                var actualServiceType = serviceType.GenericTypeArguments[0];
+                var resulutionResultEnumerable = ((IEnumerable)resolutionResult).Cast<object>();
+
+                resulutionResultEnumerable = resulutionResultEnumerable
+                    .Select(rr => new { ResolutionResult = rr, Index = _originalDescriptors.First(original => MatchedServiceDescriptor(rr, original)).Index })
+                    .OrderBy(rr => rr.Index)
+                    .Select(rr => rr.ResolutionResult).ToList();
+
+                var res = _castMethodInfo.MakeGenericMethod(actualServiceType).Invoke(null, new object[] { resulutionResultEnumerable });
+                return res;
+            }
+            else
+            {
+                return resolutionResult;
+            }
+        }
+
+        private static bool MatchedServiceDescriptor(object a, OrderedServiceDescriptor f)
+        {
+            return f.ImplementationInstance == a || (f.ImplementationInstance == null && a.GetType() == f.ImplementationType);
+        }
+    }
+
+    internal class OrderedServiceDescriptor
+    {
+        public int Index { get; set; }
+        public object ImplementationInstance { get; set; }
+        public Type ImplementationType { get; set; }
+
     }
 }
--- a/src/Autofac.Extensions.DependencyInjection/AutofacServiceProvider.cs
+++ b/src/Autofac.Extensions.DependencyInjection/AutofacServiceProvider.cs
@@ -88,12 +88,4 @@ namespace Autofac.Extensions.DependencyInjection
             return f.ImplementationInstance == a || (f.ImplementationInstance == null && a.GetType() == f.ImplementationType);
         }
     }
-
-    internal class OrderedServiceDescriptor
-    {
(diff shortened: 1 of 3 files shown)


aL3891:
Btw this could also be implemented using meta data added by the adapter in `Populate`. I chose not to go that way right now because of @tillig s comments: 

> Say it adds metadata. Now Autofac core - with every enumerable resolve - has to order by metadata that may or may not be there based on whether the registration was made in the ASP.NET abstraction or not... The ASP.NET abstraction then just bled into core Autofac and we added a perf hit to every list resolve op.

This is perhaps a more separated way of doing it, although its probably a bit of a hack. My gut says meta data in Populate is the "right" way, but i'm not really sure..

tillig:
Unpredictable order for things registered in a mix of APIs isn't acceptable. The stuff relying on order during _resolution_ doesn't know anything about _registration_.

Unfortunately we really need to see something that's _fully working_ in all use cases. We were able to hack something simple together that appeared to work on the surface, but the devil turns out to be in the details. It's sort of like async/await - once you start down the road, getting everything playing nicely together means async "infects" the code and gets deeper in than just the top level.

aL3891:
It doesn't have to be unpredictable, you could also have it ignore implementations added directly via autofac for service types that has already been added by the abstraction, or have them at the end in some deterministic way.

I'd say that if a developer configures those things in code using the abstraction, they will expect them to be in that order, but if they use some other more exotic way like registration source or assembly scanning, they don't care so much. 

After all, order is undefined in those cases so why would someone add components using those methods if they expect them to be ordered? I don't think it has to be an all or nothing approach to order, lets just care about order when the user cares, that is when they use the registration abstractions that guarantee order.

I agree that stuff relying on resolution does not know about registration, but I don't agree they have to with this approach, they just ask for an `IEnumerable<T>` and they get one. If registration was made in a way they'd get predictable order, then that's what they get, if it wasn't, they don't. It still _works_, we just can't guarantee order is some cases because there isn't any order :)

My point is that the developer knows what they want to have in order and when they don't care, the component consumers doesn't have to know.

aL3891:
And just to be clear, the order would only be unpredictable when asking for a `IEnumerable<T>`, not a `T` by itself.

src/Autofac.Extensions.DependencyInjection/AutofacServiceProvider.cs
```diff
@@ -24,27 +24,68 @@
...
+                ImplementationType = d.ImplementationType
+            }).ToList();
+            _serviceTypes = new HashSet<Type>(originalDescriptors.Select(d => d.ServiceType));
+            _castMethodInfo = typeof(Enumerable).GetTypeInfo().GetDeclaredMethod("Cast");
         }
 
         public object GetService(Type serviceType)
         {
-            return _componentContext.ResolveOptional(serviceType);
+            return GetOrdered(serviceType, _componentContext.ResolveOptional(serviceType));
```
halter73: It's not sufficient to just reorder the return value of `IServiceProvider.GetService`. Constructor injection has to work the same way.

In the past, when we required support for a fallback `IServiceProvider`, we implemented a [simple IRegistrationSource](https://github.com/aspnet/DependencyInjection/blob/1.0.0-beta1/src/Microsoft.Framework.DependencyInjection.Autofac/ChainedRegistrationSource.cs) so services from the fallback provider could be resolved both via `GetService` and constructor injection. (The irony is not lost on me btw...)

I was initially hoping that the fix might be as simple as iterating of the descriptors in reverse in `AutofacRegistration.Register`, but seeing how the service descriptors rely on multiple registration sources themselves (such as for open generics), I now understand that it's not.

alexmg:
I thought about tweaking the adaptor earlier and quickly dismissed the approach due to the ambiguity in behaviour depending on whether or not the adaptor was being used and where the registrations came from. I’ve already started on universal ordering of all components as it would appear the alternative is to be left out in the cold.

There are still more test cases I would like to write before considering the approach workable. I’ll track progress over on https://github.com/autofac/Autofac/issues/755.

src/Autofac.Extensions.DependencyInjection/AutofacServiceProvider.cs
aL3891: I thought this method was used for that as well, but maybe constructor injection doesn't go though the adapter?

aL3891:
Allright.. i still think that's making more work for yourself than is necessary in practice, especially since they now confirmed order is really only needed for components passed to populate. Constructor injection is one thing, but imo it would atleast be worth it to investigate and discuss this further. 

However seeing as you've closed this already i take it you're not interested in that?

alexmg:
@aL3891 Giving registrations made via the adaptor special treatment inside the core library is coupling it to the adaptor. It’s bad enough that the container has to behave in a specific way, having conditional logic just for the adaptor makes the situation even worse.

@halter73 Out of interest, do you require services registered via the `Populate` method to be in order of registration and at the **front** of the list of returned instances, or in order of registration and **anywhere** within the list of returned instances? In other words, if a user registered service appeared at the front of the list, but the services you registered were still in the correct order, would that cause you problems?

aL3891:
But there are no changes to the core library with this approach, no speacial treatment _inside_ the core library? I don't see how this is coupling the core library to the adaptor.. Even if the adaptor were to add some metadata, how would that be any different from the user adding metadata during regular registration? 

Universal ordering seems like a far more invasive approach from the perspective of the core library, at least to me.

src/Autofac.Extensions.DependencyInjection/AutofacServiceProvider.cs
aL3891: I've looked at the code for constructor injection, specifically `Microsoft.Extensions.Internal.ActivatorUtilities` and it does seem like constructor injection does in fact use `IServiceProvider.GetService()`? This also seems to be the case in practical tests, I do hit the breakpoints i expect when I do constructor injection of a `IEnumerable<T>` in a controller for example. Am I missing something else @halter73 ?

halter73: @aL3891 This works in a Controller, because controllers are activated using `ActivatorUtilities`. It breaks down if you have a transitive dependencies.

For example, if `MyController` injects `MyService`, and `MyService` injects `IEnumerable<AnotherService>` you should see the issue.

halter73:
@alexmg Given the choice, I think it would be nice if services registered via the `Populate` method are put at the front of this list if `Populate` is called before other explicit registrations and at the back of the list if called afterwards.

I understand this doesn't work for automatic registrations though. In that case, I would put the services registered via `Populate` at the front of this list. In the case of `IConfigureOptions<TOptions>`, framework code adds some initialization logic for `TOptions`, so if automatically registered `IConfigureOptions<TOptions>` services want `TOptions` to be initialized, it should be put later in the list in order to run later.

I wouldn't consider this critical though. If order can't be guaranteed for services not registered via `Populate`, the recommendation could be to register all `IConfigureOptions<TOptions>` services with the `IServiceCollection`.

src/Autofac.Extensions.DependencyInjection/AutofacServiceProvider.cs
aL3891: ooh i see.. thats a problem.. in that case you're already "inside" the ioc container

aL3891:
Deleted the branch since it doesnt cover constructor injection for transient components that take `IEnumerables<T>`

alexmg:
Thanks for trying anyway @aL3891 

This has been resolved in #755

PR #771, opened 2016-07-12
make AutofacServiceProvider disposable to allow WebHost to release resources

adamsitnik:
Hello everybody!

Today I was investigating some failing integration tests for my ASP.NET Core app.
I use TestHost, which creates w WebHost instance, which can be used as regular web app (send some post requests etc). Of course I dispose it once I am done using it. 

The  [current implementation of the WebHost.Dispose](
https://github.com/aspnet/Hosting/blob/e7b8c3f90a781a55658b4245571c5bf5dac5e56e/src/Microsoft.AspNetCore.Hosting/Internal/WebHost.cs#L211-#L217) is following:

```cs
public void Dispose()
{
	_logger?.Shutdown();
	_applicationLifetime.StopApplication();
	(_applicationServices as IDisposable)?.Dispose();
	_applicationLifetime.NotifyStopped();
}
```

So it assumes that custom implementation of `IServiceProvider` does not have to implement `IDisposable`.

The problem is that `AutofacServiceProvider` does not implement `IDisposable`, so the resources are never disposed. Hence my PR.

### DIFF
--- a/src/Autofac.Extensions.DependencyInjection/AutofacServiceProvider.cs
+++ b/src/Autofac.Extensions.DependencyInjection/AutofacServiceProvider.cs
@@ -24,11 +24,12 @@
 // OTHER DEALINGS IN THE SOFTWARE.
 
 using System;
+using Autofac.Util;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace Autofac.Extensions.DependencyInjection
 {
-    internal class AutofacServiceProvider : IServiceProvider, ISupportRequiredService
+    internal class AutofacServiceProvider : Disposable, IServiceProvider, ISupportRequiredService
     {
         private readonly IComponentContext _componentContext;
 
@@ -45,6 +46,16 @@ namespace Autofac.Extensions.DependencyInjection
         public object GetRequiredService(Type serviceType)
         {
             return _componentContext.Resolve(serviceType);
+        }
+
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                (_componentContext as IDisposable)?.Dispose();
+            }
+
+            base.Dispose(disposing);
         }
     }
 }


tillig:
Right now the container that you build is what owns the resources, not the service provider. In fact, the service provider is inside the container, not the other way around. Have you tried disposing the container instead of the service provider?

tillig:
Just to be clear, I'm talking about you, in your app, should try disposing the container you build during `ConfigureServices` rather than disposing of any `IServiceProvider`. I don't think we're going to change the Autofac implementation of `IServiceProvider` to be disposable because that object is, itself, _inside the container_.

adamsitnik:
@tillig sure, I don't mean to say that you have some kind of a bug or that I am right.

The thing that I observed is that this might be disposable, the test that failed for me for the first time assumes that [it can happen](https://github.com/aspnet/DependencyInjection/blob/6b94c82e0a5a12ad4316146f4478750c67caec7d/src/Microsoft.Extensions.DependencyInjection.Specification.Tests/DependencyInjectionSpecificationTests.cs#L448-#l454). For example LightInject [implements](https://github.com/seesharper/LightInject.Microsoft.DependencyInjection/blob/master/src/LightInject.Microsoft.DependencyInjection/LightInject.Microsoft.DependencyInjection.cs#L249-#L257) it as disposable

tillig:
Sure. Different backing containers will implement it in different ways. The requirement really is just `IServiceProvider` - you can't assume as you switch from container to container that it will necessarily _also_ be disposable.

There's actually a very, very long thread [going on right now in the Microsoft DI repo](https://github.com/aspnet/DependencyInjection/pull/416) about how many of the behaviors in the default `ServiceProvider` they provide are sort of _implicitly_ specifying things outside the actual interface `IServiceProvider` and how folks are just making the assumption that every container will implement `IServiceProvider` _and_ have all this extra stuff (like `IDisposable`) even though that's not really part of the implementation - because people are getting confused about "what's in and what's out."

For now, the answer with Autofac is to dispose the container rather than the application services. I'll update our docs accordingly so other folks won't run into the same issue. We do appreciate you dropping in and helping out - it's folks like you that keep us moving forward, so a sincere thanks to you.

adamsitnik:
> Just to be clear, I'm talking about you, in your app, should try disposing the container you build during ConfigureServices rather than disposing of any IServiceProvider

The thing is that it's not default/recommended way to do it. If you take a look at the [official docs](http://docs.autofac.org/en/latest/integration/aspnetcore.html) for ASP.NET Core you can see that they just build the container, and return the IServiceProvider resolved from the container.  

```cs
// Return the IServiceProvider resolved from the container.
return container.Resolve<IServiceProvider>();
```

The reference to the container is lost. Even if you keep it in some field, the `Startup.cs` does not give you any way to hook up into lifecycle of your app and call some kind of Dispose on it when it's being shut down. (previously was possible with `global.cs` or something like that). Current implementation tries to Dispose it from  WebHost.

tillig:
Those docs are ours (not ASP.NET Core's) - I wrote those. What I'm saying is I'll update those docs to include information on disposal:

> I'll update our docs accordingly so other folks won't run into the same issue.

adamsitnik:
> What I'm saying is I'll update those docs to include information on disposal:

Sure, I was writing my previous comment and did not see your latest response.

Anyway thanks for the discussion and the great product, which I really like to use - Autofac.

Btw. The solution builds up really quickly, really nice to work with!

tillig:
[Docs updated with information about disposal.](http://autofac.readthedocs.io/en/latest/integration/aspnetcore.html)

Issue #770, opened 2016-07-08
AsImplementedInterface() doesn't include itself when registering interfaces


enkafan:
I'm not sure if this is by design, but when registering MassTransit I ran into this. MassTransit has a factory that returns `IBusControl` that I had wired this up with `AsImplementedInterfaces`. To me surprise I couldn't resolve `IBusControl`. Simple fix for me was to use `AsSelf` too, but I didn't know if this was a bug or expected behavior.

Simple failing test:

```csharp
        IImplementedInterface SelfComponentFactory()
        {
            return new SelfComponent();
        }

        [Test]
        public void AsImplementedInterfaces_CanBeAppliedToInstanceRegistrationsOfInterfaces()
        {
            var builder = new ContainerBuilder();
            builder.Register(c => SelfComponentFactory()).AsImplementedInterfaces();
            var context = builder.Build();

            context.Resolve<IImplementedInterface>();
        }
```

the "fix" if this is indeed a bug is was just including the interface in the list of registrations for situations where an interface is being registered:

```csharp
        static Type[] GetImplementedInterfaces(Type type)
        {           
            var interfaces = type.GetInterfaces().Where(i => i != typeof(IDisposable));
            return type.IsInterface ? interfaces.Append(type).ToArray() : interfaces.ToArray();
        }
```

tillig:
Nice catch - thanks! We'll see about getting this in place for the 4.0 release.

enkafan:
Right on, I'm happy to submit a PR but I wasn't sure which branch to target

tillig:
Resolved by https://github.com/autofac/Autofac/pull/772

Thank you very much for the PR!

PR #772, opened 2016-07-12
Includes self on GetImplementedInterface calls to interface

enkafan:
For situations where an interface is passed to GetImplementedInterfaces
this adds the actual interface to the list of returned values.

This allows for situations such as registering the return value of a
factory method. Prior to this commit to resolve that type you'd have to
use AsSelf() rather than AsImplementedInterfaces().

Resolves issue #770


tillig:
Solid addition. Thanks for the assist! We appreciate it very much!

PR #783, opened 2016-08-16
Replace standard library meta package with only necessary packages

twsouthwick:
The .NET Standard 1.6 library package is a meta package that includes a
number of assemblies that are not needed in Autofac. This change
specifies only the assemblies needed to run autofac.

### DIFF
--- a/src/Autofac/project.json
+++ b/src/Autofac/project.json
@@ -29,8 +29,21 @@
     "net4.5": {},
     "netstandard1.1": {
       "dependencies": {
-        "NETStandard.Library": "1.6.0",
-        "System.ComponentModel": "4.0.1"
+        "System.Collections": "4.0.11",
+        "System.Collections.Concurrent": "4.0.12",
+        "System.ComponentModel": "4.0.1",
+        "System.Diagnostics.Debug": "4.0.11",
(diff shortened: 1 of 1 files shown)


tillig:
This pretty much reverts the change [from this commit](https://github.com/autofac/Autofac/commit/919c775fa961b4d13314fa1270676701a56f8730) that @alexmg intentionally made. I'm not sure the reason behind his commit, though; possibly just easier dependency management. @alexmg?

twsouthwick:
For libraries, the best practice is to trim unnecessary libraries (see https://docs.microsoft.com/en-us/dotnet/articles/core/deploying/reducing-dependencies). I hadn't seen that it had recently been switched; I'm interested to see what reasoning @alexmg used to replace with the meta package.

alexmg:
Thanks for taking a look @onovotny and setting things straight. :+1:

alexmg:
Also, thanks for the PR @twsouthwick. I know the intent was to follow best practice and help make things better! :+1:

terrajobst:
What @onovotny said is correct. @twsouthwick, we should update our guidance. Let's chat tomorrow.

PR #785, opened 2016-08-17
Optional string parameter causing resolution exception when AnyConcreteTypeNotAlreadyRegisteredSource added

### LINKED ISSUE #728: Optional string parameter causing resolution exception when AnyConcreteTypeNotAlreadyRegisteredSource added
From Google Group:

https://groups.google.com/forum/#!topic/autofac/lUZ6KAT7xXM

Hi there

I need to use the `AnyConcreteTypeNotAlreadyRegisteredSource`, to avoid having to explicitly register every concrete type we use.
Unfortunately, I'm having a real issue attempting to use it when we have classes with optional string constructor parameters.

e.g.

        public class TypeWithCtorParam : IInterface
        {
            public TypeWithCtorParam(string stringParam = "MyString") {}
        }

It throws an error

```
Autofac.Core.DependencyResolutionException: Cannot choose between multiple constructors with equal length 1 on type 'System.String'. Select the constructor explicitly, with the UsingConstructor() configuration method, when the component is registered.
```

I've attached an nunit test class repro-ing the issue, but basically, to cut it down to the bare bones

- If the class is registered AsSelf or against its interface, it is resolved fine

```
builder.RegisterType<TypeWithCtorParam>();
...
scope.Resolve<TypeWithCtorParam>();  // Fine. Yay!
```

- If the class is registered AsSelf and/or against its interface, and `AnyConcreteTypeNotAlreadyRegisteredSource` is used, it throws the error. I would expect this to not be a problem - I mean, it's already explicitly registered, surely it should be excluded from the registration source!

```
builder.RegisterType<TypeWithCtorParam>();
builder.RegisterSource(new AnyConcreteTypeNotAlreadyRegisteredSource());
...
scope.Resolve<TypeWithCtorParam>();  // Error! Boo!
```

- If the class is not explicity registered, and `AnyConcreteTypeNotAlreadyRegisteredSource` is used, it throws the error.

```
builder.RegisterSource(new AnyConcreteTypeNotAlreadyRegisteredSource());
...
scope.Resolve<TypeWithCtorParam>();  // Error! Boo!
```

I can exclude the types that aren't working, but it'd be nicer not to have to. Bug? Or am I missing something?

Thanks much

NickRusinov:
Fix #728

### DIFF
--- a/src/Autofac/Features/ResolveAnything/AnyConcreteTypeNotAlreadyRegisteredSource.cs
+++ b/src/Autofac/Features/ResolveAnything/AnyConcreteTypeNotAlreadyRegisteredSource.cs
@@ -74,6 +74,7 @@ namespace Autofac.Features.ResolveAnything
 
             var ts = service as TypedService;
             if (ts == null ||
+                ts.ServiceType == typeof(string) ||
                 !ts.ServiceType.GetTypeInfo().IsClass ||
                 ts.ServiceType.GetTypeInfo().IsSubclassOf(typeof(Delegate)) ||
                 ts.ServiceType.GetTypeInfo().IsAbstract ||
--- a/test/Autofac.Test/Features/ResolveAnything/ResolveAnythingTests.cs
+++ b/test/Autofac.Test/Features/ResolveAnything/ResolveAnythingTests.cs
@@ -136,6 +136,43 @@ namespace Autofac.Test.Features.ResolveAnything
             }
         }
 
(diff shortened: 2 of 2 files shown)


tillig:
I think omitting string here is a reasonable solution to the issue. The other primitive types are already value types and are caught by the `!type.IsClass` clause.

tillig:
Sorry for the delay, but thanks so much for the PR! We appreciate the help!

Issue #794, opened 2016-09-05
ParameterCompatibleWithTypeConstraint returns true instead of false when resolving open generic service


sebekz:
I have a simple case of an open generic type with type constraint, like:

```csharp
public class MultiTenantHandler<TRequest, TResponse> : IResponseHandler<TRequest, TResponse>
    where TResponse : IQueryable<IMultitenant> {}

public class MultiTenantOptionalHandler<TRequest, TResponse> : IResponseHandler<TRequest, TResponse>
    where TResponse : IQueryable<IMultitenantOptional> {}
```

https://gist.github.com/sebekz/4c658a5c7551ba5a2b3fd81488ea3ee7 holds an console app sample for this particular case. 

I have logged a StackOverflow question http://stackoverflow.com/questions/39291046/autofac-issues-when-resolving-generic-types-implementing-the-same-interface-nee which led to a hint of potential bug/issue in Autofac itself, therefore I am logging an issue here. 

The problem is, that when two (or more) open generic types are registered in the container

```csharp
builder.RegisterGeneric(typeof(MultiTenantHandler<,>))
    .As(typeof(IResponseHandler<,>))
    .SingleInstance();  
builder.RegisterGeneric(typeof(MultiTenantOptionalHandler<,>))
    .As(typeof(IResponseHandler<,>))
    .SingleInstance();
```

an attempt to resolve it will fail **n-1** times and succeed afterwards, where n equals number of implementations of `IResponseHandler<,>`. So if we had one more `class MultiTenantAnotherHandler` with yet another `where` type constraint, it would fail to `.Resolve()` 2 times and succeed each time after that. For a full issue description, please refer to StackOverflow question mentioned above.

Further investigation has led to a call stack:

```csharp
IsCompatibleWithGenericParameterConstraints()
  ParameterCompatibleWithTypeConstraint()
    IsAssignableFrom()
```

`IsCompatibleWithGenericParameterConstraints` method call ensures that the type parameters are compatible with the given type constraints. It uses the method `ParameterCompatibleWithTypeConstraint ` to check for that and that metod uses `IsAssignableFrom` in its implementation. So far so good, since `IsAssignableFrom` works as expected, but in the case it returns `false` for all its checks `ParameterCompatibleWithTypeConstraint` checks if it can create a generic type with the base type of the constraint and the generic argument of the parameter. 

```csharp
Traverse.Across(parameter, p => p.GetTypeInfo().BaseType)
[...]

PR #796, opened 2016-09-13
ParameterCompatibleWithTypeConstraint returns true instead of false

sebekz:
Addresses #794


tillig:
Looks nice! Grabbin' it now.

Issue #769, opened 2016-07-04
Large amount of garbage with deep dependency tree


ReneZwanenburg:
Our ASP application has a few deeply nested dependency hierarchies, with the size of a tree in the order of thousands of objects. When resolving such a dependency, Autofac 3.5.2 generates a lot of garbage: loading a  representative webpage results in ~ 50 MB worth of allocations coming from Autofac, less than one percent of the allocated bytes are for the resulting instances. The top allocations are ConcatIterators and various lambda's. It's not leaking, everything gets cleaned up nicely at the next GC cycle, but the high amount of allocations puts pressure on the GC and causes our heaps to be significantly larger than necessary.

There might be some improvements possible. I haven't looked at Autofac's code close enough to say for certain, but these kinds of changes often involve replacing linq and lambdas with something a bit less nice-looking around the heaviest allocating places. If you're potentially open to changes of this nature I'm willing to work on this.

alexmg:
We are definitely open to such improvements. I would work with the develop branch that has the current 4.0 RC version if you would like to look further into the issue.

alexmg:
I've pushed Autofac 4.3.0 to NuGet and look forward to taking this for a spin at the office tomorrow.

nblumhardt:
@ReneZwanenburg any chance you can try out the 4.3.0 build and let us know if it's improved matters? Cheers!

PR #815, opened 2016-12-20
BenchmarkDotNet harness

nblumhardt:
I noticed #769 and thought it would be an interesting ticket to look into. Exploring it would benefit from an objective way to measure and compare performance between iterations.

In the Serilog project we've had some success using [BenchmarkDotNet](https://github.com/dotnet/BenchmarkDotNet) for this, hence the current PR.

The PR adds a flag to the build script:

```powershell
./Build.ps1 -Bench
```

When this is supplied, the benchmarks in `/bench/Autofac.Benchmarks` will be run, and results like the table below written into the `/BenchmarkDotNetArtifacts` folder.

```ini

Host Process Environment Information:
BenchmarkDotNet.Core=v0.9.9.0
OS=Windows
Processor=?, ProcessorCount=8
Frequency=2533305 ticks, Resolution=394.7413 ns, Timer=TSC
CLR=CORE, Arch=64-bit ? [RyuJIT]
GC=Concurrent Workstation
dotnet cli version: 1.0.0-preview2-003121

Type=RootContainerResolveBenchmark  Mode=Throughput  

```
|              Method |        Median |     StdDev | Scaled | Scaled-SD |
|-------------------- |-------------- |----------- |------- |---------- |
|         OperatorNew |     4.1694 ns |  0.0913 ns |   1.00 |      0.00 |
| NonSharedReflection | 1,027.0683 ns | 66.2231 ns | 248.35 |     16.42 |
|   NonSharedDelegate |   446.3929 ns | 44.5342 ns | 104.32 |     10.67 |
|              Shared |   412.4813 ns |  8.0122 ns |  99.43 |      2.87 |

The results don't currently include allocation info, but this should be possible by upgrading to the latest BenchmarkDotNet, which depends on .NET Core 1.1. I have tried it in vain, but I'm sure we'll get there eventually... Until then, the raw performance is still a pretty good indicator of allocation behaviour.

The included benchmarks are only really a sample - some more thought needs to go into what should be tested.

While grappling with the build, I took the opportunity to use multi-targeting in lieu of the old _Autofac.Net46.Test_ assembly - if I overlooked something important about the distinct assembly please let me know.

So, thoughts?


tillig:
The .NET 4.6 assembly was put there in https://github.com/autofac/Autofac/commit/2206b8c5fe05bf276ff55e3650cfc1054ea37db7 to work around a challenge where xunit won't run tests against different frameworks. It looks like since that was done [xunit added multi-framework test execution](https://xunit.github.io/docs/getting-started-dotnet-core.html), so I guess we don't need that separation anymore.

bench/Autofac.Benchmarks/RootContainerResolveBenchmark.cs
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+#pragma warning disable SA1402, SA1502
+
+    internal class NonSharedDelegate { }
+
+    internal class NonSharedReflection { }
+
+    internal class Shared { }
+}
```
tillig: Do we need to re-enable the warnings after this?

bench/Autofac.Benchmarks/RootContainerResolveBenchmark.cs
```diff
@@ -0,0 +1,62 @@
...
+            builder.RegisterType<NonSharedReflection>();
+            builder.Register(c => new NonSharedDelegate());
+            _container = builder.Build();
+        }
+
+        [Benchmark(Baseline = true)]
+        public void OperatorNew()
+        {
+            var instance = new NonSharedDelegate();
+            GC.KeepAlive(instance);
```
tillig: Being unfamiliar with BenchmarkDotNet - is there a reason you keep them alive? Or is that so we don't get hit with GC overhead time after a method is executed?

bench/Autofac.Benchmarks/Harness.cs
```diff
@@ -0,0 +1,39 @@
...
+using Xunit;
+
+namespace Autofac.Benchmarks
+{
+    public class Harness
+    {
+        [Fact]
+        public void RootContainerResolve()
+        {
+            BenchmarkRunner.Run<RootContainerResolveBenchmark>();
```
tillig: Is there a way to change the output location of the reports to the `artifacts` folder? (This might be [this issue](https://github.com/dotnet/BenchmarkDotNet/issues/252).)

bench/Autofac.Benchmarks/RootContainerResolveBenchmark.cs
nblumhardt: Might as well; will do 👍 thanks

bench/Autofac.Benchmarks/RootContainerResolveBenchmark.cs
nblumhardt: It may be overly defensive; inserting `GC.KeepAlive()` is just a marker for the compiler that ensures the object will still be around at that point. E.g. in the case above, the constructor call could in theory be completely optimised away, otherwise.

bench/Autofac.Benchmarks/Harness.cs
nblumhardt: Will do it via PowerShell 👍

nblumhardt:
Alrighty, updated and rebased, should be closer to the mark now.

tillig:
Looks great! I hope we can add some more benchmarks to help determine what's going well and what's not.

nblumhardt:
Great! I think the door's open - I'll try to get a few more written soon. 👍

Issue #816, opened 2016-12-21
Generics constrained with generic parameters do not resolve


zacateras:
The issue can be described by the following failing XUnit test:

```cs
public class ConstrainedWithGenericParameter<T1, T2>
    where T1 : T2
{
}

[Fact]
public void CanResolveComponentWhenGenericParameterIsConstrainedWithOtherGenericParameter()
{
    var builder = new ContainerBuilder();

    builder.RegisterGeneric(typeof(ConstrainedWithGenericParameter<,>));

    var container = builder.Build();

    Assert.True(container.IsRegistered<ConstrainedWithGenericParameter<int, object>>());
}
```

It seems that generic parameter constraint compatibility [check](https://github.com/autofac/Autofac/blob/41044d7d1a4fa277c628021537d5a12016137c3b/src/Autofac/Util/TypeExtensions.cs#L77) always returns true for generic constraints stub types like `{T1}`, `{T2}`.

I submitted solution to the issue in #819.

PR #819, opened 2016-12-22
Proposed solution for #816

zacateras:
#816


alexmg:
Looks like it does the trick. Thanks @zacateras!

PR #821, opened 2017-01-03
Do less work and create less garbage when instantiating components via reflection

nblumhardt:
This partially addresses #769 by removing allocations  and unnecessary work from the hot path of reflection-based component activation.

```ini
Host Process Environment Information:
BenchmarkDotNet.Core=v0.9.9.0
OS=Windows
Processor=?, ProcessorCount=8
Frequency=2533308 ticks, Resolution=394.7408 ns, Timer=TSC
CLR=CORE, Arch=64-bit ? [RyuJIT]
GC=Concurrent Workstation
dotnet cli version: 1.0.0-preview2-003121
Type=DeepGraphResolveBenchmark  Mode=Throughput
```

### Before

|  Method |     Median |    StdDev |
|-------- |----------- |---------- |
| Resolve | 38.4004 us | 0.8659 us |

### After

|  Method |     Median |    StdDev |
|-------- |----------- |---------- |
| Resolve | 17.9659 us | 0.1156 us |

### DIFF
--- a/bench/Autofac.Benchmarks/DeepGraphResolveBenchmark.cs
+++ b/bench/Autofac.Benchmarks/DeepGraphResolveBenchmark.cs
@@ -1,0 +1,67 @@
+﻿using BenchmarkDotNet.Attributes;
+using System;
+
+namespace Autofac.Benchmarks
+{
+    /// <summary>
+    /// Tests the performance of retrieving a (reasonably) deeply-nested object graph.
+    /// </summary>
+    public class DeepGraphResolveBenchmark
+    {
+        private IContainer _container;
+
+        [Setup]
+        public void Setup()
+        {
+            var builder = new ContainerBuilder();
+            builder.RegisterType<A>();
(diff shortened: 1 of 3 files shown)


alexmg:
Great stuff @nblumhardt. It seems the `ReflectionActivator` is always right in the middle of performance issues. 😃

nblumhardt:
Thanks Alex. Even though the design makes some allocations unavoidable (without public API changes, anyway) I think there are still some opportunities to improve elsewhere. I'm looking forward to digging in some more when I get some time :-)

adamsitnik:
@nblumhardt BenchmarkDotNet v0.10.1 has new MemoryDiagnoser, I recommend you to use it if you want to measure the number of allocated bytes as part of your benchmarking process.

nblumhardt:
@adamsitnik thanks. I checked it out when I put this together, but it'd mean moving the project forward to the new `dotnet` SDK version so I couldn't get it in with this change in the time I had.

Issue #529, opened 2014-05-30
RegisterDecorator does not play nicely when decorating items registered with .As<IFoo>() or with .AsImplementedInterfacs()


JonathanMaccollum:
Let’s say at container builder “build time” we wanted to use configuration to turn on or off a particular decorator.

Test structures:

```
    public interface IFoo
    {
    }
    public class Foo:IFoo
    {

    }
    public class FooDecorator : IFoo
    {
        private readonly IFoo _instance;

        public FooDecorator(IFoo instance)
        {
            if (instance == null) throw new ArgumentNullException("instance");
            _instance = instance;
        }
    }
```

Prior to using RegisterDecorator<IFoo> to accomplish this the code would look like this:

```
        bool turnDecoratorOn = true; // value from configuration
        var cb = new ContainerBuilder();
        cb.RegisterType<Foo>()
            .As<IFoo>() // this is the default implementation of IFoo to be used of the decoration is not enabled.
            .Named<IFoo>("Foo"); 
        if (turnDecoratorOn)
        {
            cb.RegisterType<FooDecorator>()
                .WithParameter(
                    (pi, ctx) => pi.ParameterType == typeof (IFoo),
                    (pi, ctx) => ctx.ResolveNamed<IFoo>("Foo")) // resolve the IFoo parameter with the one registered with name of 'Foo'
                .As<IFoo>();
        }
        using (var container = cb.Build())
        {
            var foo = container.Resolve<IFoo>();
            Assert.IsInstanceOfType(foo, typeof(FooDecorator));
        }
```

And if the config value had it turned off it would test successfully like this:

```
        bool turnDecoratorOn = false;
        var cb = new ContainerBuilder();
        cb.RegisterType<Foo>()
            .As<IFoo>() // this is the default implementation of IFoo to be used of the decoration is not enabled.
            .Named<IFoo>("Foo"); 
        if (turnDecoratorOn)
        {
            cb.RegisterType<FooDecorator>()
                .WithParameter(
                    (pi, ctx) => pi.ParameterType == typeof(IFoo),
                    (pi, ctx) => ctx.ResolveNamed<IFoo>("Foo"))
                .As<IFoo>();
        }
        using (var container = cb.Build())
        {
            var foo = container.Resolve<IFoo>();
            Assert.IsInstanceOfType(foo, typeof(Foo));
        }
```

[...]

PR #826, opened 2017-01-12
Add a failing test for #529

alexandrnikitin:
#529

### DIFF
--- a/test/Autofac.Test/Features/LightweightAdapters/LightweightAdapterRegistrationExtensionsTests.cs
+++ b/test/Autofac.Test/Features/LightweightAdapters/LightweightAdapterRegistrationExtensionsTests.cs
@@ -81,6 +81,30 @@ namespace Autofac.Test.Features.LightweightAdapters
             }
         }
 
+        public class DecoratingServiceThatHasDefaultImplementation
+        {
+            readonly IContainer _container;
+
+            public DecoratingServiceThatHasDefaultImplementation()
+            {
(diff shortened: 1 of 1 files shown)


test/Autofac.Test/Features/LightweightAdapters/LightweightAdapterRegistrationExtensionsTests.cs
```diff
@@ -81,6 +81,34 @@ public void AdaptedMetadataIsPassed()
...
+        {
+            readonly IContainer _container;
+
+            public DecoratingServiceThatHasDefaultImplementation()
+            {
+                const string from = "from";
+                var builder = new ContainerBuilder();
+
+                // Implementer1 is the default implementation for IService
+                builder.RegisterType<Implementer1>().As<IService>();
```
tillig: This is two different services; does it still fail if it's one service?

`builder.RegisterType<Implementer1>().As<IService>().Named<IService>(from);`

If it does fail, can we update it to do that? Or potentially create a second test that uses a single service?

alexandrnikitin: @tillig yes, it still fails. I wanted to keep it more closer to real-world usage. Updated the commit.

tillig:
Thanks! We appreciate the help. Now to see if we can fix the issue... ;)

PR #841, opened 2017-04-01
Fix crash on IOS when the linker is enabled

tzachshabtay:
On IOS, when the mono linker is enabled (and it’s enabled by default and recommended by Xamarin) it strips out all parts of the framework which are not in use. The linker does not support reflection, however, therefore it does not recognize the call to 'GetRuntimeMethod(nameof(Convert.ChangeType)' (which is called in ConstructorParameterBinding -> ConvertPrimitiveType) as using the 'Convert.ChangeType' method and strips it away unless it's explicitly used somewhere else, crashing Autofac if that method is called (null reference exception as ‘GetRuntimeMethod’ returns null).
Added an explicit call to Convert.ChangeType to let the linker know that we need this method and with that making sure Autofac works on IOS.

### DIFF
--- a/src/Autofac/Core/Activators/Reflection/ConstructorParameterBinding.cs
+++ b/src/Autofac/Core/Activators/Reflection/ConstructorParameterBinding.cs
@@ -62,9 +62,9 @@ namespace Autofac.Core.Activators.Reflection
         {
             // On IOS, when the mono linker is enabled (and it's enabled by default) it strips out all parts of
             // the framework which are not in use. The linker does not support reflection, however, therefore it
-            // does not recognize the call to 'GetRuntimeMethod(nameof(Convert.ChangeType)' (which is called from 
-            // a method in this class) as using the 'Convert.ChangeType' method and strips it away unless it's  
-            // explicitly used somewhere else, crashing Autofac if that method is called. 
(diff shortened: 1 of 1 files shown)


tillig:
The build for this fails code analysis so we can't accept it as-is.

Even if it passed I'm thinking about whether this is the _right way_ to solve it. I understand the problem but...

1. This feels like a compiler/linker bug in the Xamarin stuff. `nameof` should be enough for things to work. Have you filed an issue with them? I can't imagine we're the only library out there using `nameof`.
2. If we do have to patch around the issue, I'm wondering if doing it in the static constructor of various classes encountering the problem is the right place. Is there a more central location we could do it? (I don't have an answer for this.)
3. How do we know this actually fixes the overall issue? It may come up elsewhere in a feature you're not using. It'd be nice to not have to keep plugging holes as they get found. I'm curious if there's some sort of analysis we can run to find these.

We've run into something similar with some UWP stuff because it doesn't necessarily enable reflection everywhere by default. The workaround there is that UWP allows for a manifest to be specified with the list of things you need to reflect over.

Is there an additional compiler or linker directive that can be specified to make this work? (I also don't know the answer to that.)

It may be worth filling an issue so we can discuss the right solution before working more on code. It'd be good to see a repro and mess around with it before pulling something I can't verify.

tillig:
Closed for now so we can iterate over design in #842.

PR #845, opened 2017-04-11
Exceptions in CurrentScopeEnding handler stops disposal

### LINKED ISSUE #844: Exceptions in CurrentScopeEnding handler stops disposal
We recently discovered a memory leak in our application(s).  It was traced back to expensive `IDisposable` instances not being disposed when container scopes were ended.

The issue was that our framework code was doing:
`container.CurrentScopeEnding += (o, eventArgs) => HttpContext.Current.Items.Remove(DependencyResolverKey);`

`HttContext.Current` property was `null` depending on how the request handling pipeline was constructed (for that particular application).  If null, a `NullReferenceException` would get thrown.  The exception happened after the request was returned and went undetected, but had an interesting side-effect, namely that it prevented the disposal of objects in the now ended container scope.

We changed the event handler to be delicate:

```
        static void AttemptToClearContainer(object sender, LifetimeScopeEndingEventArgs e)
        {
            
            try
            {
                HttpContext.Current?.Items.Remove(DependencyResolverKey);
            }
            catch
            {
                // while I don't expect any exceptions, if any exception is raised from the 
                // container.CurrentScopeEnding callback, crucial behavior is skipped 
                // (e.g. disposing all the container's IDisposables)
            }
        }
```

While this worked for us, it seems like a misbehaved handler shouldn't be able to defeat disposal.

chris-peterson:
As container lifetime scopes end, always dispose objects

### DIFF
--- a/src/Autofac/Core/Lifetime/LifetimeScope.cs
+++ b/src/Autofac/Core/Lifetime/LifetimeScope.cs
@@ -308,9 +308,15 @@ namespace Autofac.Core.Lifetime
             if (disposing)
             {
                 var handler = CurrentScopeEnding;
-                handler?.Invoke(this, new LifetimeScopeEndingEventArgs(this));
 
-                Disposer.Dispose();
+                try
+                {
+                    handler?.Invoke(this, new LifetimeScopeEndingEventArgs(this));
+                }
+                finally
+                {
+                    Disposer.Dispose();
+                }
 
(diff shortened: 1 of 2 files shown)


chris-peterson:
Fixes #844

tillig:
This sounds reasonable and looks great. Thanks for the PR!

Issue #867, opened 2017-09-14
Update build for .NET Core SDK 2.0.0


alexBraidwood:
While working on adding this for a personal project and feeling great about making a pull request, I noticed there's an extension method collision for .net standard 2.0. 

[SequenceExtensions.Append](https://github.com/autofac/Autofac/blob/develop/src/Autofac/Util/SequenceExtensions.cs#L56) (no docs?) is colliding with a shiny new gift from Microsoft: System.Linq.Enumerable.Append.

I thought this is likely used across projects and was worried about pressing forward with a breaking name change. I added conditional compilation symbols and a quick google of the repo didn't reveal any additional usages, so I carried on. 

Here's the commit, currently: https://github.com/alexBraidwood/Autofac/commit/7fe8febc5eb431ccd231625fd2ae916badf8706a

tillig:
We're currently [building against .NET SDK 1.1.0](https://github.com/autofac/Autofac/blob/5ea93974fac4399cebc6d6d0a5e75e818a4a6b88/build.ps1#L14) but when the time comes this will be a good reminder of one of the things we may need to address.

No docs on that method because it's `internal` - you shouldn't be consuming it from outside the system and we haven't kept docs on every internal and private method.

tillig:
Note we'll still try to target a lower version of `netstandard` than 2.0 to maintain a larger compatibility, but we could definitely update the SDK version.

alexBraidwood:
Ah, I see. I can just update the name of the extension method, then.

Forgot to note: I left `netstandard` 1.1.0 in-tact with the commit. All three seemed to build out just fine with the powershell script. I can certainly put up a pull request for this or wait until this is on the roadmap. Also, `Array.Empty<T>` was another changeover for `net46` and `netstandard2.0`.

hawkerm:
The Autofac website talks about .NET Core, but will it just run on .NET Standard or is that work still being done?

tillig:
@hawkerm Have you tried it or looked at the code?

alexBraidwood:
@hawkerm .NET Standard <= 1.1 should work fine with the current nuget package. 

[...]

PR #868, opened 2017-09-15
Updated Autofac to include netcore / netstandard 2.0 build

alexBraidwood:
- Changed name of SequenceExtensions.Append to AppendItem due to
collision with netstandard2 linq extension Append.
- Added conditional compilation flag for Array.Empty on netstandard2


tillig:
I was able to take the SDK updates (basically), so it builds against 2.1.2 now and the Roslyn analyzers all work in VS 15.5.2 (current as of today).

However, `netstandard1.1` is fully compatible with projects that target `netstandard2.0`. Is there something that isn't working that you were trying to address with that change?

tillig:
Oh, just found issue #867 is associated here. Will check over there.

PR #871, opened 2017-09-26
Fix for Issue 870 

starkcolin:
Error building container when class inherits from the same generic interface  more than once


alexmg:
Nice catch. Thanks for the PR.

PR #876, opened 2017-10-12
Improve implementation of DefaultValueParameter to fix several bugs.

edwardmeng:
According to the issues reported in dotnet/corefx repo, there are several bugs should be fixed before the .net framework updated. Here is the list:
dotnet/corefx#17943
dotnet/corefx#12338
dotnet/corefx#11797


tillig:
Wow, that is some insidious stuff! Thanks for the help in catching it!

tillig:
Released with 4.6.2.

PR #881, opened 2017-11-19
Deadlock timeout (remove simple lock which produce the threat of dead…

bnayae:
When resolving on multiple thread it can lead to a deadlock state when the resolved type's constructor is waiting on other resolve.
You can check the test in order to reproduce the scenario.
It may not being best practice to create this kind of code in the first place, but shit happen and sometime you do have constraint like legacy source code etc.
The fix enable to add timeout via new ContainerBuilder().WithTimeout(TimeSpan.FromMilliseconds(100)).

Internally the fix replace the simple lock with a convenient wrapper of semaphore which have timeout and support async operations.

### DIFF
--- a/src/Autofac/Core/Container.cs
+++ b/src/Autofac/Core/Container.cs
@@ -26,6 +26,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Threading.Tasks;
 using Autofac.Core.Activators.Delegate;
 using Autofac.Core.Lifetime;
 using Autofac.Core.Registration;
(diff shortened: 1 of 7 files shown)


tillig:
It's generally better to open an issue to discuss the problem you're seeing prior to going straight to PR. We might be able to help.

In this case, the general rule based on the language is that constructors themselves shouldn't be async or call async code. [Stephen Cleary gives some alternatives here](https://blog.stephencleary.com/2013/01/async-oop-2-constructors.html) and in his book. Construction should be fast and simple. I'd recommend looking at these solutions to avoid deadlock in construction instead.

I also have a bad feeling that if we introduced something like this it'd be taken as a hint that "it's okay to do expensive things like database queries during construction." I remember a bit ago similar arguments about mock frameworks that allowed you to mock private variables - "the ability is there so it endorses that behavior!" Not true, but also not an argument I'm interested in having repeatedly once people start using it.

Anyway, thanks for the PR, but I think we'll probably not be adding enforcement for timing out a constructor call at this time.

bnayae:
It's true that people should use the library according to its guidance, yet people some time miss it or don't aware of it. in those case, I believe, deadlock is really a bad choice.
Timeout can be use in order to reflect the issue and the error description of the timeout could suggest the best-practice.

Issue #733, opened 2016-04-11
DefaultConstructorFinder should throw an exception if type being inspected has no public constructors.


orhtej2:
Currently when registering a type that has no public constructors (namely, a type that only has a private/protected default constructor) DefaultConstructorFinder will return an empty array of constructor which in turn will throw a generic, unhelpful DependencyResolutionException in default IConstructorSelector implementations.
However, in my opinion, if a postcondinion is implemented on IConstructorFinder that forces it to return a no-empty array it would be easier to identify the problem.

dyronald:
We can have DefaultConstructorFinder::GetDefaultPublicConstructors throw an exception if the constructor array is empty. This would move the exception earlier to when the container is built.

Is this is what is wanted here?

orhtej2:
Yes, as long as name of the type and the reason will be in one exception I think this will help a lot.

tillig:
Fix released in [v4.7.0](https://www.nuget.org/packages/Autofac/4.7.0).

PR #895, opened 2018-02-01
Building a container now fails if a type without a public constructor is registered

dyronald:
Fix for Issue #733


tillig:
Nice! Thanks! I have a few comments I'll put inline.

src/Autofac/Core/Activators/Reflection/DefaultConstructorFinder.cs
```diff
@@ -73,8 +73,15 @@ public ConstructorInfo[] FindConstructors(Type targetType)
 
         private static ConstructorInfo[] GetDefaultPublicConstructors(Type type)
         {
-            return DefaultPublicConstructorsCache.GetOrAdd(
+            var retval = DefaultPublicConstructorsCache.GetOrAdd(
                 type, t => t.GetTypeInfo().DeclaredConstructors.Where(c => c.IsPublic).ToArray());
+
+            if (retval.Count() == 0)
```
tillig: `retval` should be an array so just use `.Length` directly.

src/Autofac/Core/Activators/Reflection/NoPublicConstructorException.cs
```diff
@@ -0,0 +1,59 @@
...
+// WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+// OTHER DEALINGS IN THE SOFTWARE.
+
+using System;
+using System.Globalization;
+
+namespace Autofac.Core.Activators.Reflection
+{
+    public class NoPublicConstructorException : Exception
```
tillig: Can we call this `NoConstructorsFoundException` ? Reason being, if someone implements their own constructor finder, it may be that they don't require _public_ constructors. This would allow the exception to work for all cases but still communicate the issue pretty clearly.

src/Autofac/Core/Activators/Reflection/NoPublicConstructorException.cs
```diff
@@ -0,0 +1,59 @@
...
+            : base(FormatMessage(type))
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="NoPublicConstructorException"/> class.
+        /// </summary>
+        /// <param name="type">The type with no public constructor but was registered.</param>
+        /// <param name="innerException">The inner exception.</param>
+        public NoPublicConstructorException(Type type, Exception innerException)
```
tillig: We probably need a couple of additional constructors:

- Type and message
- Type, message, and inner exception

That way a custom message could be put in to override the default.

src/Autofac/Core/Activators/Reflection/NoPublicConstructorException.cs
```diff
@@ -0,0 +1,59 @@
...
+
+namespace Autofac.Core.Activators.Reflection
+{
+    public class NoPublicConstructorException : Exception
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="NoPublicConstructorException"/> class.
+        /// </summary>
+        /// <param name="type">The type with no public constructor but was registered.</param>
+        public NoPublicConstructorException(Type type)
```
tillig: Add a public property with the Type so anyone catching the exception can pull that info back out.

src/Autofac/Core/Activators/Reflection/NoPublicConstructorExceptionResources.resx
```diff
@@ -0,0 +1,123 @@
...
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="Message" xml:space="preserve">
+    <value>Type {0} has no public constructors</value>
```
tillig: `No accessible constructors were found for the type `{0}'.`

The differences being...
- Slightly more generic default but still communicates the issue.
- Type name in single quotes and period at the end of the sentence.

tillig:
Looks great! Thanks so much!

tillig:
Fix released in [v4.7.0](https://www.nuget.org/packages/Autofac/4.7.0).

Issue #624, opened 2015-02-24
IStartable components should be started in dependency order


OZCMRDM:
Service C depends on Service A and B via constructor dependency. All services implement IStartable. In Start() for C, A and B have not yet been started. Logically, IStartable components should be started in dependency order (A,B, then C).

jchapuis:
I have the same issue here, thanks

aschaible:
Same here.  Seems strange that this is not supported.

tillig:
We do accept pull requests... 😉

aschaible:
Yes apologies for the tone - just that Autofac has so much circulation, I was just surprised.  TYVM!

tillig:
No worries. It's just that there are only two project owners for core Autofac as well as all of the supported integrations (I think there are 23 integration and feature libraries now?), so we end up having to prioritize our limited time on the most heavily used / heavily affecting issues. We just ate a little over a full year trying to nail down .NET Core support without much chance to address anything else.

`IStartable` isn't one of the most common things, and `IStartable` _that relies on other `IStartable`_ makes it even more edge case. Combine that with the fact that the solution isn't super quick to implement and this sort of thing gets prioritized down the list a bit.

From a design perspective, you sort of have to walk the entire resolution chain for every `IStartable` and build the dependency chain; then determine if any one `IStartable` depends on any other `IStartable`; then resolve/execute accordingly, which _again_ walks that dependency tree, this time actually doing the resolution/instantiation.

It becomes a challenge when you start factoring in additional things like lambda registrations - you can resolve things using `context.Resolve<T>()` inside a lambda registration; what if one of those uses or provides an `IStartable` (or both)? You don't really know what happens inside that lambda until you execute it.

Sooooo... there's a non-trivial amount of thought that has to go into implementing the feature and making sure it works in all the crazy edge cases in which folks are expecting it to "just work," and given the other issues that are pending out there, we just haven't gotten to this one.

**We truly would be open to a pull request to help with this.** If you're interested, drop some thoughts on the design in here so we can talk about it **before you spend a lot of time coding**. We should make sure all the use cases are thought through and can be addressed before we can take code to solve the issue.

ggrnd0:
Why `IStartable` needed if `IRegistrationBuilder.OnActivated, etc...` methods exist and provide high control of component lifetime?

How this problems solved for `IRegistrationBuilder.OnActivated, etc...`? Is them solved?

tillig:
[...]

PR #904, opened 2018-04-07
#624 Start startable components in dependency order

babrekel:
Hi guys,

I wanted to take a stab at issue #624 to address a practical need in some large systems where `IStartable` is heavily utilized.
To give some context on our usage of `IStartable` without going into specifics -- distributed systems often have recurring tasks, metadata retrieval operations, things that need to happen when booting up a service or application. These operations usually connect to external systems.
We have unit tests that build containers using the `IgnoreStartableComponents` option. This allows us to ensure that the container is correctly defined at compile time, while eliminating the need for mocking or overriding components just to verify that the container builds.

I added some basic unit tests to verify that this works. The unit test fails without the corresponding code changes in the Autofac code. I'll also verify this in a larger scale (1000+ class container, with 100+ `IStartable` components).

Please let me know if you see any issues, my main concerns are:

- Adding code to the `InstanceLookup` class, which is a hot code path
- If Start() throws, should it be wrapped in a DependencyResolutionException? 

Thanks,

Bas

### DIFF
--- a/src/Autofac/Builder/MetadataKeys.cs
+++ b/src/Autofac/Builder/MetadataKeys.cs
@@ -30,5 +30,7 @@ namespace Autofac.Builder
         internal const string RegistrationOrderMetadataKey = "__RegistrationOrder";
 
         internal const string AutoActivated = "__AutoActivated";
+
+        internal const string StartOnActivatePropertyKey = "__StartOnActivate";
     }
 }
--- a/src/Autofac/ContainerBuilder.cs
+++ b/src/Autofac/ContainerBuilder.cs
@@ -156,39 +156,50 @@ namespace Autofac
 
         private static void StartStartableComponents(IComponentContext componentContext)
         {
-            // We track which registrations have already been auto-activated by adding
-            // a metadata value. If the value is present, we won't re-activate. This helps
-            // in the container update situation.
             const string started = MetadataKeys.AutoActivated;
-            object meta;
+            const string startOnActivate = MetadataKeys.StartOnActivatePropertyKey;
 
-            foreach (var startable in componentContext.ComponentRegistry.RegistrationsFor(new TypedService(typeof(IStartable))).Where(r => !r.Metadata.TryGetValue(started, out meta)))
+            var componentRegistry = componentContext.ComponentRegistry;
+            try
             {
-                try
+                componentRegistry.Properties[startOnActivate] = true;
+
+                // We track which registrations have already been auto-activated by adding
+                // a metadata value. If the value is present, we won't re-activate. This helps
+                // in the container update situation.
+                object meta;
+
+                foreach (var startable in componentRegistry.RegistrationsFor(new TypedService(typeof(IStartable))).Where(r => !r.Metadata.TryGetValue(started, out meta)))
                 {
-                    var instance = (IStartable)componentContext.ResolveComponent(startable, Enumerable.Empty<Parameter>());
-                    instance.Start();
+                    try
+                    {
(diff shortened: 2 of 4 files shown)


tillig:
This is pretty nice, thanks for the PR. Let me take a bit to digest what's going on and walk through various scenarios in my head to see if there's a reason we couldn't do it this way or if there's some alternative being missed; though on the face of it, this looks pretty good to me.

tillig:
I used the benchmark tests for `DeepGraphResolve` and `RootContainerResolve` to test the before and after effects of the PR. Overall the effect seems negligible - I didn't notice any real difference before and after; but I can't really run the two different branches side-by-side in the same run that easily, so there's definitely some environmental flux to account for. If anything, we're talking roughly 10ns difference.

One thing I plan on trying - the difference between TryGetValue and ContainsKey. Since the values of the metadata in the dictionary are only used when checking presence rather than actually getting and using the value, I'm curious if one measurably outperforms the other. If so, we might be able to gain a little back by switching. I'll run some tests now.

babrekel:
Makes sense, `ContainsKey` is slightly more lightweight than `TryGetValue`. Another approach I was thinking about is to add a boolean property onto the `IComponentContext`, but it would require either public visiblity or a cast, which is pretty ugly.

tillig:
Yeah, we don't want to change `IComponentContext` if we can avoid it. I'm still of the mind that what's here in this PR is good to go, just looking at little microperf things now because, like you said, it's a little bit in the hot path.

tillig:
OK, I verified that `ContainsKey` does pretty consistently outperform `TryGetValue` when not needing the value, by just a little. The success case (value found in the dictionary) is a more noticeable difference than the failure case (value not found). I think if you can switch the uses of `TryGetValue` to `ContainsKey` in the spots where the value isn't being used we can call this good.

tillig:
One thing I thought of that I'm not sure if is an issue or not right now: If you have an `IStartable` component that depends on something `AutoActivate` (or vice versa) - are we going to get any duplicate resolution ops going on at container build time? Maybe that's something we can add as a separate issue; this gets us a lot as it is. Just didn't know if you'd thought of that. Like:

```csharp
public class Startable : IStartable
{
  public Startable(AutoActivated activated) { ... }
}

public class AutoActivated { }

// and registration like

builder.RegisterType<Startable>();
builder.RegisterType<AutoActivated>().AutoActivate();
```

Ostensibly the AutoActivated class should start first and it should only be resolved _once_. Of course, generally I'd gather folks should register those things as singletons so it wouldn't matter anyway.

But like I said, we can add that as an issue after the fact. I'm guessing we already have that problem in the existing code.

tillig:
Oh, the Roslyn analyzer thing is an interesting idea. I like how xunit started shipping with that. I'll have to see if that's something we can do. Maybe I'll add an enhancement issue around that so I don't forget.

Anyway, this looks great, thanks for the PR here. I'll merge it in and will try to file a sort of "known issue" issue about the potential multiple instantiation thing prior to cutting a release that includes this.

babrekel:
Great, I'd be willing to build a prototype analyzer to play around with. Would you like me to send a PR to update the documentation as well?

tillig:
That'd be super, sure. And thanks again!

tillig:
Oh, looks like [the docs]() already have a note about not registering the startable items as anything other than SingleInstance so I guess we can call the known issue part good.

> It isn’t recommended that startable components implement other services, or be registered as anything other than SingleInstance().

PR #919, opened 2018-05-30
Memory traffic optimization

Belorus:
Hi!

I work on mobile games written in C#/Xamarin, and start time is a bit critical there.
To improve it a bit, i slightly optimized Autofac memory allocations during registration phase to get rid of some simple ones:
1. foreach over interface (boxing of struct iterator)
2. allocation of iterators inside LINQ calls
3. allocation of delegates for LINQ calls 

Also i introduced a bit more fine-grained control of default services.

In my codebase with 1600 registrations results are following:
4.8.1: 122K allocations (3.45MB), 80K collected (2.2MB)
4.8.1+PR: 82K allocations (2.25MB), 46K collected (1.20MB)

Speedup was about ~10% on .NET 4.6 (No sample, made 10 runs with logging of Stopwatch and compared best timings). 

We use this fork for over a year, and i though someone else could find it useful.

TO BE UPDATED: I'll add some runtime optimizations (Resolve-time)

### DIFF
--- a/src/Autofac/Autofac.csproj
+++ b/src/Autofac/Autofac.csproj
@@ -4,7 +4,7 @@
     <Description>Autofac is an IoC container for Microsoft .NET. It manages the dependencies between classes so that applications stay easy to change as they grow in size and complexity.</Description>
     <VersionPrefix>4.8.1</VersionPrefix>
     <TargetFrameworks>netstandard1.1;net45</TargetFrameworks>
-    <NoWarn>$(NoWarn);CS1591</NoWarn>
+    <NoWarn>$(NoWarn);CS1591;CA1819</NoWarn>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <AssemblyName>Autofac</AssemblyName>
--- a/src/Autofac/Builder/ContainerBuildOptions.cs
+++ b/src/Autofac/Builder/ContainerBuildOptions.cs
@@ -40,12 +40,6 @@ namespace Autofac.Builder
         None = 0,
 
         /// <summary>
-        /// Prevents inclusion of standard modules like support for
-        /// relationship types including <see cref="IEnumerable{T}"/> etc.
-        /// </summary>
-        ExcludeDefaultModules = 2,
-
-        /// <summary>
         /// Does not call <see cref="IStartable.Start"/> on components implementing
         /// this interface (useful for module testing.)
         /// </summary>
--- a/src/Autofac/Builder/IRegistrationBuilder.cs
+++ b/src/Autofac/Builder/IRegistrationBuilder.cs
@@ -278,5 +278,19 @@ namespace Autofac.Builder
         /// </param>
         /// <returns>A registration builder allowing further configuration of the component.</returns>
         IRegistrationBuilder<TLimit, TActivatorData, TRegistrationStyle> WithMetadata<TMetadata>(Action<MetadataConfiguration<TMetadata>> configurationAction);
+
+        /// <summary>
+        /// Configure the services that the component will provide.
+        /// </summary>
+        /// <param name="service">Service types to expose.</param>
+        /// <returns>A registration builder allowing further configuration of the component.</returns>
+        IRegistrationBuilder<TLimit, TActivatorData, TRegistrationStyle> As(Type service);
+
+        /// <summary>
+        /// Configure the services that the component will provide.
+        /// </summary>
+        /// <param name="service">Services to expose.</param>
+        /// <returns>A registration builder allowing further configuration of the component.</returns>
+        IRegistrationBuilder<TLimit, TActivatorData, TRegistrationStyle> As(Service service);
     }
 }
--- a/src/Autofac/Builder/ReflectionActivatorData.cs
+++ b/src/Autofac/Builder/ReflectionActivatorData.cs
@@ -108,11 +108,11 @@ namespace Autofac.Builder
         /// <summary>
         /// Gets the explicitly bound constructor parameters.
         /// </summary>
-        public IList<Parameter> ConfiguredParameters { get; } = new List<Parameter>();
+        public List<Parameter> ConfiguredParameters { get; } = new List<Parameter>();
 
         /// <summary>
         /// Gets the explicitly bound properties.
         /// </summary>
-        public IList<Parameter> ConfiguredProperties { get; } = new List<Parameter>();
+        public List<Parameter> ConfiguredProperties { get; } = new List<Parameter>();
     }
 }
--- a/src/Autofac/Builder/RegistrationBuilder.cs
+++ b/src/Autofac/Builder/RegistrationBuilder.cs
@@ -133,7 +133,7 @@ namespace Autofac.Builder
                 builder.RegistrationStyle.Id,
                 builder.RegistrationData,
                 builder.ActivatorData.Activator,
-                builder.RegistrationData.Services,
+                builder.RegistrationData.Services.ToArray(),
                 builder.RegistrationStyle.Target);
         }
 
@@ -149,7 +149,7 @@ namespace Autofac.Builder
             Guid id,
             RegistrationData data,
             IInstanceActivator activator,
-            IEnumerable<Service> services)
+            Service[] services)
         {
             return CreateRegistration(id, data, activator, services, null);
         }
@@ -170,7 +170,7 @@ namespace Autofac.Builder
             Guid id,
             RegistrationData data,
             IInstanceActivator activator,
-            IEnumerable<Service> services,
+            Service[] services,
             IComponentRegistration target)
         {
             if (activator == null) throw new ArgumentNullException(nameof(activator));
@@ -179,11 +179,15 @@ namespace Autofac.Builder
             var limitType = activator.LimitType;
             if (limitType != typeof(object))
             {
-                foreach (var ts in services.OfType<IServiceWithType>())
+                foreach (var ts in services)
                 {
-                    if (!ts.ServiceType.GetTypeInfo().IsAssignableFrom(limitType.GetTypeInfo()))
+                    var asServiceWithType = ts as IServiceWithType;
+                    if (asServiceWithType != null)
                     {
-                        throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, RegistrationBuilderResources.ComponentDoesNotSupportService, limitType, ts));
+                        if (!asServiceWithType.ServiceType.GetTypeInfo().IsAssignableFrom(limitType.GetTypeInfo()))
+                        {
+                            throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, RegistrationBuilderResources.ComponentDoesNotSupportService, limitType, ts));
+                        }
                     }
                 }
             }
--- a/src/Autofac/Builder/RegistrationBuilder{TLimit,TActivatorData,TRegistrationStyle}.cs
+++ b/src/Autofac/Builder/RegistrationBuilder{TLimit,TActivatorData,TRegistrationStyle}.cs
@@ -241,11 +241,24 @@ namespace Autofac.Builder
         /// <returns>A registration builder allowing further configuration of the component.</returns>
         public IRegistrationBuilder<TLimit, TActivatorData, TRegistrationStyle> As(params Type[] services)
         {
-            return As(services.Select(t =>
-                t.FullName != null
-                ? new TypedService(t)
-                : new TypedService(t.GetGenericTypeDefinition()))
-                .Cast<Service>().ToArray());
+            Service[] argArray = new Service[services.Length];
+
+            for (int i = 0; i < services.Length; i++)
+            {
+                argArray[i] = new TypedService(services[i]);
+            }
+
+            return As(argArray);
+        }
+
+        /// <summary>
+        /// Configure the services that the component will provide.
+        /// </summary>
+        /// <param name="service">Service types to expose.</param>
+        /// <returns>A registration builder allowing further configuration of the component.</returns>
+        public IRegistrationBuilder<TLimit, TActivatorData, TRegistrationStyle> As(Type service)
+        {
+            return As(new TypedService(service));
         }
 
         /// <summary>
@@ -258,6 +271,20 @@ namespace Autofac.Builder
             if (services == null) throw new ArgumentNullException(nameof(services));
 
             RegistrationData.AddServices(services);
+
+            return this;
+        }
+
+        /// <summary>
+        /// Configure the services that the component will provide.
+        /// </summary>
+        /// <param name="service">Services to expose.</param>
+        /// <returns>A registration builder allowing further configuration of the component.</returns>
+        public IRegistrationBuilder<TLimit, TActivatorData, TRegistrationStyle> As(Service service)
+        {
+            if (service == null) throw new ArgumentNullException(nameof(service));
+
+            RegistrationData.AddService(service);
 
             return this;
         }
--- a/src/Autofac/Builder/RegistrationBuilder{TLimit,TActivatorData,TRegistrationStyle}.cs
+++ b/src/Autofac/Builder/RegistrationBuilder{TLimit,TActivatorData,TRegistrationStyle}.cs
@@ -245,7 +245,10 @@ namespace Autofac.Builder
 
             for (int i = 0; i < services.Length; i++)
             {
-                argArray[i] = new TypedService(services[i]);
+                if (services[i].FullName != null)
+                    argArray[i] = new TypedService(services[i]);
+                else
+                    argArray[i] = new TypedService(services[i].GetGenericTypeDefinition());
             }
 
             return As(argArray);
--- a/src/Autofac/Builder/RegistrationData.cs
+++ b/src/Autofac/Builder/RegistrationData.cs
@@ -41,7 +41,7 @@ namespace Autofac.Builder
         private bool _defaultServiceOverridden;
         private Service _defaultService;
 
-        private readonly ICollection<Service> _services = new HashSet<Service>();
+        private readonly HashSet<Service> _services = new HashSet<Service>();
 
         private IComponentLifetime _lifetime = new CurrentScopeLifetime();
 
@@ -147,17 +147,17 @@ namespace Autofac.Builder
         /// <summary>
         /// Gets the handlers for the Preparing event.
         /// </summary>
-        public ICollection<EventHandler<PreparingEventArgs>> PreparingHandlers { get; } = new List<EventHandler<PreparingEventArgs>>();
+        public List<EventHandler<PreparingEventArgs>> PreparingHandlers { get; } = new List<EventHandler<PreparingEventArgs>>();
 
         /// <summary>
         /// Gets the handlers for the Activating event.
         /// </summary>
-        public ICollection<EventHandler<ActivatingEventArgs<object>>> ActivatingHandlers { get; } = new List<EventHandler<ActivatingEventArgs<object>>>();
+        public List<EventHandler<ActivatingEventArgs<object>>> ActivatingHandlers { get; } = new List<EventHandler<ActivatingEventArgs<object>>>();
 
         /// <summary>
         /// Gets the handlers for the Activated event.
         /// </summary>
-        public ICollection<EventHandler<ActivatedEventArgs<object>>> ActivatedHandlers { get; } = new List<EventHandler<ActivatedEventArgs<object>>>();
+        public List<EventHandler<ActivatedEventArgs<object>>> ActivatedHandlers { get; } = new List<EventHandler<ActivatedEventArgs<object>>>();
 
         /// <summary>
         /// Copies the contents of another RegistrationData object into this one.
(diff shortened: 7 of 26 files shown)


alexmg:
Thanks for the PR @Belorus. There are some great optimizations here. I agree with @tillig that we need to figure out what can come straight in and what might have to wait for a `5.0` release when breaking changes are to be expected. There are a few places that are technically breaking changes because a class is `public` but the likelihood of someone using it directly are very low. 

I've wondered for a while now how much time would be saved at resolve time (particularly the first call that initializes the `ServiceRegistrationInfo`) by removing `IRegistrationSources` that are not required.

We have a benchmarks project that might be a good way to quantify just how much benefit some of the changes bring. In the past BenchmarkDotNet had the `MemoryDiagnoser` enabled by default but that was changed in a later version so the GC and memory allocation information isn't currently available but could be added back easily.

I will try to take a first pass over the review and see if things can be flagged for immediate inclusion. I'm very excited to see a PR that is focused on performance.

alexmg:
I just noticed that the PR targets `master` directly as well. They would need to come into the `develop` branch first. Let's wait and see how the slicing goes before doing anything though.

src/Autofac/Builder/ContainerBuildOptions.cs
```diff
@@ -39,12 +39,6 @@ public enum ContainerBuildOptions
         /// </summary>
         None = 0,
 
-        /// <summary>
-        /// Prevents inclusion of standard modules like support for
-        /// relationship types including <see cref="IEnumerable{T}"/> etc.
-        /// </summary>
-        ExcludeDefaultModules = 2,
```
alexmg: This is a breaking change. We could add an overload to `Build` takes the `DefaultServiceFlags` and passes through `None`.

src/Autofac/Builder/IRegistrationBuilder.cs
```diff
@@ -278,5 +278,19 @@ public interface IRegistrationBuilder<out TLimit, out TActivatorData, out TRegis
         /// </param>
         /// <returns>A registration builder allowing further configuration of the component.</returns>
         IRegistrationBuilder<TLimit, TActivatorData, TRegistrationStyle> WithMetadata<TMetadata>(Action<MetadataConfiguration<TMetadata>> configurationAction);
+
+        /// <summary>
+        /// Configure the services that the component will provide.
+        /// </summary>
+        /// <param name="service">Service types to expose.</param>
+        /// <returns>A registration builder allowing further configuration of the component.</returns>
+        IRegistrationBuilder<TLimit, TActivatorData, TRegistrationStyle> As(Type service);
```
alexmg: Looks like these are to prevent the need to call the `params` method with the array allocation.

Belorus:
Thanks for the review guys! 
Then let me split this PR into 2 parts. One nonbreaking and second breaking. And I'll rebase it on develop.

src/Autofac/Builder/ReflectionActivatorData.cs
```diff
@@ -108,11 +108,11 @@ public IConstructorSelector ConstructorSelector
         /// <summary>
         /// Gets the explicitly bound constructor parameters.
         /// </summary>
-        public IList<Parameter> ConfiguredParameters { get; } = new List<Parameter>();
+        public List<Parameter> ConfiguredParameters { get; } = new List<Parameter>();
```
alexmg: I don't think this would cause any issues from the API perspective.

src/Autofac/Builder/RegistrationBuilder.cs
```diff
@@ -149,7 +149,7 @@ public static class RegistrationBuilder
             Guid id,
             RegistrationData data,
             IInstanceActivator activator,
-            IEnumerable<Service> services)
+            Service[] services)
```
alexmg: These are definitely breaking changes but I don't think these methods would be used directly by consumers very often.

alexmg:
I'm doing a quick pass through now and adding a few comments that might help. There are a few changes that are technically breaking that I would probably be fine with.

src/Autofac/Builder/RegistrationBuilder.cs
```diff
@@ -179,11 +179,15 @@ public static class RegistrationBuilder
             var limitType = activator.LimitType;
             if (limitType != typeof(object))
             {
-                foreach (var ts in services.OfType<IServiceWithType>())
+                foreach (var ts in services)
```
alexmg: All good.

src/Autofac/Builder/RegistrationBuilder{TLimit,TActivatorData,TRegistrationStyle}.cs
```diff
@@ -241,11 +241,27 @@ public RegistrationBuilder(Service defaultService, TActivatorData activatorData,
         /// <returns>A registration builder allowing further configuration of the component.</returns>
         public IRegistrationBuilder<TLimit, TActivatorData, TRegistrationStyle> As(params Type[] services)
         {
-            return As(services.Select(t =>
-                t.FullName != null
-                ? new TypedService(t)
-                : new TypedService(t.GetGenericTypeDefinition()))
-                .Cast<Service>().ToArray());
+            Service[] argArray = new Service[services.Length];
```
alexmg: Nice

src/Autofac/Builder/RegistrationBuilder{TLimit,TActivatorData,TRegistrationStyle}.cs
```diff
@@ -262,6 +278,20 @@ public RegistrationBuilder(Service defaultService, TActivatorData activatorData,
             return this;
         }
 
+        /// <summary>
+        /// Configure the services that the component will provide.
+        /// </summary>
+        /// <param name="service">Services to expose.</param>
+        /// <returns>A registration builder allowing further configuration of the component.</returns>
+        public IRegistrationBuilder<TLimit, TActivatorData, TRegistrationStyle> As(Service service)
```
alexmg: No problem here. I think the `AddService` method has only been used by internal registration sources but has always been part of the public API.

src/Autofac/ContainerBuilder.cs
```diff
@@ -139,11 +140,11 @@ public ContainerBuilder RegisterBuildCallback(Action<IContainer> buildCallback)
         /// first create the container, then call Update() on the builder.
         /// </remarks>
         /// <returns>A new container with the configured component registrations.</returns>
-        public IContainer Build(ContainerBuildOptions options = ContainerBuildOptions.None)
+        public IContainer Build(ContainerBuildOptions options = ContainerBuildOptions.None, DefaultServiceFlags flags = DefaultServiceFlags.All)
```
alexmg: This is where a second overload to `Build` might prevent breaking changes by mapping `ExcludeDefaultModules` to `DefaultServiceFlags.None`.

src/Autofac/ContainerBuilder.cs
```diff
@@ -234,28 +235,89 @@ private void Build(IComponentRegistry componentRegistry, bool excludeDefaultModu
...
         }
 
         private List<Action<IContainer>> GetBuildCallbacks()
         {
             return (List<Action<IContainer>>)Properties[BuildCallbackPropertyKey];
         }
     }
+
+    [Flags]
+    public enum DefaultServiceFlags
```
alexmg: No problem with the concept but could do with a better name. Perhaps something like `FeatureToggles`.

src/Autofac/Core/Activators/Reflection/ConstructorParameterBinding.cs
```diff
@@ -66,7 +66,7 @@ public class ConstructorParameterBinding
         /// <param name="context">Context in which to construct instance.</param>
         public ConstructorParameterBinding(
             ConstructorInfo ci,
-            IEnumerable<Parameter> availableParameters,
+            Parameter[] availableParameters,
```
alexmg: A breaking change but I don't see this class being called directly so probably fine.

src/Autofac/Core/Activators/Reflection/ReflectionActivator.cs
```diff
@@ -70,7 +70,14 @@ public class ReflectionActivator : InstanceActivator, IInstanceActivator
             ConstructorFinder = constructorFinder;
             ConstructorSelector = constructorSelector;
             _configuredProperties = configuredProperties.ToArray();
-            _defaultParameters = configuredParameters.Concat(new Parameter[] { new AutowiringParameter(), new DefaultValueParameter() }).ToArray();
+
+            _defaultParameters = new Parameter[configuredParameters.Count + 2];
```
alexmg: Not as readable but much more efficient.

src/Autofac/Core/Activators/Reflection/ReflectionActivator.cs
```diff
@@ -128,10 +135,25 @@ public object ActivateInstance(IComponentContext context, IEnumerable<Parameter>
 
         private ConstructorParameterBinding[] GetValidConstructorBindings(IComponentContext context, IEnumerable<Parameter> parameters)
         {
-            // Most often, there will be no `parameters` and/or no `_defaultParameters`; in both of those cases we can avoid allocating.
-            var prioritisedParameters = parameters.Any() ?
-                (_defaultParameters.Length == 0 ? parameters : parameters.Concat(_defaultParameters)) :
-                _defaultParameters;
+            // Most often, there will be no `parameters`we can avoid allocating.
+            Parameter[] prioritisedParameters;
```
alexmg: Looks good

src/Autofac/Core/IComponentRegistration.cs
```diff
@@ -63,7 +63,7 @@ public interface IComponentRegistration : IDisposable
         /// <summary>
         /// Gets the services provided by the component.
         /// </summary>
-        IEnumerable<Service> Services { get; }
+        Service[] Services { get; }
```
alexmg: I think this should be alright.

src/Autofac/Core/Registration/ComponentRegistration.cs
```diff
@@ -56,20 +56,21 @@ public class ComponentRegistration : Disposable, IComponentRegistration
             IComponentLifetime lifetime,
             InstanceSharing sharing,
             InstanceOwnership ownership,
-            IEnumerable<Service> services,
+            Service[] services,
```
alexmg: Breaking change but also probably not something called directly too often.

src/Autofac/Core/Registration/ComponentRegistry.cs
```diff
@@ -162,12 +162,17 @@ public void Register(IComponentRegistration registration, bool preserveDefaults)
 
         private void UpdateInitialisedAdapters(IComponentRegistration registration)
         {
-            var adapterServices = _serviceInfo
-                .Where(si => si.Value.ShouldRecalculateAdaptersOn(registration))
-                .Select(si => si.Key)
-                .ToArray();
+            List<Service> adapterServices = new List<Service>();
```
alexmg: Looks good

src/Autofac/Core/Registration/ServiceRegistrationInfo.cs
```diff
@@ -88,10 +88,16 @@ public IEnumerable<IComponentRegistration> Implementations
             get
             {
                 RequiresInitialization();
-                return Enumerable
-                    .Reverse(_defaultImplementations)
-                    .Concat(_sourceImplementations)
-                    .Concat(_preserveDefaultImplementations);
+
+                var resultingCollection = Enumerable.Reverse(_defaultImplementations);
```
alexmg: Everything in this file looks good.

src/Autofac/Core/Resolving/CircularDependencyDetector.cs
```diff
@@ -61,8 +61,11 @@ public static void CheckForCircularDependency(IComponentRegistration registratio
                 throw new DependencyResolutionException(string.Format(CultureInfo.CurrentCulture, CircularDependencyDetectorResources.MaxDepthExceeded, registration));
 
             // Checks for circular dependency
-            if (activationStack.Any(a => a.ComponentRegistration == registration))
-                throw new DependencyResolutionException(string.Format(CultureInfo.CurrentCulture, CircularDependencyDetectorResources.CircularDependency, CreateDependencyGraphTo(registration, activationStack)));
+            foreach (var a in activationStack)
```
alexmg: Nice

src/Autofac/Features/OpenGenerics/OpenGenericDecoratorRegistrationSource.cs
```diff
@@ -77,13 +77,19 @@ public IEnumerable<IComponentRegistration> RegistrationsFor(Service service, Fun
...
 
-        private static IEnumerable<Parameter> AddDecoratedComponentParameter(Type decoratedParameterType, IComponentRegistration decoratedComponent, IEnumerable<Parameter> configuredParameters)
+        private static IList<Parameter> AddDecoratedComponentParameter(Type decoratedParameterType, IComponentRegistration decoratedComponent, IList<Parameter> configuredParameters)
         {
             var parameter = new ResolvedParameter(
                 (pi, c) => pi.ParameterType == decoratedParameterType,
                 (pi, c) => c.ResolveComponent(decoratedComponent, Enumerable.Empty<Parameter>()));
 
-            return new[] { parameter }.Concat(configuredParameters);
+            var resultArray = new Parameter[configuredParameters.Count + 1];
```
alexmg: No problem

alexmg:
I've added some comments in a review @tillig. Would be interested to see if you agree on some of the changes that are technically breaking but very unlikely to have been used directly.

src/Autofac/Autofac.csproj
```diff
@@ -4,7 +4,7 @@
     <Description>Autofac is an IoC container for Microsoft .NET. It manages the dependencies between classes so that applications stay easy to change as they grow in size and complexity.</Description>
     <VersionPrefix>4.8.1</VersionPrefix>
     <TargetFrameworks>netstandard1.1;net45</TargetFrameworks>
-    <NoWarn>$(NoWarn);CS1591</NoWarn>
+    <NoWarn>$(NoWarn);CS1591;CA1819</NoWarn>
```
tillig: [CA1819 indicates that public/protected properties shouldn't return arrays.](https://msdn.microsoft.com/en-us/library/0fss9skc.aspx) The reason they note is that if the property doesn't return a _copy_ of the array then people can (and likely will) start using that property as an indexed property to start poking data into places they possibly shouldn't.

I'm not a big stickler for stuff like this, but I feel like there's a decent point here - where we may have public properties (like in `IComponentRegistration`) that were returning `IEnumerable<T>` and the switch to `T[]` enables people to do things they maybe shouldn't be doing. I've read too many Raymond Chen posts where people write programs using APIs that were never intended for a particular purpose, but now it's all locked in.

I'd recommend not ignoring this and, instead, being very thoughtful about switching away from an interface and using `[SuppressMessage]` to pinpoint the specific choices and document why that choice was made.

From a larger perspective, moving from `IList<T>` to `List<T>`, from `IEnumerable<T>` to `T[]`... that stuff may limit our ability to change underlying implementations. For example, if we need to store things in a `SortedList<T>` or start using `Span<T>` in certain places, switching from an interface to a concrete type makes that implementation detail a breaking change.

src/Autofac/Builder/ContainerBuildOptions.cs
tillig: Agreed, an overload would be better.

src/Autofac/Builder/IRegistrationBuilder.cs
tillig: Adding this is technically a breaking change. Super low risk, but breaking. Curious if adding this would be a problem during an upgrade where custom extension methods (that people have written) indirectly call this and the compiler will get confused as to which method it should be calling. We had some problems before adding optional parameters to an existing public API method where it was seen as breaking in odd and unfortunate ways.

src/Autofac/Builder/ReflectionActivatorData.cs
tillig: Agreed, likely not a big deal. Unclear how much this stuff gets used, if ever, outside of here.

src/Autofac/Builder/RegistrationBuilder.cs
tillig: Agreed. Pretty low risk.

src/Autofac/Builder/RegistrationData.cs
```diff
@@ -147,17 +147,17 @@ public IComponentLifetime Lifetime
         /// <summary>
```
tillig: The switch from interface to concrete implementation here is one of the areas I'm a little concerned about. Exposing the set of handlers like this makes any underlying improvement a breaking change.

src/Autofac/ContainerBuilder.cs
tillig: It appears to me that the `DefaultServiceFlags` (or `FeatureToggles` or `ContainerFeatures` or whatever) is an attempt at a more granular way to include or exclude default features... in which case, yeah, an overload where the current `ContainerBuildOptions` maps to some set of features would be better.

src/Autofac/ContainerBuilder.cs
tillig: Or `ContainerFeatures`.

If this is intended to be a more granular way to control what gets built into the container, I'd recommend also including `StartableComponents` as a flag so it can do everything `ContainerBuildOptions` can do and more.

tillig: It should also be in its own file, not attached to the bottom of `ContainerBuilder.cs`.

src/Autofac/Builder/ReflectionActivatorData.cs
tillig: I'm trying to put this together with the change from `Autofac.Core.Activators.Reflection.ReflectionActivator`. It looks like the change here wants to expose these as `List<T>`. These appear to be primarily used when instantiating a `ReflectionActivator`. The change in `ReflectionActivator` is to stop taking an `IEnumerable<T>` and take an `IList<T>` instead.

Could we leave this alone (as `IList<T>`) and _only_ make the change in `ReflectionActivator`? As it stands there's still going to be some implicit casting going on, which I'm guessing is the point of this.

tillig:
I took some time and brought in some, but not all, of the changes here. From a public interface perspective, I really am concerned that using arrays in public properties may open doors to people shooting themselves in the foot by trying to change values in the array. Like, "I know, let's swap the order of this and that because reasons!" It sounds silly, but we do see a lot of this sort of thing come through on StackOverflow and other venues.

I'll mark the things I brought in.

src/Autofac/Builder/RegistrationBuilder.cs
tillig: I brought all the changes from this file into the `develop` branch.

src/Autofac/Builder/RegistrationBuilder{TLimit,TActivatorData,TRegistrationStyle}.cs
tillig: I brought all the changes from this file into the `develop` branch.

src/Autofac/Core/Registration/ServiceRegistrationInfo.cs
tillig: I brought all the changes from this file into the `develop` branch.

src/Autofac/Core/Resolving/CircularDependencyDetector.cs
tillig: I brought all the changes from this file into the `develop` branch.

src/Autofac/Features/OpenGenerics/OpenGenericRegistrationSource.cs
```diff
@@ -60,7 +60,7 @@ public IEnumerable<IComponentRegistration> RegistrationsFor(Service service, Fun
             if (registrationAccessor == null) throw new ArgumentNullException(nameof(registrationAccessor));
 
             Type constructedImplementationType;
-            IEnumerable<Service> services;
+            Service[] services;
```
tillig: I brought all the changes from this file into the `develop` branch.

src/Autofac/Features/OpenGenerics/OpenGenericServiceBinder.cs
```diff
@@ -41,7 +41,7 @@ internal static class OpenGenericServiceBinder
             IEnumerable<Service> configuredOpenGenericServices,
             Type openGenericImplementationType,
             out Type constructedImplementationType,
-            out IEnumerable<Service> constructedServices)
+            out Service[] constructedServices)
```
tillig: I brought all the changes from this file into the `develop` branch.

src/Autofac/Features/OpenGenerics/OpenGenericDecoratorRegistrationSource.cs
tillig: I brought all the changes from this file into the `develop` branch.

tillig:
I think there are still things we may be able to pull in here for the Autofac 5 release, but I'm not sure we can really pull it in as-is anymore. Given that, I'm going to close the PR and tag @autofac/autofac so we can refer to it later when doing additional refactoring.

PR #940, opened 2018-10-28
Registration source registration consistent with module registration

eugbaranov:
`RegisterSource` API is not as convenient as those for `RegisterModule`.
I've added some features to bring it on par with module registrations:

1. Generic overload:
`
builder.RegisterSource<SomeSource>();
`
2. Chaining:
`
builder.RegisterSource<SourceA>().RegisterSource<SourceB>();
`

### DIFF
--- a/test/Autofac.Test/Core/Registration/SourceRegistrarTests.cs
+++ b/test/Autofac.Test/Core/Registration/SourceRegistrarTests.cs
@@ -1,0 +1,50 @@
+﻿using System;
+using System.Collections.Generic;
+using Autofac.Core.Registration;
+using Autofac.Test.Scenarios.RegistrationSources;
+using Xunit;
+
+namespace Autofac.Test.Core.Registration
+{
+    public sealed class SourceRegistrarTests
+    {
+        private static readonly object O1 = new object();
+        private static readonly object O2 = new object();
+
+        [Fact]
+        public void Ctor_RequiresContainerBuilder()
+        {
+            Assert.Throws<ArgumentNullException>(() => new SourceRegistrar(null));
+        }
+
+        [Fact]
+        public void RegisterSource_ChainsSourceRegistrations()
+        {
+            var builder = new ContainerBuilder();
+            var registrar = new SourceRegistrar(builder);
+
+            registrar.RegisterSource<SourceA>()
+                .RegisterSource(new ObjectRegistrationSource(O2));
+
+            var container = builder.Build();
+            var objects = container.Resolve<IEnumerable<object>>();
+            Assert.Contains(O1, objects);
+            Assert.Contains(O2, objects);
+        }
+
+        [Fact]
+        public void RegisterSource_RequiresRegistrationSource()
+        {
+            var registrar = new SourceRegistrar(new ContainerBuilder());
+            Assert.Throws<ArgumentNullException>(() => registrar.RegisterSource(null));
+        }
+
+        private class SourceA : ObjectRegistrationSource
+        {
+            public SourceA()
+                : base(O1)
+            {
+            }
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


src/Autofac/RegistrationExtensions.cs
```diff
@@ -62,19 +62,6 @@ public static void RegisterComponent(this ContainerBuilder builder, IComponentRe
             builder.RegisterCallback(cr => cr.Register(registration));
         }
 
-        /// <summary>
```
tillig: We can't move this out of here because that's a breaking API change.

test/Autofac.Test/Core/Registration/SourceRegistrarTests.cs
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        [Fact]
+        public void RegisterSource_ChainsSourceRegistrations()
+        {
+            var builder = new ContainerBuilder();
+            var registrar = new SourceRegistrar(builder);
+
+            registrar.RegisterSource<SourceA>()
+                .RegisterSource(new ObjectRegistrationSource(O2));
```
tillig: In our existing builder syntax, the chaining allows you to configure the thing first being registered. That is...

```csharp
builder.RegisterType<Foo>().As<IFoo>();
```

The `As` is configuring the type registration. You can't do:

```csharp
builder.RegisterType<Foo>().RegisterType<Bar>();
```

Adding chaining like this makes for an inconsistent API, so we probably don't want to do that.

src/Autofac/RegistrationExtensions.cs
tillig: If you're looking for `RegisterSource<T>` you could probably pretty easily just add

```csharp
public static void RegisterSource<T>(this ContainerBuilder builder) where T : IRegistrationSource, new
{
  var source = new T();
  builder.RegisterSource(source);
}
```

test/Autofac.Test/Core/Registration/SourceRegistrarTests.cs
eugbaranov: But you can currently do `builder.RegisterModule<ModuleA>().RegisterModule<ModuleB>()`.
Registration sources feel more like modules in a sense that they provide extensibility to the container so I expected them to have API similar to modules.

eugbaranov: I don't mind moving it back but just wondering - is it really that breaking? Sure, it is binary incompatible but anyone using it as an extensions method won't have to change anything, just recompile.

tillig: Yeah, this is one we've got before. The problem is transitive dependencies. My project needs Foo, Foo needs Autofac. I can't recompile Foo, so I'm stuck unable to upgrade.

eugbaranov: Okay, I moved the original extension method back but since made it inconsistent in a sense that you can only perform chained registrations if you started with generic method since original method not returning anything.

Would you consider breaking change (as per original PR) for some future major version release, e.g. 5.0.0?

tillig: Maybe. I'm guessing there are very few folks out there actually registering sources so it'd be breaking but fairly minimally breaking. I'd be OK with that in a major version release.

This binary compatibility stuff makes it really painful to make much progress or refactor well, as you'll see. We probably need to make a list of things we want to do for a 5.0 release and possibly branch to implement them when it's time. Of course, there's also the challenge where it takes a while to do some of these larger changes and long running branches with lots of changes are hard to maintain.

eugbaranov: Yeah, makes sense.

tillig:
Looks good - thanks for helping!

PR #942, opened 2018-10-30
Inherit Registered callback in child scopes (closes #218)

### LINKED ISSUE #218: Module.AttachToComponentRegistration() does not detect new registrations in nested scopes
_From [nicholas...@gmail.com](https://code.google.com/u/104985896277148547614/) on May 10, 2010 06:23:53_

See: http://groups.google.com/group/autofac/browse_thread/thread/86592faf0bb5aad9

_Original issue: http://code.google.com/p/autofac/issues/detail?id=218_

eugbaranov:
Makes module behavior more consistent in lifetime scopes and my log4net module PR obsolete.

### DIFF
--- a/test/Autofac.Test/ModuleTests.cs
+++ b/test/Autofac.Test/ModuleTests.cs
@@ -59,6 +59,42 @@ namespace Autofac.Test
             Assert.Equal(container.ComponentRegistry.Registrations.Count(), attachingModule.Registrations.Count);
         }
 
+        [Fact]
+        public void AttachesToRegistrationsInScope()
+        {
+            var attachingModule = new AttachingModule();
+            Assert.Equal(0, attachingModule.Registrations.Count);
+
+            var builder = new ContainerBuilder();
+            builder.RegisterModule(attachingModule);
+
+            using (var container = builder.Build())
+            using (var scope = container.BeginLifetimeScope(c => c.RegisterType(typeof(int))))
+            {
+                var expected = container.ComponentRegistry.Registrations.Count() + scope.ComponentRegistry.Registrations.Count();
+                Assert.Equal(expected, attachingModule.Registrations.Count);
+            }
+        }
+
+        [Fact]
+        public void AttachesToRegistrationsInNestedScope()
+        {
+            var attachingModule = new AttachingModule();
+            Assert.Equal(0, attachingModule.Registrations.Count);
+
+            var builder = new ContainerBuilder();
+            builder.RegisterModule(attachingModule);
+
+            using (var container = builder.Build())
+            using (var outerScope = container.BeginLifetimeScope(c => c.RegisterType(typeof(int))))
+            using (var innerScope = outerScope.BeginLifetimeScope(c => c.RegisterType(typeof(double))))
+            {
+                var expected = container.ComponentRegistry.Registrations.Count()
+                    + outerScope.ComponentRegistry.Registrations.Count() + innerScope.ComponentRegistry.Registrations.Count();
+                Assert.Equal(expected, attachingModule.Registrations.Count);
+            }
+        }
+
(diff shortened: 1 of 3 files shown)


tillig:
_Very_ cool! I thought this would be much harder.

The only thing I need to noodle for a bit is the lock. Any time we lock on anything we get folks with odd situations who complain that throughput just dropped. For example, if you have request lifetime scopes in a high-traffic ASP.NET app the lock may cause some challenges. I'm not sure if it'll matter right here or not since the new lifetime scope is going to get a new component registry. Registration is generally a single-threaded process anyway so... it shouldn't matter?

I'm curious if @alexmg has anything sitting in the back of his mind. Since it's all of like five lines, maybe he can chime in to see if I'm off in the weeds here.

eugbaranov:
The lock there is only for adding/removing handlers to the event which shouldn't happen very often, e.g. usually once per module registration, the actual calling of event is lock-free.
I could have added more fine grained lock just for that event but that felt like an overkill to be honest.

test/Autofac.Test/ModuleTests.cs
```diff
@@ -59,6 +59,42 @@ public void AttachesToRegistrations()
...
+
+            var builder = new ContainerBuilder();
+            builder.RegisterModule(attachingModule);
+
+            using (var container = builder.Build())
+            using (var outerScope = container.BeginLifetimeScope(c => c.RegisterType(typeof(int))))
+            using (var innerScope = outerScope.BeginLifetimeScope(c => c.RegisterType(typeof(double))))
+            {
+                var expected = container.ComponentRegistry.Registrations.Count()
+                    + outerScope.ComponentRegistry.Registrations.Count() + innerScope.ComponentRegistry.Registrations.Count();
```
tillig: Should we test that the module _removes_ its subscriptions as scopes are disposed?

eugbaranov: But why would it need to? Once the scope gets out of scope (pun intended) and becomes unreachable module's registration will became irrelevant and waiting GC together will scope's component registry. Or am I missing something?

tillig: I just want to avoid multigenerational garbage collection problems (ie some event handler inadvertently keeping references alive) and/or the `Properties[MetadataKeys.RegisteredPropertyKey]` set of registrations becoming so large it becomes problematic. The whole way this works is because the properties dictionary is inherited from parent to child, right? And as it reads-through to the parent scope, it's actually attaching the new set of event handlers to the object reference in the parent, not something scoped to only the current scope. Wouldn't that mean if I fire up 100 child lifetime scopes I'd be adding to the subscription list and never removing them? Would that cause problems in that the delegate chain in the event handler would become invalid?

I'm not entirely sure, it just occurred to me that we might have a leak of some nature here given there's no cleanup. I've had some experience with that - I tried to implement a thing that would track child lifetime scopes in an effort to dispose of them when a parent was disposed, did something sort of similar to this... and found that I accidentally ended up keeping references to every child scope so nothing got disposed. It was... not good. ;)

eugbaranov: Yes, but in this instance module is not attaching to component registry - only to possibly _new registrations_ and those will be brand new and nothing to do with reference in the parent.

Only if you try adding new handler to `Registered` inside new scope you'd be looking at the original delegate from the parent registry - but the moment you attempt to modify it new properties dictionary will be created with original values copied (well, sort of) and since delegates are immutable the parent and child will end up with their distinct versions of it. See the unit test I just added.

tillig:
I'm pretty stoked - this closes a pretty long running issue. Four years?! Super cool. Huge thanks.

eugbaranov:
You can tell than I really wanted that integration sample to work 😄

alexmg:
Nice work @eugbaranov!

PR #946, opened 2018-11-05
Propagate ObjectDisposedException

### LINKED ISSUE #945: Sometimes resolve operation throws DependencyResolutionException if the scope is disposed from another thread
I have registered instance `A` of interface `IA`. 
Sometimes when I resolve an interface `IA` and dispose a scope from concurrent thread I get `DependencyResolutionException` while `ObjectDisposedException` is expected.

**Sample test to reproduce:**
```sh
        private interface IA
        {
        }

        private class A : IA
        {
        }

        [Fact]
        public async Task RepeatedResolveExternallyOwnedInstanceWhileTheScopeIsDisposing_ObjectDisposedExceptionThrownOnly()
        {
            for (int i = 0; i < 200; i++)
            {
                await ResolveExternallyOwnedInstanceWhileTheScopeIsDisposing_ObjectDisposedExceptionThrownOnly();
            }
        }

        private async Task ResolveExternallyOwnedInstanceWhileTheScopeIsDisposing_ObjectDisposedExceptionThrownOnly()
        {
            var container = cb.Build();

            var scope = container.BeginLifetimeScope(
                "name", builder =>
                {
                    var a = new A();
                    builder.RegisterInstance<IA>(a);
                });

            var t = Task.Run(
                () =>
                {
                    try
                    {
                        while (true)
                        {
                            scope.Resolve<IA>();
                        }
                    }
                    catch (ObjectDisposedException e)
                    {
                    }
                });
            await Task.Delay(5);

            scope.Dispose();

            await t;
        }
```

eugbaranov:
Closes #945


src/Autofac/Core/Activators/InstanceActivatorResources.resx
```diff
@@ -0,0 +1,123 @@
...
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="InstanceActivatorDisposed" xml:space="preserve">
+    <value>Instances cannot be created from this ProvidedInstance as it has already been disposed.</value>
```
tillig: Since this gets used by several activators, can the message be more generic, like

`Instances cannot be created by this activator as it has already been disposed.`

eugbaranov: Well spotted - I started in `ProvidedInstance` activator but then figured they all would benefit from same check. Updated.

tillig:
Perfecto! Thanks!

Issue #635, opened 2015-04-13
SingleInstance lock 


nillis:
We were encountering problems on our highload multithreaded application with complex dependency graph caused by the LifetimeScope lock. 

Problem:
The lock is provided to work around the non-threadsafe Dictionary. Although this lock is cheap, it's acquired for each SingleInstance resolve operation and shared across threads. In our case, this caused problems (high-cpu) when GC occurs and the triggering thread was blocking all other threads.

Solution:
For 4.0: replace the Dictionary with a ConcurrentDictionary and remove the Monitor lock
For 3.5: replace the Monitor lock with a ReaderWriterLockSlim lock (https://gist.github.com/nillis/0c63930229ac9164ff09)

Are there other known cases where this problem arises? Are there any pitfalls on the proposed solution?

tillig:
We've had a couple of reports of this sort of thing in the past but we've had some challenges in addressing it:

- **Solutions have usually introduced problems in other areas.** For example, it may solve the single instance problem but cause issues with things that aren't single instance. Or maybe it solves the performance problem but adds a memory usage issue. That kind of thing.
- **We've not really seen any hard data to back up the issue.** For example, in my personal experience at work we also have a pretty high volume app and we were doing "memory dump" style "performance testing" trying to catch memory dumps at random intervals to see if we could find hot spots under load. However, when we actually attached a profiler to it, it turned out we were catching memory dumps at the wrong spot and the optimization actually needed to happen elsewhere.

**Not that I'm saying there's no opportunity to optimize, just that we want to be *very careful* here.**

Do you have any profiler data you could share - even a screen shot from ANTS or something like that? Before we dive into solving the problem, we just want to make sure that there's something backing up the issue. It'll also be good to ensure that once we issue a fix, we can see the hot spot go away to know we've fixed it.

Note that we're neck deep in ASP.NET v5 (vNext?) and the whole "DNX" thing right now, so we probably won't be able to do much with this until that's settled out. Lots of moving pieces. That also means it'd be solved in the 4.x codeline, not back-ported to 3.x.

nillis:
The lock is only used when resolving Shared instances (f.e. SingleInstance) to ensure the thread-safety of the dictionary which contains these instances. The problem with the lock is that it's always done, also after initialization of the shared instance.

[...]

Issue #666, opened 2015-07-24
Threading issue in LifeTimeScope.GetOrCreateAndShare


hackfaq:
I have WCF service, and on method calls I get exception:

```
An item with the same key has already been added.

System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   в System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   в System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
   в Autofac.Core.Lifetime.LifetimeScope.GetOrCreateAndShare(Guid id, Func`1 creator)
```

as I investigated sources such error can't happen, because it checks before add.

tillig:
Can you please provide a repro and post it somewhere? Given many people are successfully using Autofac with WCF and not seeing this error, we'll need to see it ourselves in order to troubleshoot.

hackfaq:
look at image, I already have TicketController in _sharedInstances with id, but it tries tp insert again.
![image](https://cloud.githubusercontent.com/assets/2344814/8888401/5eac270a-32b4-11e5-985c-c4f5c5815fbf.png)

hackfaq:
I have rewrited method:GetOrCreateAndShare, added check for key, and it it works now.

public object GetOrCreateAndShare(Guid id, Func<object> creator)
		{
			if (creator == null)
			{
				throw new ArgumentNullException("creator");
			}
			lock (_synchRoot)
			{
				object result;
				if (!_sharedInstances.TryGetValue(id, out result))
				{
					result = creator();
					lock (_synchRoot)
					{
						if (!_sharedInstances.ContainsKey(id))
							_sharedInstances.Add(id, result);
					}
				}
				return result;
			}
		}

alexmg:
A unit test showing the issue along with the fix would be best. Wondering what exactly the cause is and a screen capture isn't something we can execute to understand the problem.

hackfaq:
Dictionary is not thread safe, so it should be double checked, I think

alexmg:
I think I see the problem. I'll take a look. I suspect we should also prevent the creator function from being invoked a second time too, even if the second lock prevents the instance from being added to the dictionary.

tillig:
I'm guessing the ContainsKey call is what made the difference (though it escapes me how, since it should have been covered by the try-get). The second/nested lock won't have any effect on the use of the dictionary because no other threads can get into that critical section while the outer lock is held.

[...]

PR #952, opened 2018-12-07
Fix for #635 SingleInstance lock

avtc:
Use lock only if instance not created yet

ConcurrentDictionary GetOrAdd AddOrUpdate with value factory is not thread safe, so lock still needed
https://github.com/dotnet/corefx/issues/25609

### DIFF
--- a/src/Autofac/Core/Lifetime/LifetimeScope.cs
+++ b/src/Autofac/Core/Lifetime/LifetimeScope.cs
@@ -24,6 +24,7 @@
 // OTHER DEALINGS IN THE SOFTWARE.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
@@ -47,7 +48,7 @@ namespace Autofac.Core.Lifetime
         /// Protects shared instances from concurrent access. Other members and the base class are threadsafe.
         /// </summary>
         private readonly object _synchRoot = new object();
-        private readonly IDictionary<Guid, object> _sharedInstances = new Dictionary<Guid, object>();
+        private readonly ConcurrentDictionary<Guid, object> _sharedInstances = new ConcurrentDictionary<Guid, object>();
 
         private readonly ISharingLifetimeScope _parent;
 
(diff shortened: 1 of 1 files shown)


eugbaranov:
What's the benefit of having `ConcurrentDictionary` given that adding value is behind the lock?
Technically speaking `ConcurrentDictionary` is thread-safe but two competing threads might end up filling the same key discarding one of the results. Instead of double locking (explicit `_syncRoot` and implicit within `TryAdd`) usually it is better to have `ConcurrentDictionary<Guid, Lazy<object>>` where `Lazy` will guarantee execution protection.

avtc:
Performance issue is not in Adding instances, but in Resolving already Added instances.
Proposed solution provides lock-free Resolve (when instance exists).

Lazy with Execution protection also uses Monitor lock, so it will still be two locks (TryAdd and Lazy): 
https://github.com/Microsoft/referencesource/blob/60a4f8b853f60a424e36c7bf60f9b5b5f1973ed1/mscorlib/system/Lazy.cs#L382

eugbaranov:
Fair enough. Also on a second thought although `Lazy` provides finer lock granularity coarser lock it required since #666 (and exception you mentioned) suggests that `creator()` might result in creating multiple instances. @alexmg will probably know the best.

tillig:
Hi! I want to acknowledge that I recognize this PR is here, but with the holidays approaching and family/job obligations increasing, I likely won't personally have time to review and/or approve this in the super near future. @alexmg may have some additional available time, but I'm a little swamped. I apologize in advance for any delay.

alexmg:
I have used `Lazy<T>` with `ConcurrentDictionary.GetOrAdd` in the past successfully, but in this case the requirement to throw the `DependencyResolutionException` from #666 would makes its use rather akward. There is no simple way to know if an existing instance was returned or not, so in the case of the recursive resolve operation the `Value` property of the `Lazy<T>` instance would be accessed from inside the value factory function, causing an `InvalidOperationException` to be thrown, and that would need to be caught as a unreliable means of detecting the cycle.

In addition, because the `Func<object>` is passed into the `LifetimeScope.GetOrCreateAndShare` method it would need to be wrapped in a `Lazy<T>` or `Func<Lazy<T>>` in order to call `ConcurrentDictionary<Guid, Lazy<object>>.GetOrAdd`. This extra allocation would be needed every time the method was called, even in cases where the instance already exists in the dictionary. The fact that `Lazy<T>` is being used as an optimization inside the `LifetimeScope` class should not be visible externally so changing the signature of `GetOrCreateAndShare` isn't desirable either.

I think this is a good change overall and the trade off with manual locking to meet the requirement of #666 is reasonable. I'm going to pull this in for the 4.9.0 release.

Issue #941, opened 2018-10-29
contention using autofac in multi-threaded high throughput applications


oshvartz:
Hi
We have asp.net application with high load ~5000 rps. we use orchard cms that uses autofac under the hood. recently our web application is utilizing the the machine to ~100% CPU and after reviewing the dump we discover that most of the threads had the same call stack (~95%):
System.Threading.Monitor.Enter(System.Object)
Autofac.Core.Registration.ComponentRegistry.RegistrationsFor(Autofac.Core.Service)
System.Threading.Monitor.Enter(System.Object)
Autofac.Core.Registration.ComponentRegistry.TryGetRegistration(Autofac.Core.Service, Autofac.Core.IComponentRegistration ByRef)

This all are waiting for the monitor.Enter - (i found the code has simple lock)
does it means autofac is not build to be used in high throughput application ? isn't it a common use case ?
Thanks
Offir

tillig:
Duplicate #906. See that issue for more details.

Short version: We need to lock when resolving things to ensure thread safety on resolve. If you can find additional ways to improve locking and throughput without removing the existing level of thread safety, we are absolutely open to pull requests.

Autofac is being successfully run in many high-throughput apps, but sometimes app code may need to be tweaked to handle things. I'd recommend gathering some hot-path data from your profiler to see what's causing all the resolutions to happen and see if you can optimize. If you are resolving tons and tons and tons of things, your overall app performance may improve by reducing resolutions - which will also reduce memory allocations, garbage collections, and a lot of other things, making your overall app perform better.

oshvartz:
Hi @tillig 
I know that is asp.net application common DI life cycle is "Scoped" - lifetime services are created once per request.
so if i have web API with DI using autofac - my request per second is throttled because only one resolve can happen in a given time using the same container
what am i missing here ?

tillig:
If a component isn't getting resolved from the root container then the root container isn't locked. What you're missing is that if you have a scope - either the root container or a child scope - that resolving from _that scope_ should be thread safe. Which is to say, you can only do one resolve from _that scope_ at a time, and generally the locking only happens for shared components (e.g., InstancePerLifetimeScope). You are not as limited as you think.

[...]

Issue #906, opened 2018-04-11
Lock contention in ComponentRegistry.RegistrationsFor


bluetentacle:
Dear all,

In production, we are encountering high lock contention in the ComponentRegistry.RegistrationsFor method. It appears that every single object resolution has to go through this lock, and it's therefore a performance bottleneck in highly multithreaded scenarios.

Is there a way of working around this bottleneck, and are there any plans to make Autofac lockless?

Thanks.

bluetentacle:
Updating title

tillig:
I don't think it does because we still have to calculate the services exposed by registration sources (totally dynamic).

bluetentacle:
Looking at the source, I see what you mean. Thanks!

tillig:
Since there's no hot path data here and there's not much we can do without that to optimize, I'm going to close this issue for now. That said, we're open to...

- Pull requests that improve the locking that's there or adjust internals to remove the need for it (assuming it maintains the current or better levels of thread safety for resolve operations); OR
- An issue with a more specific proposed solution that can be discussed prior to a PR of that nature.

We're always open to help and improvements on perf, it just won't help to keep this specific issue open and "on the books" as we iterate over the notion of perf in general.

eugbaranov:
I wonder if switching `_serviceInfo` to `ConcurrentDictionary` could improve this?

PR #953, opened 2018-12-21
Fix lock contention #941 #906

avtc:
Nolock solution for `TryGetRegistration`, `IsRegistered`, `RegistrationsFor`. Issues: #941, #906

For case when `ServiceRegistrationInfo` is already exists and `IsInitialized`, i.e. after service was Resolved once.

There can be little memory overhead during initialization, but that is a small price for allowing it to work in multi-core high load environment.

### DIFF
--- a/src/Autofac/Core/Registration/ServiceRegistrationInfo.cs
+++ b/src/Autofac/Core/Registration/ServiceRegistrationInfo.cs
@@ -197,15 +197,35 @@ namespace Autofac.Core.Registration
         {
             RequiresInitialization();
 
-            registration = Interlocked.CompareExchange(
-                ref _defaultImplementation,
-                Volatile.Read(ref _defaultImplementations).LastOrDefault() ?? Volatile.Read(ref _sourceImplementations)?.First() ?? Volatile.Read(ref _preserveDefaultImplementations)?.First(),
-                null);
+            registration = Volatile.Read(ref _defaultImplementation);
+            if (registration != null)
+                return true;
 
-            if (registration == null)
-                registration = Volatile.Read(ref _defaultImplementation);
+            var defaultImplementations = Volatile.Read(ref _defaultImplementations);
+            if (defaultImplementations.Count > 0)
+            {
+                registration = defaultImplementations[defaultImplementations.Count - 1];
+                Volatile.Write(ref _defaultImplementation, registration);
+                return true;
+            }
 
-            return registration != null;
+            var sourceImplementations = Volatile.Read(ref _sourceImplementations);
+            if (sourceImplementations != null && sourceImplementations.Count > 0)
+            {
+                registration = sourceImplementations[0];
+                Volatile.Write(ref _defaultImplementation, registration);
+                return true;
+            }
+
+            var preserveDefaultImplementations = Volatile.Read(ref _preserveDefaultImplementations);
+            if (preserveDefaultImplementations != null && preserveDefaultImplementations.Count > 0)
+            {
+                registration = preserveDefaultImplementations[0];
+                Volatile.Write(ref _defaultImplementation, registration);
+                return true;
+            }
+
+            return false;
         }
 
         public void Include(IRegistrationSource source)
--- a/src/Autofac/Core/Registration/ServiceRegistrationInfo.cs
+++ b/src/Autofac/Core/Registration/ServiceRegistrationInfo.cs
@@ -27,6 +27,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
+using System.Threading;
 
 namespace Autofac.Core.Registration
 {
@@ -41,21 +42,27 @@ namespace Autofac.Core.Registration
         /// <summary>
         ///  List of implicit default service implementations. Overriding default implementations are appended to the end,
         ///  so the enumeration should begin from the end too, and the most default implementation comes last.
+        ///  Used from multithreaded environment, use Interlocked/Volatile for read and write new List each time.
         /// </summary>
-        private readonly List<IComponentRegistration> _defaultImplementations = new List<IComponentRegistration>();
+        private List<IComponentRegistration> _defaultImplementations = new List<IComponentRegistration>();
 
         /// <summary>
         ///  List of service implementations coming from sources. Sources have priority over preserve-default implementations.
         ///  Implementations from sources are enumerated in preserve-default order, so the most default implementation comes first.
+        ///  Used from multithreaded environment, use Interlocked/Volatile for read and write new List each time.
         /// </summary>
         private List<IComponentRegistration> _sourceImplementations = null;
 
         /// <summary>
         ///  List of explicit service implementations specified with the PreserveExistingDefaults option.
         ///  Enumerated in preserve-defaults order, so the most default implementation comes first.
+        ///  Used from multithreaded environment, use Interlocked/Volatile for read and write new List each time.
         /// </summary>
         private List<IComponentRegistration> _preserveDefaultImplementations = null;
 
+        /// <summary>
+        ///  Used from multithreaded environment, use Interlocked/Volatile for read and write each time.
+        /// </summary>
         [SuppressMessage("Microsoft.ApiDesignGuidelines", "CA2213", Justification = "The creator of the compponent registration is responsible for disposal.")]
         private IComponentRegistration _defaultImplementation;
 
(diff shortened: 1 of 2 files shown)


eugbaranov:
It's possible to create some sort of benchmark to test both old and new implementations?

Do you have any screenshots from concurrency visualizer comparing the two?

avtc:
> Do you have any screenshots from concurrency visualizer comparing the two?

No.
I have is 85% CPU of 48 cores used by program without the fix, with profiler shows 99.6% of lock contention is inside `TryGetRegistration` and host become unresponsive. After fix the CPU is 5-10% and program works fine in the long run.

> some sort of benchmark

Used https://github.com/danielpalme/IocPerformance
1) Without the fix applied
```
Autofac 4.9.0-beta1                          Single      Multi
 Singleton                                      698        867
 Transient                                      988        940
 Combined                                      2811       2637
 Complex                                       8619       8366
 Property                                      8562       8175
 Generics                                      3381       2400
 IEnumerable                                  11996       6747
 Conditional                                   2026       1890
 Benchmark 'Child Container' (single thread) was stopped after 1,0 minutes. 266000 of 500000 instances have been resolved. Total execution would have taken: 1,9 minutes.
 Benchmark 'Child Container' (multiple threads) was stopped after 1,0 minutes. 476303 of 500000 instances have been resolved. Total execution would have taken: 1,0 minutes.
 Child Container                            112842*     62986*
 Asp Net Core                                 20338      13647
 Interception With Proxy                      28231      14754
 Prepare And Register                           338           
 Prepare And Register And Simple Resolve        395           
```


2) With applied this pull and SingleInstance nolock pull https://github.com/autofac/Autofac/pull/952
```
Autofac 4.9.0-beta-2-nolock                  Single      Multi
 Singleton                                      671        426
 Transient                                      817        570
 Combined                                      2438       1531
 Complex                                       7408       4164
 Property                                      7420       4141
 Generics                                      3027       1795
 IEnumerable                                  11513       6406
 Conditional                                   1969       1238
 Benchmark 'Child Container' (single thread) was stopped after 1,0 minutes. 248000 of 500000 instances have been resolved. Total execution would have taken: 2,0 minutes.
 Benchmark 'Child Container' (multiple threads) was stopped after 1,0 minutes. 455460 of 500000 instances have been resolved. Total execution would have taken: 1,1 minutes.
 Child Container                            121120*     65868*
 Asp Net Core                                 19796      11432
 Interception With Proxy                      28335      14758
 Prepare And Register                           460           
 Prepare And Register And Simple Resolve        529          
```

Host:
```
OS=Windows 10.0.17134.472 (1803/April2018Update/Redstone4)
AMD FX(tm)-8320 Eight-Core Processor (Max: 3.50GHz), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=2.1.403
```

alexmg:
Thanks for putting together another good PR @avtc.

I wrote a quick multi threading test that could be used for some profiling and to visualize the concurrency improvements. It does 1,000,000 resolve operations  as 10,000 resolves across 100 concurrent tasks.

In the screen capture below is the current code without the PR. I have highlighted **Lock contention** and there is plenty to be seen (the test was designed to really highlight the bottleneck). The total duration is around 347ms but this is with the profiler attached and should only be used as an indicator for comparison to the following run that also has the profiler attached.

concurrency locks

As expected the reduced locking definitely helps when concurrency comes into play. The screen capture below is with the changes in the PR. There is some lock contention up front as seen above, but after that locking is eliminated and the total time is around 160ms.

concurrency pr

I ran the same benchmarks and below is the 4.9.0 baseline on my machine.

```
Autofac 4.9.0                                Single      Multi
 Singleton                                      352        362
 Transient                                      451        348
 Combined                                      1272       1090
 Complex                                       3641       3511
 Property                                      3705       3415
 Generics                                      1259        997
 IEnumerable                                   4321       2737
 Conditional                                    868        848
 Child Container                              48406      26499
 Asp Net Core                                  7943       6236
 Interception With Proxy                      13240       6795
 Prepare And Register                           160
 Prepare And Register And Simple Resolve        191
```

Here is the run with the changes and most things look better (the multi tests in particular).

```
Autofac 4.9.0                                Single      Multi
 Singleton                                      329        198
 Transient                                      431        260
 Combined                                      1129        626
 Complex                                       3637       1873
 Property                                      3463       1845
 Generics                                      1257        699
 IEnumerable                                   4363       2349
 Conditional                                    840        470
 Child Container                              52734      27608
 Asp Net Core                                  7877       4323
 Interception With Proxy                      13827       7088
 Prepare And Register                           224
 Prepare And Register And Simple Resolve        254
```

Without the `ServiceRegistrationInfo` change the number are slightly better so if not needed I would like to leave them out. Did you come across a particular scenario that required the `ServiceRegistrationInfo` changes? I ran the same test without them and didn't hit a problem but also didn't create any additional and more advanced tests either.

```
Autofac 4.9.0                                Single      Multi
 Singleton                                      332        196
 Transient                                      410        249
 Combined                                      1110        612
 Complex                                       3433       1812
 Property                                      3546       1817
 Generics                                      1206        665
 IEnumerable                                   4304       2293
 Conditional                                    810        468
 Child Container                              51897      27174
 Asp Net Core                                  7881       4276
 Interception With Proxy                      13124       6793
 Prepare And Register                           226
 Prepare And Register And Simple Resolve        255
```

I was going to leave this PR for after the `4.9.0` release but am now considering including them.

src/Autofac/Core/Registration/ServiceRegistrationInfo.cs
```diff
@@ -122,54 +144,88 @@ public bool IsRegistered
...
-                    if (_preserveDefaultImplementations == null)
+                    // called inside of lock, just making List _preserveDefaultImplementations thread safe
+                    if (Volatile.Read(ref _preserveDefaultImplementations) == null)
                     {
-                        _preserveDefaultImplementations = new List<IComponentRegistration>();
+                        Volatile.Write(ref _preserveDefaultImplementations, new List<IComponentRegistration>());
                     }
 
-                    _preserveDefaultImplementations.Add(registration);
+                    var newPreserveDefaultImplementations = new List<IComponentRegistration>(Volatile.Read(ref _preserveDefaultImplementations));
```
tillig: PR #919 attempts to optimize the amount of memory traffic going on in an effort to speed up things like mobile applications, where allocations cost time and battery. Having to allocate a whole new list every time here seems like it might counteract some of those optimizations. We might have to balance the lock contention improvement with the memory usage/optimization improvements.

tillig:
I got a chance to look at this and, just eyeballing it, I think the `ComponentRegistry` changes are good. I made a note on `ServiceRegistrationInfo` where the things done to optimize lock contention may help there but may also work against memory traffic optimizations from PR #919. If we can take just the `ComponentRegistry` changes as an incremental improvement, I think that'd be good.

alexmg:
I have taken the `ComponentRegistry` changes and included them in the `4.9.0` release. Thank you for the PR @avtc.

avtc:
@alexmg you can not just take changes in `ComponentRegistry` without taking changes from `ServiceRegistrationInfo`. Because changes in `ComponentRegistry` uses `ServiceRegistrationInfo` non thread safe properties `IsRegistered`, `Implementations`, `IsInitialized` and method `TryGetRegistration`.

Please consider reverting your partial changes or take full changes.

alexmg:
Hi @avtc. Your response does come a little late as the release has been pushed to NuGet, but not too late for another release if required. I guess you must have missed the comment from @tillig and I took the lack of response to mean you had no concerns.

`IsRegistered`, `Implementations`, and `TryGetRegistration` all require that `IsInitialized` be `true` before they are accessed as checked by the `RequiresInitialization` method in `ServiceRegistrationInfo`. That property is checked in `GetInitializedServiceInfoOrDefault` so if a `ServiceRegistrationInfo` is picked up before it has been initialized the code in the `ComponentRegistry` should drop into the lock statement that then uses `GetInitializedServiceInfo`.

In the concurrency tests I created I did not see any `InvalidOperationException` being thrown from the `RequiresInitialization` check indicating the `IsInitialized` check in `GetInitializedServiceInfoOrDefault` seemed to be doing its job. I might be missing something else that you noticed during your testing though, and if truly required can add the changes and push out a `4.9.1` release.

avtc:
@alexmg you are right, i missed that point. _Maybe only in case registrations will be added after `ServiceRegistrationInfo` is initialized it will not work as expected._

Also, i would suggest to make `IsInitialized` `volatile`.

alexmg:
@avtc I actually went and had another look at the `ServiceRegistrationInfo` changes last night and found the `volatile` field for `IsInitialized` and brought it across for the next release. Because the value only ever transitions from `false` to `true`, it seems the worst case for reordering is that a thread that could have avoided the lock, would end up taking one because `false` was returned when it had actually already transitioned to `true`. That should be alright to wait for the next bug fix release.

alexmg:
I have released `4.9.1` which includes the `volatile` change for `IsInitialized`.

PR #961, opened 2019-02-14
Do not reuse ExternalComponentRegistration

### LINKED ISSUE #960: ExternalRegistrySource reuses ExternalComponentRegistration instances, leading to use-after-dispose
**Edit: updated issue description here. Original below and in comments.**

Consider a three-level lifetime scope hierarchy:

1. root
1. middle
1. child

Register type T in `root` with InstancePerLifetimeScope.

Register some irrelevant types in `middle` and in `child` (to materialize a scope-specific registry and not reuse the copy-on-write registry).

Now resolve type T from `middle`. This results in a new `ExternalComponentRegistration` being added to the registry of `middle`, based on the registration in `root`.

Now resolve type T from `child`. This results in **the same instance** of `ExternalComponentRegistration` being added to the registry of `child` as was added to `middle`.

Now dispose of `child`. The registration **still shared with `middle`** is disposed of!

Now attempt to resolve T from `middle`. You get ObjectDisposedException from the activator.

The root cause is that `ExternalRegistrySource.RegistrationsFor` will reuse existing instances from the parent if they are already ExternalComponentRegistrations.

Proposed fix: do not reuse instances of  `ExternalComponentRegistration`.

----------

I am trying to debug a registration disposal issue and request hints at what to look at, so I could properly analyze the situation and formulate a sensible bug report (or discover my error).

The situation is:

```
Exception thrown: 'System.ObjectDisposedException' in Autofac.dll
Instances cannot be created by this activator as it has already been disposed.
```

I have a root scope and a child scope. Neither is disposed. I am resolving from the child scope.

The failure appears to be when resolving my type T that has a parameter of type `ILifetimeScope` (used by my object T to create child scopes). The registration for `ILifetimeScope` has `IsDisposed=true` (looking in childScope.ComponentRegistry.Registrations).

In fact, registrations for all types inherited via `InstancePerLifetimeScope()` from my root scope are also `IsDisposed=true`. One registration that is defined directly on the child scope is `IsDisposed=false`.

Example of registering a type on the root scope (registration IsDisposed=true when exception occurs)

```
builder.RegisterType<DefaultSynchronousRetryMechanism>()
                .As<ISynchronousRetryMechanism>()
                .InstancePerLifetimeScope();
```

Example of creating child scope and registering a type on it (registration IsDisposed=false):

```
childScope = rootScope.BeginLifetimeScope(builder =>
{
    builder.RegisterType<JobExecutionAttempt>()
        .InstancePerLifetimeScope();
});
```

I later create a next layer child scope and try to resolve:

```
using (var jobScope = childScope.BeginLifetimeScope())
{
    var attempt = jobScope.Resolve<JobExecutionAttempt>();
    await attempt.TryExecuteAsync(nextJobId.Value);
}
```

`JobExecutionAttempt` takes parameter of type `ILifetimeScope` which triggers ObjectDisposedException in .Resolve().

I do not expect you to have a solution for me based on this meager info but I request input on what I should be looking at. It is very much unexpected to me that ILifetimeScope and inherited registrations are IsDisposed=true. What can cause registrations to be IsDisposed=true without the lifetime scope being disposed?

This seems to happen after some time - not on the first entry into the last using block but perhaps the second?

I will continue to dig and improve the issue report once I find more relevant data.

sandersaares:
Fixes #960

### DIFF
--- a/src/Autofac/Core/Registration/ExternalRegistrySource.cs
+++ b/src/Autofac/Core/Registration/ExternalRegistrySource.cs
@@ -68,13 +68,9 @@ namespace Autofac.Core.Registration
 
             // Issue #272: Taking from the registry the following registrations:
             //   - non-adapting own registrations: wrap them with ExternalComponentRegistration
-            //   - if the registration is from the parent registry of this registry,
-            //     it is already wrapped with ExternalComponentRegistration,
-            //     share it as is
             return _registry.RegistrationsFor(service)
                 .Where(r => r is ExternalComponentRegistration || !r.IsAdapting())
                 .Select(r =>
-                    r as ExternalComponentRegistration ??
 
                     // equivalent to following registation builder
                     //    RegistrationBuilder.ForDelegate(r.Activator.LimitType, (c, p) => c.ResolveComponent(r, p))
--- a/test/Autofac.Test/Core/Registration/ExternalRegistrySourceTests.cs
+++ b/test/Autofac.Test/Core/Registration/ExternalRegistrySourceTests.cs
@@ -36,5 +36,31 @@ namespace Autofac.Test.Core.Registration
             var allImplementationsOfServiceA = lifetime.Resolve<IEnumerable<IServiceA>>();
             Assert.Equal(2, allImplementationsOfServiceA.Count());
         }
+
+        // Issue #960
+        [Fact]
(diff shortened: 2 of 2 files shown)


alexmg:
Thank you! 👍

Issue #343, opened 2014-01-22
Include inner exception message in outer exception message when wrapping exceptions from constructors


alexmg:
_From [rikard.p...@gmail.com](https://code.google.com/u/106336222488889808820/) on September 06, 2011 17:39:15_

I have a problem propagating exceptions from my library.
For example:

public class ThrowException
{
    public ThrowException()
    {
        throw new FileNotFoundException("Can't find file");
    }
}

[TestMethod]
public void CorrectExceptionWhenResolving()
{
    var builder = new ContainerBuilder();
    builder.RegisterType<ThrowException>();
    var container = builder.Build();

```
try
{
    container.Resolve<ThrowException>();
    Assert.Fail("Exception expected");
}
catch(FileNotFoundException)
{
}
catch(DependencyResolutionException dre)
{
    Assert.IsTrue(dre.Message.Contains("Can't find file"));
}
catch
{
    Assert.Fail("Unexpected exception");
}
```

}

Ideally I would like to catch FileNotFoundException, but I understand why that is not possible. 
What I think would be great is that DependencyResolutionException message contains inner exception message. That way I don't need to explain to users that they should look at the inner exception instead of main exception.

For example this message "An exception was thrown while invoking the constructor 'Void .ctor()' on type 'ThrowException'. Exception: Can't find file" would be much better.

Regards,
Rikard

_Original issue: http://code.google.com/p/autofac/issues/detail?id=343_

Issue #648, opened 2015-06-19
Circular component dependency detected with manual IEnumerable<T> registration


felixwatts:
Here is a console application that causes a DependencyResolutionException. There is no actual problem with the dependency tree. The error message wrongly states that IEnumerable&lt;IPlugin&gt; depends on PluginsViewModel.

```
A first chance exception of type 'Autofac.Core.DependencyResolutionException' occurred in Autofac.dll

Additional information: Circular component dependency detected: AutofacBug.RootViewModel -> System.Collections.Generic.IEnumerable`1[[AutofacBug.IPlugin, AutofacBug, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]] -> AutofacBug.PluginsViewModel -> System.Collections.Generic.IEnumerable`1[[AutofacBug.IPlugin, AutofacBug, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]].
```

``` csharp
using System.Collections.Generic;
using System.Linq;
using Autofac;
using Autofac.Core;

namespace AutofacBug
{
    class Program
    {
        static void Main(string[] args)
        {
            var containerBuilder = new ContainerBuilder();

            containerBuilder.RegisterType<RootViewModel>().AsSelf().SingleInstance();
            containerBuilder.RegisterType<PluginsViewModel>().AsSelf().SingleInstance();

            RegisterPlugins(containerBuilder);

            var container = containerBuilder.Build();

            // the exception on this line wrongly states that IEnumerable<IPlugin> depends on PluginsViewModel. It does not and can be resolved.
            var rootViewModel = container.Resolve<RootViewModel>();
        }

        private static void RegisterPlugins(ContainerBuilder containerBuilder)
        {
            containerBuilder.RegisterType(typeof(Plugin1)).Named<IPlugin>("Plugin1");
            containerBuilder.RegisterType(typeof(Plugin2)).Named<IPlugin>("Plugin2");
            containerBuilder.Register(SafeResolveAllPlugins).As<IEnumerable<IPlugin>>().SingleInstance();
        }

        private static IEnumerable<IPlugin> SafeResolveAllPlugins(IComponentContext core)
        {
            var sucessfullyResolvedPlugins =
                new[] {"Plugin1", "Plugin2"}
                .Select(name => SafeResolvePlugin(name, core))
                .Where(p => p != null)
                .ToArray();

            return sucessfullyResolvedPlugins;
        }

[...]

PR #962, opened 2019-03-03
Reduce the verbosity of activator exception messages

nblumhardt:
Hello! Briefly discussed in an offline chat - this PR proposes to cut down the (now ridiculously) verbose exception messages that result from failures deep in the construction of a dependency graph.

For example, before the change `C -> B -> A` results in this monster:

```
An error occurred during the activation of a particular registration. See the inner exception
for details. Registration: Activator = C (ReflectionActivator), Services =
[Autofac.Test.ExceptionReportingTests+C], Lifetime = Autofac.Core.Lifetime.CurrentScopeLifetime,
Sharing = None, Ownership = OwnedByLifetimeScope ---> An error occurred during the activation of
a particular registration. See the inner exception for details. Registration: Activator = B
(DelegateActivator), Services = [Autofac.Test.ExceptionReportingTests+B], Lifetime =
Autofac.Core.Lifetime.CurrentScopeLifetime, Sharing = None, Ownership = OwnedByLifetimeScope --->
An error occurred during the activation of a particular registration. See the inner exception for
details. Registration: Activator = A (ReflectionActivator), Services =
[Autofac.Test.ExceptionReportingTests+A], Lifetime = Autofac.Core.Lifetime.CurrentScopeLifetime,
Sharing = None, Ownership = OwnedByLifetimeScope ---> An exception was thrown while invoking the
constructor 'Void .ctor()' on type 'A'. ---> This is the original exception. (See inner exception
for details.) (See inner exception for details.) (See inner exception for details.) (See inner
exception for details.)
```

(Note _four_ repeats of the phrase "See inner exception for details" :-) ... the inner exception here turns out to be a similarly verbose Autofac exception/message, nested several deep, before we eventually hit something more useful.)

After:

```
An exception was thrown while activating DependencyResolutionExceptionTests+C -> 
λ:DependencyResolutionExceptionTests+B -> DependencyResolutionExceptionTests+A.
```

The new message no longer attempts to embed the inner exception message (reverts #343), and does not direct the user to it, because inner exceptions are common .NET machinery and the broad convention is for code that displays exceptions to include them, rather than for them to be duplicated by the wrapping exception. See, for example, `Exception.ToString()`, which performs this walk of the inner exception tree:

```
Autofac.Core.DependencyResolutionException: An exception was thrown while activating 
Autofac.Test.Core.DependencyResolutionExceptionTests+C -> 
λ:Autofac.Test.Core.DependencyResolutionExceptionTests+B -> 
Autofac.Test.Core.DependencyResolutionExceptionTests+A. ---> 
Autofac.Core.DependencyResolutionException: An exception was thrown while invoking the constructor 
'Void .ctor()' on type 'A'. ---> System.InvalidOperationException: This is the original exception.
   at Autofac.Test.Core.DependencyResolutionExceptionTests.A..ctor()
   at lambda_method(Closure , Object[] )
   at Autofac.Core.Activators.Reflection.ConstructorParameterBinding.Instantiate()
...
```

There's also a new shorthand for delegate activators `λ:{type}`, which will reduce the potential for confusion when the delegate's return type is an interface or other non-specific type.

The full component information is no longer included, but, this can be obtained by inspecting the `ComponentRegistry` in rare cases where multiple implementations might share the same limit type.

Thoughts?


test/Autofac.Test/Core/Resolving/CircularDependencyDetectorTests.cs
```diff
@@ -19,11 +15,11 @@ public void OnCircularDependency_MessageDescribesCycle()
...
             var target = builder.Build();
             var de = Assert.Throws<DependencyResolutionException>(() => target.Resolve<object>());
-            Assert.Contains("System.Object -> System.Object", de.Message);
-            Assert.DoesNotContain("System.Object -> System.Object -> System.Object", de.Message);
+            Assert.Contains("λ:System.Object -> λ:System.Object", de.ToString());
+            Assert.DoesNotContain("λ:System.Object -> λ:System.Object -> λ:System.Object", de.Message);
         }
 
-        [Fact(Skip = "Issue #648")]
+        [Fact]
```
nblumhardt: It appears #648 is resolved.

tillig:
I'll have to check this out locally and try it. I'd like to see what it looks like for things like where the cause is something like an unfulfilled dependency - the innermost exception actually being a DependencyResolutionException. As long as it's still clear, I'm pretty fine with making the exception simpler to read.

I'd be curious if there's a way to both provide a simpler message and keep some of the more detailed data. Then again, did folks use that extra data? Maybe not.

src/Autofac/Core/Resolving/ActivatorExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+    internal static class ActivatorExtensions
+    {
+        // This shorthand name for the activator is used in exception messages; for activator types
+        // where the limit type generally describes the activator exactly, we use that; for delegate
+        // activators, a variation on the type name is used to indicate this.
+        public static string DisplayName(this IInstanceActivator activator)
+        {
+            var fullName = activator.LimitType.FullName ?? "";
+            return activator is DelegateActivator ?
+                $"λ:{fullName}" :
```
tillig: Would it be interesting to have info about whether it was reflection or provided instance activation?

nblumhardt: Possibly; I didn't add this because there are few (no?) scenarios where a provided instance will trigger an activation exception.

I also was aiming to describe the type of the component, rather than its internal representation in Autofac - delegates are the special case because their `LimitType` might be surprising, e.g. `System.Object` when the actual component is one of several other alternatives.

nblumhardt:
Thanks for taking a look 👍 

Because the unwrapping only takes place when the `DependencyResolutionException` is tagged with an `ActivatorChain`, cases that involve other kinds of exceptions will utilise the `InnerException`. These should be much friendlier to view via the usual means, such as `Console.WriteLine()`, since this will perform the chaining together of outer and inner exceptions. If you're just examining the exceptions in a debugger, you need to expand `InnerException` to see all of the details, now.

alexmg:
I was actually hit by the verbose exception message today looking for the cause of an exception. The first sign of the actual error that I needed to find was at column 891. Without word wrap this was way off screen on my 4K monitor. With word wrap on it was a wall of hard to digest text. Let's take this in and iterate on it if we find any shortcomings. The delegate indicator is a nice touch too.

alexmg:
I never looked into why #648 is passing now but have decided to count it as a win and move on. 😄

nblumhardt:
Awesome 😄  I feel almost like a useful contributor again! Thanks for giving this one a shot, I'll pull it into our own dev branch and see if I can spot any problems in practical use over the next few weeks 👍

nblumhardt:
Published as https://www.myget.org/feed/autofac/package/nuget/Autofac/4.9.1-develop-00528 - just a thought, should this be a `4.10.0` or similar?

alexmg:
I've bumped SemVer to `4.9.2` for a different bug fix. I'm not sure this should push it to `4.10.0`. Feels like a bit of grey area. I'm not sure if we are adding, removing or replacing functionality in a non-breaking manner. 😃

tillig:
I'd stick with the 0.0.1 - the behavior is different but not in any area that should be relied upon at a programmatic level.

That said, devil's advocate, if someone had a logging system with triggers on the exception message via regex or something... We'd be breaking that, so a 1.0.0 full version change could also be justified.

But me, personally... 0.0.1 increment and done.

PR #977, opened 2019-04-04
Fix #976 - Throw an exception if a lifetime scope is created with a duplicate tag

### LINKED ISSUE #976: BeginLifetimeScope(object, Action<ContainerBuilder>) doesn't fail when using a duplicated tag
There is inconsistent behaviour when beginning a new lifetime scope between `LifetimeScope.BeginLifetimeScope(object)` and `LifetimeScope.BeginLifetimeScope(object, Action<ContainerBuilder>)`.

As indicated by the test `NestedScopeTests.BeginLifetimeScopeCannotBeCalledWithDuplicateTag` this will fail:

    var rootScope = new ContainerBuilder().Build();
    const string duplicateTagName = "ABC";
    var taggedScope = rootScope.BeginLifetimeScope(duplicateTagName);
    var differentTaggedScope = taggedScope.BeginLifetimeScope("DEF");

    Assert.Throws<InvalidOperationException>(() => differentTaggedScope.BeginLifetimeScope(duplicateTagName));

If I use `LifetimeScope.BeginLifetimeScope(object, Action<ContainerBuilder>)` then this will not throw an exception:

    var rootScope = new ContainerBuilder().Build();
    const string duplicateTagName = "ABC";
    var taggedScope = rootScope.BeginLifetimeScope(duplicateTagName);
    var differentTaggedScope = taggedScope.BeginLifetimeScope("DEF");

    Assert.Throws<InvalidOperationException>(() => differentTaggedScope.BeginLifetimeScope(duplicateTagName, builder => builder.RegisterType<object>()));

weelink:


tillig:
This is excellent. Thanks for the help!

Issue #972, opened 2019-03-27
Resolving services registered as open generics with generic constraints throws ComponentNotRegisteredException


ghen:
Possibly related to #688.

Whenever registered open generic has another generic class or interface reference in its type constraint, service resolution fails with ComponentNotRegisteredException error.

Code to reproduce:

```csharp
    private static void Main(String[] args) {

      var builder = new ContainerBuilder();
      builder.RegisterGeneric(typeof(GenericTypeClass<,>)).AsSelf();
      builder.RegisterGeneric(typeof(GenericCollectionClass<,>)).AsSelf();
      var container = builder.Build();

      // This would resolve nicely
      var resolved = container.Resolve<GenericTypeClass<StringContrainClassImp, String>>();

      // This would throw ComponentNotRegisteredException
      var notResolved = container.Resolve<GenericCollectionClass<CollectionContrainClassImp, String>>();

    }
  }

  public abstract class ConstraintClass<TInput> { }

  public sealed class GenericTypeClass<TConstraint, TInput> where TConstraint : ConstraintClass<TInput> { }
  public sealed class StringContrainClassImp : ConstraintClass<String> { }

  public sealed class GenericCollectionClass<TConstraint, TInput> where TConstraint : ConstraintClass<IEnumerable<TInput>> { }
  public sealed class CollectionContrainClassImp : ConstraintClass<IEnumerable<String>> {}
```

Exception details:

```
Unhandled Exception: Autofac.Core.Registration.ComponentNotRegisteredException: The requested service 'TestCore.GenericCollectionClass`2[[TestCore.CollectionContrainClassImp, TestCore, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null],[System.String, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]' has not been registered. To avoid this exception, either register a component to provide the service, check for service registration using IsRegistered(), or use the ResolveOptional() method to resolve an optional dependency.
   at Autofac.ResolutionExtensions.ResolveService(IComponentContext context, Service service, IEnumerable`1 parameters) in C:\projects\autofac\src\Autofac\ResolutionExtensions.cs:line 880
   at Autofac.ResolutionExtensions.Resolve[TService](IComponentContext context, IEnumerable`1 parameters) in C:\projects\autofac\src\Autofac\ResolutionExtensions.cs:line 300
[...]

PR #978, opened 2019-04-09
Match non-interface constraints using IsClosedTypeOf

Tawnos:
This addresses #972 by looking for non-interface generic type constraints and verifying all of the parameter's base type's generic type arguments are closed types of the constraint. Note: there may be additional cases of nesting that need to be tested and verified. I fixed the one that had a test case.


tillig:
Very slick. We appreciate the help! Thank you so much!

PR #981, opened 2019-04-11
Fix #948 make component registry immutable

### LINKED ISSUE #948: Discussion: Making container and scopes truly immutable after build
I really like that Autofac discourages from modifying built container/scope without creating a new scope but currently it can be easily circumvented (admittedly in a ugly way):
```c#
using (var container = builder.Build())
{
    container.ComponentRegistry.Register(RegistrationBuilder.ForType<A>().CreateRegistration());
    var a = container.Resolve<A>();
    using (var scope = container.BeginLifetimeScope())
    {
        scope.ComponentRegistry.Register(RegistrationBuilder.ForType<B>().CreateRegistration());
        var b = container.Resolve<B>();
    }
}
```
I haven't looked at how much effort it might take but would you consider tightening this in a such way so that `ComponentRegistry` cannot be modified directly?

Two possible approaches I could think of on top of my head are to either make `ComponentRegistry` property or its `Register*` methods internal. Obviously a breaking change but maybe something that can be marked as obsolete as get removed over time?

weelink:
Make `IComponentRegsitry` immutable by moving all mutable actions to `IComponentRegistryBuilder`.


weelink:
### Notes

I have assumed that this PR will be merged into a version where `IContainerBuilder.Update` will no longer be needed (#811)

- Moved all mutable methods to `IComponentRegistryBuilder`.
- `CopyOnWriterRegistry` is not necessary anymore. It only did something when the `IComponentRegistry` of a lifetime scope is modified. Since this PR disallows it, it has been removed.
- `IComponentRegistry.RegistrationsFor` is quering the registration sourced, that may add new registrations to the existing component registry, making this not a pure immutable registry. That is why I added a `IRegisteredServicesTracker` to share the registrations in the builder and the registrations from registration sources.

tillig:
Thanks for this! It will take a while to get through since a lot has changed, and it'll also require some time to think through all the use cases. I'm excited to dive in and see how you solved one of the larger challenges on the backlog. I'd like to make sure @alexmg has a chance to check it out, too, before any final decision gets made.

Just to set up expectations that it probably won't happen, like, _tomorrow_ or something. Appreciated, not forgotten, just not immediate. Yay for "real life" work taking over, right? ;)

tillig:
I haven't forgotten! I've pulled the PR down to my local machine and I'm looking at it now. For the most part, I dig it. I'm trying to mess around with the changes from a consumer perspective and see what's different, running the benchmarks to see how perf is affected, if at all.

That's actually where I've gotten stuck. The `ConcurrencyBenchmark` effectively never finishes for me. That's not a problem in this branch, the same thing happens on the `develop` branch for me. I think it's hammering through the "1,000 tasks each trying to concurrently resolve 10,000 object graphs" that it's going. There's not any visual feedback on the execution progress so it's hard to say, but I do see my RAM get pegged and I see the `Autofac.Benchmarks.ConcurrencyBenchmark` log gets created and sits at 0 bytes.

I'll have to work through that so I can see what the changes mean as far as performance. It may require some additional benchmarks to exercise areas like lifetime scope creation and resolution from scopes. Most of the benchmarks currently resolve directly from a container, which is good to exercise certain subsystems (like decorators) but it doesn't give us insight into more hierarchical systems that are more real-world.

Not to say that's a task for this PR, just keeping things up to date. If I update the benchmarks in the develop branch, I may do a pull from that branch into this PR to keep things apples-to-apples. That'd also resolve some of the recent merge conflicts due to me moving things around in the unit tests. :)

weelink:
Sure, no problem. I'll just check in once in a while.

tillig:
OK, here's what I did:

- Pulled this PR into a `pr-immutable-container` branch.
- Merged the latest `develop` changes into that branch.
- Pushed that branch to the main repo.

What this effectively does is "take the PR" but allow us to test some things like adding more benchmarks without holding this up or leaving it open forever. It also lets us target it for a 5.0 release and plan around when it should be brought into develop without holding up any minor changes that need to happen first. Likely this will end up being the "breaking changes for 5.0" branch. ;)

This is definitely one of the largest and most sweeping changes I've seen come in a PR, and I can't overstate how much I appreciate the work that happened here. I like what I see and I think if anything needs to change we can do whatever minor massaging needs to happen right in the branch.

So... thanks! 🎉

src/Autofac/Core/Registration/ComponentRegistryBuilder.cs
```diff
@@ -0,0 +1,184 @@
...
+{
+    /// <summary>
+    /// Used to build a <see cref="IComponentRegistry" />.
+    /// </summary>
+    internal class ComponentRegistryBuilder : Disposable, IComponentRegistryBuilder
+    {
+        /// <summary>
+        /// Protects instance variables from concurrent access.
+        /// </summary>
+        private readonly object _synchRoot = new object();
```
nblumhardt: Really amazing, first-class PR :-) !!! 👏 🎉 

Just one (late) question - does this class need synchronization now? The only times registry builders are mutated are during initial container build-up (required to be single-threaded) and during scope creation (also single-threaded).

src/Autofac/Core/Registration/ComponentRegistryBuilder.cs
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+        /// <summary>
+        /// Gets the set of properties used during component registration.
+        /// </summary>
+        /// <value>
+        /// An <see cref="IDictionary{TKey, TValue}"/> that can be used to share
+        /// context across registrations.
+        /// </value>
+        public IDictionary<string, object> Properties { get; }
```
nblumhardt: For example, regarding synchronization, `Properties` here will allow the underlying `Dictionary<,>` to be accessed outside the `_syncRoot` lock, which is potentially non-threadsafe.

(I'm happy to attempt a PR for these things but just thought I should confirm my understanding first.)

src/Autofac/Core/Registration/ComponentRegistryBuilder.cs
weelink: Thank you :-).
Now that I look at this again, I think you are right and this can be removed. I don't remember why I left it in, or if I actually had a reason to leave it in.

src/Autofac/Core/Registration/ComponentRegistryBuilder.cs
weelink: That is true. I kept this as it is now. This is something that could be improved.

src/Autofac/Core/Registration/ComponentRegistryBuilder.cs
nblumhardt: Thanks for the reply! @alexmg @tillig I'll PR some changes through around this and `Properties` if it sounds sane to you both :-)

tillig: Sounds great! This is all in the pr-immutable-container branch. I still haven't had a chance to really dig in due to some personal stuff coming up unexpectedly.

alexmg:
Great PR @weelink. Thanks for taking the time to put this together.

I have a branch with changes to get the `Service` being resolved into the resolution pipeline. This is also a breaking change so I have put it off for a while. I'm wondering if we should create a `v5.0` branch and start merging breaking changes into that from various feature branches such as this.

nblumhardt:
Would be a nice target for the `Properties` changes I am mulling over 👍 

Is there a plan to introduce new optimisations now that component registries are immutable? (Intrigued where all this is going to land - lots of possibilities :-))

alexmg:
@weelink I just wanted to let you know this has now been incorporated into the `develop` branch. Thanks again for the awesome PR. We really appreciate the effort you put into this.

PR #994, opened 2019-06-11
Faster RegisterAssemblyTypes

RogerKratz:
We use autofac in lots of our tests where a new container is created for each test, so fast container creation is important for us. When looking at what takes time for us, it's the "IsCompilerGenerated" call which internally looks for custom attributes.

Perf dump from our env when we run some of our tests...
image

This fix simply moves this check to after "the filters". In most cases this should be more optimal (not that often there's _lots_ of types marked with [CompilerGenerated] anyhow).

The "test"/example in this PR is about 80% faster after this fix.


tillig:
LGTM. Thanks! We don't get a ton of PRs and we appreciate greatly those we get!

RogerKratz:
@tillig 
Great! A new release would be very much appreciated! :)

PR #1005, opened 2019-07-23
Provide the Service being resolved to ResolveComponent method

alexmg:
A definite breaking changes here so will be targeted for `5.0`.

The main API change is in `IComponentContext.ResolveComponent` and by extension also `Container.ResolveComponent` and `LifetimeScope.ResolveComponent`. An additional `Service` parameter has been added to the method signature. We have a `Service` in almost all entry resolution entry points but lose it at the last moment inside `TryResolveService` after we find the appropriate `IComponentRegistration`.

I don't think the `ResolveComponent` method would be called directly very often, or in the odd case where it was being called, the appropriate `IComponentRegistration` was probably found by looking for a `Service`.

The most recent need for this is working in the new decorators. Knowing what `Service` is being resolved makes a few things a little easier and has allowed me to fix some of the recently raised issues.

### DIFF
--- a/src/Autofac/Autofac.csproj
+++ b/src/Autofac/Autofac.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <Description>Autofac is an IoC container for Microsoft .NET. It manages the dependencies between classes so that applications stay easy to change as they grow in size and complexity.</Description>
-    <VersionPrefix>4.9.3</VersionPrefix>
+    <VersionPrefix>5.0.0</VersionPrefix>
     <TargetFrameworks>netstandard2.0;netstandard1.1;net45</TargetFrameworks>
     <LangVersion>latest</LangVersion>
     <NoWarn>$(NoWarn);CS1591</NoWarn>
--- a/src/Autofac/Autofac.csproj
+++ b/src/Autofac/Autofac.csproj
@@ -4,6 +4,7 @@
     <Description>Autofac is an IoC container for Microsoft .NET. It manages the dependencies between classes so that applications stay easy to change as they grow in size and complexity.</Description>
     <VersionPrefix>4.9.3</VersionPrefix>
     <TargetFrameworks>netstandard2.0;netstandard1.1;net45</TargetFrameworks>
+    <LangVersion>latest</LangVersion>
     <NoWarn>$(NoWarn);CS1591</NoWarn>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
--- a/src/Autofac/Core/Lifetime/LifetimeScope.cs
+++ b/src/Autofac/Core/Lifetime/LifetimeScope.cs
@@ -25,7 +25,6 @@
 
 using System;
 using System.Collections.Concurrent;
-using System.Collections.Generic;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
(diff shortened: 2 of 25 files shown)


src/Autofac/Core/Lifetime/LifetimeScope.cs
```diff
@@ -259,17 +259,8 @@ private ScopeRestrictedRegistry CreateScopeRestrictedRegistry(object tag, Action
...
-        /// <param name="registration">The registration.</param>
-        /// <param name="parameters">Parameters for the instance.</param>
-        /// <returns>
-        /// The component instance.
-        /// </returns>
-        /// <exception cref="Autofac.Core.Registration.ComponentNotRegisteredException"/>
-        /// <exception cref="DependencyResolutionException"/>
-        public object ResolveComponent(IComponentRegistration registration, IEnumerable<Parameter> parameters)
+        /// <inheritdoc />
+        public object ResolveComponent(Service service, IComponentRegistration registration, IEnumerable<Parameter> parameters)
```
tillig: This would be
`public object ResolveComponent(ResolutionContext context)`
or something like that.

```c#
public class ResolutionContext
{
  public Service Service { get; set; }
  public IComponentRegistration Registration { get; set; }
  public IEnumerable<Parameter> Parmameters { get; set; }
}
```

nblumhardt:
At first glance I thought `IResolveOperation` might be the place to stash the `RootService` - is that the kind of thing you mean, @tillig ?

I've thought about this change in the past, it'll take me a little while to get my bearings, the main thing that held me back IIRC is that it could introduce problems if an implementer of a service makes any dynamic decisions based on the value of the root service. Switching a component from transient &rarr; singleton, or a resolve operation from `T` &rarr; `Lazy<T>` might have more subtle systemic effects if dynamic behavior based on the root service can creep in.

tillig:
@nblumhardt Maybe, yeah. Just thinking about a way to encapsulate the set of contextual parameters - service being resolved, registration being activated, etc. - such that if we need to add something else, maybe some arbitrary properties flowing down the line or whatever, we wouldn't have to add another parameter to the methods all the way down the stack. If `IResolveOperation` is the place to do that, then... that.

src/Autofac/Core/Lifetime/LifetimeScope.cs
alexmg: I think we spoke about the parameter object some time after I created this branch. I definitely prefer that approach, but thought I would start with this to get some feedback before making further changes.

alexmg:
@nblumhardt In this case I'm not trying to pass the original or root `Service` all the way through. This would be possible with a parameter object, but potentially dangerous as you suggested, and I haven't had the need. At the moment subsequent resolve operations are passed the `Service` that is relevant to that particular resolution. It was not knowing which `Service` was being requested for the current `IComponentRegistration` that was making things difficult.

alexmg:
I originally thought about using an `AsyncLocal<T>` with stack approach to avoid changing any signatures at all, but thought that might increase the chances of hitting edge cases.

Have you hit any such issues with `LogContext` in Serilog @nblumhardt?

tillig:
I feel like global statics are a magic thing waiting to bite you. It'd be nice to just pass parameters if possible.

nblumhardt:
Ah I see, thanks for the clarification @alexmg - I completely misunderstood the goal 👍 - I think the considerations regarding composability still apply somewhat (footguns for multi-service components that respond dynamically) but overall it's a lot simpler to reason about this feature than what I thought you meant.

I'm with you and @tillig on statics/thread statics/async local - well worth avoiding if at all possible. The async local `LogContext` in Serilog has been a big time sink over the years, particularly when running on the earlier .NET Framework versions.

alexmg:
I've introduced a new parameter object called `ResolveRequest`. I started with `ResolutionContext` but adding that to `IComponentContext` caused naming to get weird fast.

Before we pull this in I'm going to see how many decorator fixes can be done with and without this to judge if loading the footgun is definitely worthwhile.

src/Autofac/Autofac.csproj
```diff
@@ -4,6 +4,7 @@
     <Description>Autofac is an IoC container for Microsoft .NET. It manages the dependencies between classes so that applications stay easy to change as they grow in size and complexity.</Description>
     <VersionPrefix>4.9.3</VersionPrefix>
     <TargetFrameworks>netstandard2.0;netstandard1.1;net45</TargetFrameworks>
+    <LangVersion>latest</LangVersion>
```
tillig: [This is neat](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version) and something I didn't know about. 👍

tillig:
I like the new `ResolveRequest`. I think that solves the issue nicely and will help if we need to add more to the stack.

Is there any need for a corresponding `ResolveResponse` sort of construct? I can't think of anything immediately, just thinking out loud. Metadata around what happened during the resolution? A stack of types or registrations that were resolved for easier troubleshooting or better diagnostics or something? Timings?

Actually, timings makes me think again about putting the .NET Core diagnostic listener stuff in here so we can have events that folks can intercept specifically around diagnostics. Off topic, but, again, just thinking out loud.

PR #1011, opened 2019-08-01
Solve #Issue 1007

RaymondHuy:
This problem happens when we registered Func<IService> and also registered decorator for IService, then when we call resolve for Func<IService>: the decorator will be returned although the expected output should be delegate function of Func<IService>

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,45 @@
+# CODE OF CONDUCT
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as contributors and maintainers pledge to making participation in our project and our community a harassment-free experience for everyone, regardless of age, body size, disability, ethnicity, sex characteristics, gender identity and expression, level of experience, education, socio-economic status, nationality, personal appearance, race, religion, or sexual identity and orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment include:
+
+- Using welcoming and inclusive language
+- Being respectful of differing viewpoints and experiences
+- Gracefully accepting constructive criticism
+- Focusing on what is best for the community
+- Showing empathy towards other community members
+
+Examples of unacceptable behavior by participants include:
+
+- The use of sexualized language or imagery and unwelcome sexual attention or advances
+- Trolling, insulting/derogatory comments, and personal or political attacks
+- Public or private harassment
+- Publishing others’ private information, such as a physical or electronic address, without explicit permission
+- Other conduct which could reasonably be considered inappropriate in a professional setting
+
+## Our Responsibilities
+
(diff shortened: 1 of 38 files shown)


nblumhardt:
Hi Raymond,

I think the question was more whether the decorator or the manually-registered component should be returned by default; both are still valid components, so I don't think we should special-case and exclude delegates this way.

(The behavior described in the issue is a consequence of how the registration source mechanism works, I think changing it would require some substantial reorganizing of the code. A good alternative starting point for a change, if one is possible, would be going back to the rules determining which registration becomes the default in what circumstances, and work forward from there.)

HTH!

RaymondHuy:
Hi @nblumhardt I would like to explain about my PR :) There are 2 points:
```csharp

builder.RegisterType<ImplementorA>()
           .As<IDecoratedService>();

builder.RegisterDecorator<IDecoratedService>((_, __, inner) => new DecoratorA(inner));  //line 1

builder.Register<Func<IDecoratedService>>(ctx => () => new ImplementorB());  //line 2

var func = container.Resolve<Func<IDecoratedService>>();   //line 3

var instance = func();   //line 4

Assert.IsType<ImplementorB>(instance);
````
First point as a framework API developer (don't need to look detail in internal implementation), I think that when I have registered Func<IDecoratedService> directly in line 2, I should have this component returned in line 3 instead of a decorator component from line 1, somehow it seems those codes aren't predictable. Moreover it doesn't make sense to decorate for delegate type. The main thing we should decorate is the function executed by delegate instead of delegate.

Second point I would like to go detail into the internal implementation when line 3 is executing:
 - Step 1: The first component found for Func<IDecoratedService> is the component registered is line 2 (that is right)
-  Step 2: The autofac system will try to find the decorators available for this component. It has added a new Component with DelegateActivator as an available decorator:
```csharp
{(c, p) => () => Convert(c.ResolveComponent(
new ResolveRequest(Decorator (IDecoratedService), #component#, new [] {})), IDecoratedService)}
````
#component# in the above code is the component registered in line 1. This new component is returned to func after line 3 is executed.
- Step 3: When line 4 is executing: the container will try to resolve the #component# This #component# required parameters (as you see in line 2) but the resolve request doesn't have it included so the exception is thrown in https://github.com/RaymondHuy/Autofac/blob/Issue1007/src/Autofac/RegistrationExtensions.cs#L1388
Therefore the new Component added automatically in step 2 is the root cause for this bug, I have fixed by remove the decoration process for delegate type. https://github.com/RaymondHuy/Autofac/blob/Issue1007/src/Autofac/Core/Registration/ComponentRegistry.cs#L268

Feel free to ask me if you have questions ;)

alexmg:
Thanks for jumping in @RaymondHuy. I believe this should be fixed with #1023. Please let me know if this is not the case.

Issue #879, opened 2017-10-27
Make IDisposable tracking configurable per builder/scope


eugbaranov:
Sometimes it is much easier to leave GC to deal with `IDisposable` objects than refactoring the entire application to create separate lifetimes per units of work.

Something like this maybe:
```csharp
builder.AutomaticDisposal = false;
```

tillig:
I assume this is so you don't have to register everything as `ExternallyOwned()`?

Regent:
Indeed, effectively having an option to specify whether default option should be `ExternallyOwned` or else.
Maybe an enum allowing to specify whether it should be `ExternallyOwned`, `InstancePerOwned` or `OwnedByLifetimeScope` by default?

tillig:
Just brainstorming since I'm out and only have my phone right now... I wonder if this could be done with a module and `AttachToComponentRegistration`...

Regent:
But wouldn't `registration.Ownership` be read-only at that point?
Also, will it allow to change behavior per scope?

tillig:
As I said, I'm out and in my phone. Feel free to test this out and see what happens. Maybe it works, maybe it doesn't. In was trying to unstick you. If you choose not to try it out, well, that's fine, too.

Regent:
It tried it briefly but couldn't find how to make it work, that's said I'm not a big expert in Autofac, maybe I'm missing something...

johneking:
Looks like this isn't supported with the existing code. There are two distinct features we could add here I think, but I'm just trying to get my head around whether providing support for each set of use-cases is really desirable. Units of work seem somewhat fundamental to using scopes properly and I wonder if the ability to disable ownership across the board via a built-in feature might encourage mis-use (/laziness!)

The simple part would be adding the default ownership on the registration side - that could be done relatively easily by adding an optional parameter to the constructors of `RegistrationData`, `RegistrationBuilder` and `ContainerBuilder`, and conceptually it does seem to make sense that those classes would have that. The `ContainerBuilder` default could then perhaps be passed to the other objects. But then it does bloat out those constructors a bit. Is it worth adding that in just to make the syntax for having a lot of `ExternallyOwned` calls on `IDisposable` types neater? I'm not convinced, but can submit a PR if so.

[...]

PR #1014, opened 2019-08-14
Issue 879 - Add default InstanceOwnership to ContainerBuilder

johneking:
This is a draft PR for a somewhat crude way of setting the default ownership (owned by lifetime scope or externally owned) for registrations as per #879 . I'm still not 100% convinced this is a valuable addition but have put this up just to show what it might look like.

It expands out the constructors for `RegistrationData`, `RegistrationBuilder` and `ContainerBuilder` to accept an optional ownership parameter so generally would be backwards compatible. The `ContainerBuilder` then passes that default to any calls made on it to create new registrations.

An interesting related distinction I noticed is that `CollectionRegistrationSource` currently defaults to `ExternallyOwned` but all other registration sources default to `OwnedByLifetimeScope`. That's likely immaterial anyway and is unaffected by this change, but not sure if it's by design.

### DIFF
--- a/src/Autofac/Builder/RegistrationBuilder.cs
+++ b/src/Autofac/Builder/RegistrationBuilder.cs
@@ -54,13 +54,17 @@ namespace Autofac.Builder
         /// </summary>
         /// <typeparam name="T">Instance type returned by delegate.</typeparam>
         /// <param name="delegate">Delegate to register.</param>
+        /// <param name="ownership">Initial <see cref="InstanceOwnership"/> for the registration,
+        /// determining whether components implementing IDisposable are disposed by Autofac at the end of a unit of work.
+        /// Defaults to OwnedByLifetimeScope.</param>
         /// <returns>A registration builder.</returns>
-        public static IRegistrationBuilder<T, SimpleActivatorData, SingleRegistrationStyle> ForDelegate<T>(Func<IComponentContext, IEnumerable<Parameter>, T> @delegate)
+        public static IRegistrationBuilder<T, SimpleActivatorData, SingleRegistrationStyle> ForDelegate<T>(Func<IComponentContext, IEnumerable<Parameter>, T> @delegate, InstanceOwnership ownership = InstanceOwnership.OwnedByLifetimeScope)
         {
             return new RegistrationBuilder<T, SimpleActivatorData, SingleRegistrationStyle>(
                 new TypedService(typeof(T)),
                 new SimpleActivatorData(new DelegateActivator(typeof(T), (c, p) => @delegate(c, p))),
-                new SingleRegistrationStyle());
+                new SingleRegistrationStyle(),
+                ownership);
         }
 
         /// <summary>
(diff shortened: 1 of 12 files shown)


nblumhardt:
Hi @johneking ! Implementation-wise the PR looks good. Feature-wise, I've steered away from this approach in the past - I think it could lead to a reduction in readability and composability.

Regarding readability, as a long-time Autofac user I like being able to scan registration code and understand exactly what the resulting components look like. This is weakened if we start adding configurable defaults at the container builder level. It's not a massive reduction in readability - but it introduces the same kind of intellectual overhead that things like operator overloading introduce in programming languages: over time, it gets harder to understand what a line of code is doing without looking at other bits of code. This is probably a weak, slippery-slope, argument against, but worth a thought.

Composability is the bigger issue for me - I don't think modules (or chunks of reusable code) in separate projects or assemblies should change semantics based on global configuration. If I write an Autofac module for DB connection handling that assumes container-ownership, it shouldn't be broken if it's loaded into a container builder configured with incompatible defaults. Within a module, at least, registration semantics should be fixed and consistent, otherwise reusability of modules would be reduced.

My $0.02 - hopefully some food for thought, anyway! :+1:

johneking:
Ah yeah, both good points, 100% agree. Sounds like making the default module-level one way or another should be clearer - I'll take another look.

johneking:
@nblumhardt I've kept the bulk of the previous changes in place (extra constructor parameters on `RegistrationData` etc.), including the additional configurability for the `ContainerBuilder`, but made it only available internally. That's now used directly + exclusively within `Module.Configure`, based on a new virtual property on the `Module`.

So the usage would now be more like:
```
public class ExternallyOwnedModule : Module
{
	protected override InstanceOwnership DefaultInstanceOwnership => InstanceOwnership.ExternallyOwned;

	protected override void Load(ContainerBuilder builder)
	{
		builder.RegisterType<object>();
	}
}
```

That seems to address the composability concern, and at least partly deals with readability too. There is still an additional overhead in being aware of a context in which the registrations are made, but it should be pretty clear within each module what that context is. What do you think?

tillig:
First, if we support something like this, great work in getting a POC out there. This looks great.

> I'm still not 100% convinced this is a valuable addition but have put this up just to show what it might look like.

Honestly, I think this is what worries me the most about this feature. I feel like this is offering folks a foot-gun and troubleshooting problems waiting to happen. I can think of a whole myriad of new challenges for folks with StackOverflow questions like:

- I'm a member of a large group where one team writes the application host and I only write modules for the host. I just took an update of our application host and now I'm seeing a massive unexpected memory leak. What's wrong with Autofac? [App host team changed the default behavior and didn't notify anyone.]
- I know I want to control all disposal myself but when I switched that on, suddenly there are a ton of memory leaks. What's wrong with Autofac? [I didn't realize how many places I'd have to add disposal that I wasn't expecting to have to update.]

I don't actually know of any other IoC container that offers the ability to default to "I control all disposal." In some cases, you can't actually opt out of container-controlled disposal. I kind of feel like the folks asking for this don't really realize what they're asking for.

My boss had a good analogy for this.

> A guy walks into a mobile phone store and demands to see their selection of antennas. Store clerk tells him they don't sell antennas. Guy is adamant, wants a bigger antenna. "Why do you want a bigger antenna?" asks the clerk.
>
> "I keep dropping calls, I need to make sure when I go into various remote areas, I don't drop calls."
>
> "Oh," says the clerk, "you don't want a _bigger antenna_, you want _better reception_."

Folks are asking for the solution to a problem but haven't really explained the problem. If you have an application where you're trying to retrofit some DI into it and you already have code that disposes of dependencies... great. Calling dispose on components twice shouldn't cause any issue. If it does, _fix that_. Updating a legacy app and you haven't quite got unit of work / request scope sort of semantics in there yet? Add that first, rather than fighting the container disposal. If there's an object graph that you want to control, `Owned<T>` handles that. There are different (better?) ways to handle most of these things and let the container do what it's there to do.

If other folks on the team think this is valuable, I guess go for it; however, I think there's some value in providing a "pit of success" for folks to fall into by making it hard to do the wrong thing.

johneking:
Thanks for the feedback - and yep, as per my initial [comment on the issue](https://github.com/autofac/Autofac/issues/879#issuecomment-519139510), I tend to agree. Even if we can address composability and readability in the implementation, it does seem like the feature would largely just help steer people away from best practice for not much of a gain. The scenario where this would be required at all really seems to me to be quite a bit of an edge-case: many types (enough to make the existing `ExternallyOwned()` syntax inconvenient) requiring custom disposal, but where the timing of the disposal can't be aligned with a `LifetimeScope` or controlled via `Owned<T>` for some reason.
I'll close this PR unless anyone has any objections, and suggest we do the same with the Issue - will add a comment there.

PR #1042, opened 2019-11-01
Move registration source extension method to a dedicated class

eugbaranov:
As previously discussed in #940, submitting this with breaking changes.

### DIFF
--- a/src/Autofac/RegistrationExtensions.cs
+++ b/src/Autofac/RegistrationExtensions.cs
@@ -64,19 +64,6 @@ namespace Autofac
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


alexmg:
Thanks @eugbaranov!

alexmg:
I just noticed this was targeting the `pr-immutable-container` branch. We are working through some performance issues in that branch before merging it into `develop` (which is now targeting `5.0.0`). Would you be able to submit a PR for the `develop` branch?

eugbaranov:
Haven't realised develop is now targeting 5.0.0 - will do.

PR #1045, opened 2019-11-05
Fix Issue#1040: IsRegistered throws IndexOutOfRangeException for open generic type

RaymondHuy:


nblumhardt:
It might be more succinct to use `TypeInfo.IsGenericTypeDefinition` to test the service type, here; e.g.:

```
> var t = typeof(List<int>);
> var u = typeof(List<>);
> t.IsGenericType
true
> t.IsGenericTypeDefinition
false
> u.IsGenericType
true
> u.IsGenericTypeDefinition
true
>
```

(Apologies if I've overlooked platform compat requirements :-))

RaymondHuy:
Thanks @nblumhardt I have just updated my PR because your suggestion made code more readable.

nblumhardt:
Thanks! I think there's still one possible improvement - the check can move to the condition a few lines up, reflecting what I think the `if` clause is trying to assert:

```csharp
if (swt != null && swt.ServiceType.GetTypeInfo().IsGenericType && !swt.ServiceType.IsGenericTypeDefinition)
```

tillig:
Looks like if we can move that check then we can merge this bad boy. Thanks for the PR!

PR #1052, opened 2019-11-25
Fix Issue#1030

### LINKED ISSUE #1030: KeyFilterAttribute behavior different from AutowiringParameter
When there the parameter type is not registered in the DI, the `KeyFilterAttribute` behavior is different from the `AutowiringParameter` behavior.

# Examples

## "Normal" required parameter

```csharp
class Test
{
    public Test(int parameter) => Console.WriteLine($"Test(parameter = {parameter})");
}

cb.RegisterType<Test>()
  .WithAttributeFiltering()
  .As<Test>();
cb.Build().Resolve<Test>();
```

### Output
Throws:
```
Autofac.Core.DependencyResolutionException: An error occurred during the activation of a particular registration. See the inner exception for details. Registration: Activator = Test (ReflectionActivator), Services = [Submission#46+Test], Lifetime = Autofac.Core.Lifetime.CurrentScopeLifetime, Sharing = None, Ownership = OwnedByLifetimeScope ---> None of the constructors found with 'Autofac.Core.Activators.Reflection.DefaultConstructorFinder' on type 'Submission#46+Test' can be invoked with the available services and parameters:
Cannot resolve parameter 'Int32 parameter' of constructor 'Void .ctor(Int32)'. (See inner exception for details.)
```

## "Normal" optional parameter

```csharp
class Test
{
    public Test(int parameter = 1) => Console.WriteLine($"Test(parameter = {parameter})");
}

cb.RegisterType<Test>()
  .WithAttributeFiltering()
  .As<Test>();
cb.Build().Resolve<Test>();
```

### Output
Prints:
```
Test(parameter = 1)
```

## Required parameter with `KeyFilter`

```csharp
class Test
{
    public Test([KeyFilter(0)] int parameter) => Console.WriteLine($"Test(parameter = {parameter})");
}

cb.RegisterType<Test>()
  .WithAttributeFiltering()
  .As<Test>();
cb.Build().Resolve<Test>();
```

### Output
Prints:
```
Test(parameter = 0)
```

### Expected Output?
Throws:
```
Autofac.Core.DependencyResolutionException: ...
```

## Optional parameter with `KeyFilter`

```csharp
class Test
{
    public Test([KeyFilter(0)] int parameter = 1) => Console.WriteLine($"Test(parameter = {parameter})");
}

cb.RegisterType<Test>()
  .WithAttributeFiltering()
  .As<Test>();
cb.Build().Resolve<Test>();
```

### Output
Prints:
```
Test(parameter = 0)
```

### Expected Output?
Prints:
```
Test(parameter = 1)
```

# Detailed analysis

The `AutowiringParameter` [checks whether the parameter type is registered or not](https://github.com/autofac/Autofac/blob/6dde84e5b0a3f82136a0567a84da498b04e1fa2d/src/Autofac/Core/Activators/Reflection/AutowiringParameter.cs#L55). If not, the the `AutowiringParameter` returns `false` in the `CanSupplyValue` method.

The `WithAttributeFiltering` method [will allows return `true` in the `CanSupplyValue` if a `ParameterFilterAttribute` is found](https://github.com/autofac/Autofac/blob/6dde84e5b0a3f82136a0567a84da498b04e1fa2d/src/Autofac/Features/AttributeFilters/RegistrationExtensions.cs#L69). 

Here is a table summarizing the returned value of `Parameter.CanSupplyValue` method:

|                                    | `AutowiringParameter` | `KeyFilterAttribute` |
| ---------------------------------- | --------------------- | -------------------- |
| Registered                         | true                  | true                 |
| Not registered, Required parameter | false                 | true                 |
| Not registered, Optional parameter | false                 | true                 |

# Question

Is it the expected behavior?
If so, why is it preferred over consistency with "standard" parameters?

RaymondHuy:
Fix #1030: Change `ParameterFilterAttribute.ResolveParameter` method signature. Enhanced `KeyFilter` attribute: when key is not found the instance can not be resolved (it also has been mentioned in SO: https://stackoverflow.com/questions/52948278/how-can-i-prevent-autofac-from-injecting-null-when-accidentally-used-an-invalid)

### DIFF
--- a/src/Autofac/Core/ResolvedParameter.cs
+++ b/src/Autofac/Core/ResolvedParameter.cs
@@ -34,21 +34,16 @@ namespace Autofac.Core
     /// </summary>
     public class ResolvedParameter : Parameter
     {
-        private readonly Func<ParameterInfo, IComponentContext, bool> _predicate;
-        private readonly Func<ParameterInfo, IComponentContext, object> _valueAccessor;
+        private readonly Func<ParameterInfo, IComponentContext, (bool canResolve, Func<object> valueAccessor)> _resolveParameter;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="ResolvedParameter"/> class.
         /// </summary>
-        /// <param name="predicate">A predicate that determines which parameters on a constructor will be supplied by this instance.</param>
-        /// <param name="valueAccessor">A function that supplies the parameter value given the context.</param>
-        public ResolvedParameter(Func<ParameterInfo, IComponentContext, bool> predicate, Func<ParameterInfo, IComponentContext, object> valueAccessor)
+        /// <param name="resolveParameter">A predicate that determines which parameters on a constructor will be supplied
+        ///     by this instance and a function that supplies the parameter value given the context.</param>
+        public ResolvedParameter(Func<ParameterInfo, IComponentContext, (bool canResolve, Func<object> valueAccessor)> resolveParameter)
         {
-            if (predicate == null) throw new ArgumentNullException(nameof(predicate));
-            if (valueAccessor == null) throw new ArgumentNullException(nameof(valueAccessor));
-
-            _predicate = predicate;
-            _valueAccessor = valueAccessor;
+            _resolveParameter = resolveParameter;
         }
 
         /// <summary>
@@ -59,15 +54,17 @@ namespace Autofac.Core
         /// <param name="valueProvider">If the result is true, the valueProvider parameter will
         /// be set to a function that will lazily retrieve the parameter value. If the result is false,
         /// will be set to null.</param>
-        /// <returns>True if a value can be supplied; otherwise, false.</returns>
+        /// <returns>true if a value can be supplied; otherwise, false.</returns>
         public override bool CanSupplyValue(ParameterInfo pi, IComponentContext context, out Func<object> valueProvider)
         {
             if (pi == null) throw new ArgumentNullException(nameof(pi));
             if (context == null) throw new ArgumentNullException(nameof(context));
 
-            if (_predicate(pi, context))
+            (bool canResolve, Func<object> valueAccessor) = _resolveParameter(pi, context);
+
+            if (canResolve)
             {
-                valueProvider = () => _valueAccessor(pi, context);
+                valueProvider = valueAccessor;
                 return true;
             }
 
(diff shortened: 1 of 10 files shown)


src/Autofac/Features/AttributeFilters/MetadataFilterAttribute.cs
```diff
@@ -160,9 +161,11 @@ public override object ResolveParameter(ParameterInfo parameter, IComponentConte
             var elementType = GetElementType(parameter.ParameterType);
             var hasMany = elementType != parameter.ParameterType;
 
-            return hasMany
+            value = hasMany
                 ? FilterAllMethod.MakeGenericMethod(elementType).Invoke(null, new[] { context, Key, Value })
                 : FilterOneMethod.MakeGenericMethod(elementType).Invoke(null, new[] { context, Key, Value });
+
+            return true;
```
tillig: If this is a `Try` method... is there any point at which it'd return false? Or is it still "success or throw?"

src/Autofac/Core/ResolvedParameter.cs
```diff
@@ -34,21 +34,16 @@ namespace Autofac.Core
...
 
         /// <summary>
         /// Initializes a new instance of the <see cref="ResolvedParameter"/> class.
         /// </summary>
-        /// <param name="predicate">A predicate that determines which parameters on a constructor will be supplied by this instance.</param>
-        /// <param name="valueAccessor">A function that supplies the parameter value given the context.</param>
-        public ResolvedParameter(Func<ParameterInfo, IComponentContext, bool> predicate, Func<ParameterInfo, IComponentContext, object> valueAccessor)
+        /// <param name="resolveParameter">A predicate that determines which parameters on a constructor will be supplied
+        ///     by this instance and a function that supplies the parameter value given the context.</param>
+        public ResolvedParameter(Func<ParameterInfo, IComponentContext, (bool canResolve, Func<object> valueAccessor)> resolveParameter)
```
tillig: I'm sort of wrapped around the axle on this parameter type.

`Func<ParameterInfo, IComponentContext, (bool canResolve, Func<object> valueAccessor)> resolveParameter`

I get that it sort of wraps a `Try` mechanism into a single line, but... man. I'm not a fan, I have to say. This seems super unusable.

The intent of the `predicate` there was to allow a `Try` without having the weird extra thing. Was there no way to just keep that instead?

RaymondHuy:
Hi @tillig  Thanks for your comment. I see this approach is not good so I have created another PR with different approach. Can you check it ?

PR #1054, opened 2019-12-03
Invoke Build Callbacks at Lifetime Scope Creation

### LINKED ISSUE #985: Enable build callbacks to run when registered in child lifetime scopes
When registering a Build Callback (i.e. `RegisterBuildCallback`) within a lifetime scope secondary registration do not get called.

I apologize if this is incorrect usage.

`packages.config`:
```xml
<?xml version="1.0" encoding="utf-8"?>
<packages>
  <package id="Autofac" version="4.9.2" targetFramework="net471" />
  <package id="xunit" version="2.4.1" targetFramework="net471" />
  <package id="xunit.abstractions" version="2.0.3" targetFramework="net471" />
  <package id="xunit.analyzers" version="0.10.0" targetFramework="net471" />
  <package id="xunit.assert" version="2.4.1" targetFramework="net471" />
  <package id="xunit.core" version="2.4.1" targetFramework="net471" />
  <package id="xunit.extensibility.core" version="2.4.1" targetFramework="net471" />
  <package id="xunit.extensibility.execution" version="2.4.1" targetFramework="net471" />
</packages>
```

```csharp
using System;
using Xunit;

namespace Autofac.Tests
{
    public class Tests
    {
        [Fact] // Passes successfully
        public void RootCallbackThrows_ExpectThrows()
        {

            var expected = new Exception();
            builder.RegisterBuildCallback(c => throw expected);

            var ex = Assert.Throws<Exception>(() => builder.Build());
            Assert.Same(expected, ex);
        }

        [Fact] // Failure - no exception thrown
        public void NestedCallbackFromDelegateThrows_ExpectThrows()
        {

            using (var outer = builder.Build())
            {
                var expected = new Exception();

                var ex = Assert.Throws<Exception>(
                    () => outer.BeginLifetimeScope(b => { b.RegisterBuildCallback(c => throw expected); }));
                Assert.Same(expected, ex);
            }
        }

        [Fact] // Failure - no exception thrown
        public void NestedCallbackFromModuleThrows_ExpectThrows()
        {

            using (var outer = builder.Build())
            {
                var expected = new Exception();

                var ex = Assert.Throws<Exception>(
                    () => outer.BeginLifetimeScope(b => { b.RegisterModule(new CallbackModule(expected)); }));
                Assert.Same(expected, ex);
            }
        }

        private sealed class CallbackModule : Module
        {
            private readonly Exception _expectedException;

            public CallbackModule(Exception expectedException)
            {
                _expectedException = expectedException;
            }

            protected override void Load(ContainerBuilder builder)
            {
                base.Load(builder);

                builder.RegisterBuildCallback(c => throw _expectedException);
            }
        }
    }
}
```

alistairjevans:
Following the discussion in #985, these changes allow build callbacks to execute at the point of lifetime scope creation if new ones have been registered.

So, this works:

```
var scope = container.BeginLifetimeScope(cfg =>
{
    cfg.RegisterBuildCallback(scope => { /* do something */ });
});
```

The callback will be invoked just prior to BeginLifetimeScope exiting, after any startable components are instantiated (same behaviour as the Container).

As discussed, we now register a component ``BuildCallbackService`` that contains the set of callbacks to register; nested scopes then add replacement registrations of that service.

``BuildCallbackManager`` is responsible for executing the callbacks.

Fixes #985


test/Autofac.Specification.Test/ContainerBuilderTests.cs
```diff
@@ -38,6 +38,121 @@ public void BuildCallbacksInvokedWhenRegisteredInModuleLoad()
...
         }
 
+        [Fact]
+        public void BuildCallbacksInvokedWhenRegisteredInBeginLifetimeScope()
+        {
+            var called = 0;
+
+            void BuildCallback(ILifetimeScope c)
+            {
+                Assert.False(c is IContainer);
```
tillig: Boo on xunit for not having `Assert.IsNotAssignableFrom<T>`

src/Autofac/Builder/BuildCallbackServiceResources.resx
```diff
@@ -0,0 +1,123 @@
...
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="BuildCallbacksAlreadyRun" xml:space="preserve">
+    <value>This set of build callbacks have already run.</value>
```
tillig: I think it's "This set ... _has_ already run." Yeah?

test/Autofac.Specification.Test/ContainerBuilderTests.cs
alistairjevans: Agreed, boo on xunit. Not sure if there's a better way to do it that raises a more useful xunit error...

src/Autofac/Builder/BuildCallbackServiceResources.resx
tillig: I've never been good at singular/plural when the singular is a plural. Like, a company, say Microsoft, is a singular entity. So "Microsoft has released a new version of Windows." But in some places a company is seen as a plural of individuals so it's "Microsoft have released a new version of Windows." I could go either way I guess.

src/Autofac/Builder/BuildCallbackService.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Linq;
+
+namespace Autofac.Builder
+{
+    /// <summary>
+    /// Represent a collection of callbacks that can be run after a container or scope is 'built'.
+    /// </summary>
+    internal class BuildCallbackService
+    {
+        private List<Action<ILifetimeScope>> Callbacks { get; set; } = new List<Action<ILifetimeScope>>();
```
tillig: Should this be a member variable rather than a property? Unclear if the compiler just optimizes that out anyway.

src/Autofac/Builder/BuildCallbackServiceResources.resx
alistairjevans: I've amended the English to 'has'; it is somewhat up for debate...

Perhaps I should change it to 'This set of build callbacks would have already run', just to mess with the tenses.

src/Autofac/Builder/BuildCallbackService.cs
alistairjevans: The compiler will optimise that to a directly field access, but it should just be a member; I had an older variant that allowed external access to the callbacks, but the property hung around after the change.

nullabilitynotes.md
```diff
@@ -0,0 +1,105 @@
+
```
tillig: I think this file may have snuck in as part of your nullable references work.

alistairjevans: Ah, yes. I was doing so well having both branches going for a while there.

PR #1055, opened 2019-12-03
bypass IComponentContext error

rkammerman:


tillig:
It would be good to open an issue to describe the challenge you're having (and include some sort of repro) prior to submitting a pull request. Thanks!

PR #1056, opened 2019-12-04
Add Nullable Reference Type support to Autofac

alistairjevans:
Ok, so there's quite a lot of change here, but there are little to no code behaviour changes. 

The outcome of these changes is that in a consuming library, I get useful warnings:

image

No doubt there'll a few things to discuss on this, but it get's us going!

# Notes

I've got a selection of comments to go over which provide background/explanation, so here we go:

## Targeting

I haven't had to change the target frameworks, since the attributes used for marking methods are switched out for our own internals in net461 and ns2.0; see the [NullableAttributes](https://github.com/alistairjevans/Autofac/blob/4186d0d728f7511078f6460c74c89bdc01d22769/src/Autofac/Util/NullableAttributes.cs) file.

The relevant warnings for the compiler are also disabled in the project file when compiling for those older targets.

## Try... Methods

Anywhere that we have a 'Try....' style method, we now apply [NotNullWhen(returnValue: true)] on the out parameter.  This should probably be enforced via code review in the future.

For example, the [TryResolve ](https://github.com/alistairjevans/Autofac/blob/4186d0d728f7511078f6460c74c89bdc01d22769/src/Autofac/ResolutionExtensions.cs#L919) method:
``` csharp
public static bool TryResolve<T>(this IComponentContext context, [NotNullWhen(returnValue: true)] out T? instance)
```

This annotation also only works if called directly inside an if:

``` csharp

var success = built.TryResolve<MyClass2>(out var instance2);

if (success)
{
    // Still get the warning here, even though it's fine
    var result2 = instance2.MyValue;
}
```

That's a bit annoying, but the roslyn people know about it.

## Registration Method Constraints

So, this is one of the bigger changes, but the limits on registration now constrain to the 'notnull' generic:

``` csharp
public static IRegistrationBuilder<TImplementer, ConcreteReflectionActivatorData, SingleRegistrationStyle>
     RegisterType<TImplementer>(this ContainerBuilder builder)
     where TImplementer : notnull
```

This was necessary to make the rest of the Autofac internals make sense; as far as I'm aware, the only impact for users of the library is if someone tries to use a ``Nullable<int>`` or something as a service.

I personally don't see why someone would do that in the first place.

## Comparison Overloads

This was something that confused me immensely until I realised what was going on; if you have a type, like ``Service``, that implements an overload for ``==``, then it must accept nullable types, otherwise you can't compare the object to null and things get a little weird:

```csharp
/// <summary>
/// Implements the operator ==.
/// </summary>
/// <param name="left">The left operand.</param>
/// <param name="right">The right operand.</param>
/// <returns>The result of the operator.</returns>
public static bool operator ==(Service? left, Service? right)
{
    return Equals(left, right);
}
```

## ResolveOptional Returns Nullable References

Any methods that may optionally return null values for generic TService types are now marked as nullable return, **but** it was also necessary to mark the constraint as ``class``, rather than ``notnull``.

``` csharp
public static TService? ResolveOptional<TService>(this IComponentContext context)
    where TService : class
{
    return ResolveOptional<TService>(context, NoParameters);
}
```

## Updated Roslyn Analyzers to Beta

Without the latest beta of the rosyln analyzers, I was getting stylecop failures on nullable arrays:

```
 private Type[]? setOfTypes; // Incorrect Stylecop failure.
```

So I've taken a reference to that beta.

## Uninitialised Members

The Roslyn compiler reports a class field as being uninitialized (and potentially null) if it can't tell it's been initialised at the end of the constructor.

This happens even if the constructor calls a method or property that always instantiates the field. 

You can set 'default!' or 'null!' on the field to get round this (not ideal, but explainable).

```csharp
// _successfulActivations can never be null, but the roslyn compiler doesn't look deeper than
// the initial constructor methods yet.
private List<InstanceLookup> _successfulActivations = default!;
```

There are a couple of open rosyln issues that are related to this, some discussion about adding flow analysis to check that constructors do in fact exit having initialised those properties:

- https://github.com/dotnet/roslyn/issues/39291
- https://github.com/dotnet/roslyn/issues/39090
- https://github.com/dotnet/roslyn/issues/37975

This is also a pain whenever I have a class that I want to instantiate using 'Property' notation, e.g. ``DecoratorContext``:

``` csharp
private DecoratorContext()  // warning on this line
{

}

internal static DecoratorContext Create(Type implementationType, Type serviceType, object implementationInstance)
{
    var context = new DecoratorContext
    {
        ImplementationType = implementationType,
        ServiceType = serviceType,
        AppliedDecorators = new List<object>(0),
        AppliedDecoratorTypes = new List<Type>(0),
        CurrentInstance = implementationInstance
    };
    return context;
}
```
In this particular case I switched to using a constructor and it was fine.

## Reflection-Only Cases

There's a couple of places where we create objects passing default for every constructor parameter, just to define an ``Expression``, which will never actually call the code and pass those nulls. In that case I've just specified the ``default!`` marker to tell the compiler it's fine:

```csharp
// The explicit '!' default is ok because the code is never executed, it's just used by
// the expression tree.
private static readonly ConstructorInfo RequestConstructor
    = ReflectionExtensions.GetConstructor(() => new ResolveRequest(default!, default!, default!));
```


tillig:
Wow, this is some crazy stuff, good work! And helpful notes, since this is definitely new territory.

> ...the limits on registration now constrain to the 'notnull' generic...
> This was necessary to make the rest of the Autofac internals make sense; as far as I'm aware, the only impact for users of the library is if someone tries to use a `Nullable<int>` or something as a service.

Hmmm. I'd guess this would also apply, then, to any `Nullable<T>` value type? That is, if I have a `struct` of any type that I need nullable, this would be affected? That may be problematic. I can't come up with a specific example, but things that are floating around in my mind are like how ASP.NET Core has gone through and done a lot more around memory optimization and allocations, caring where things get allocated (stack vs. heap) and there could be a need for `Nullable<SomeStructureValueType>` support.

What are our options here? Is this just kinda "how it is?" I'm sort of wading into speculation territory.

> Any methods that may optionally return null values for generic `TService` types are now marked as nullable return, but it was also necessary to mark the constraint as `class`, rather than `notnull`.

This overlaps a tiny bit with the above. `Nullable<T>` is a class, so that's fine... I think we've had some weird issues around resolving and value types anyway, like resolving an `int` that was registered. I'm not actually sure what `ResolveOptional` does when it's a value type like `int` that isn't registered. Zero (`default(T)`)?

I wonder if this all just solidifies that we can only really deal with reference types. We've sort of been on the fence where it _pretty much mostly works_ but I recall seeing issues and things in docs where we _sometimes sort of don't_ support them.

alistairjevans:
Yes, I believe any type of ``Nullable<TWhatever>`` would be affected. This is just compiler sugar right now, someone can still use the non-generic form to pass in a ``typeof(Nullable<int>)`` if they really need to (I've not put any runtime checks for reference types in).  I will look at this in more detail to understand the restrictions and impact on consumers some more.

Generally, I feel like it's not a great design to store a ``Nullable<TStruct>`` as a service anyway, it's probably more sensible to store the value inside a proper service class, or conceivably use ``Lazy<T>``. This may be an opportunity for Autofac to encourage people towards better design choices?

As for value types as services, the dotnet core DI system at least 'prefers' reference types (i.e. you can't register structs via a generic mechanism), judging from the implementation of the [ServiceCollectionServiceExtensions](https://github.com/aspnet/Extensions/blob/master/src/DependencyInjection/DI.Abstractions/src/ServiceCollectionServiceExtensions.cs):

```csharp
public static IServiceCollection AddTransient<TService, TImplementation>(this IServiceCollection services)
            where TService : class
            where TImplementation : class, TService
        {
```

alistairjevans:
Oh, also, I believe if we wanted to support Nullable<TStruct> in ResolveOptional, we can do that by adding overloads to explicitly accept a struct type, but the same method can't handle both.

tillig:
I think you're right, that this is likely just fine (and maybe a good thing). People do weird stuff they shouldn't do... but then, this is a 5.0 "breaking changes" thing, so maybe that's just fine. Stop doing _this particular_ weird thing. I'm sure we'll hear about it if there's someone out there who just can't live without it, sort of like folks who can't live without `Update` because of bad design.

alistairjevans:
Ok, good to hear. I'm mulling over whether I should enable nullable in the tests, but that will make it more difficult to write tests if I do it everywhere.

I may just add some new tests that have nullable on for the file, and add some tests that will fail if the Nullable attribute markers are not present where we expect them to be.

alistairjevans:
Ok, so I've added some tests to try and help us assert on how the compiler responds to us putting nullable behaviours on the externally facing methods.

The way I've done it however leaves me a bit unsure, because my literal brain answered the question "how do we check that our code will generate compiler warnings?", with "Let's compile the code and check!".

So, I've added a new Autofac.Tests.Compilation project, and in that project there are tests that literally compile code using Roslyn and assert we get errors, using the utility class ``AutofacCompile``, like so:

```csharp
[Fact]
public void NullableTypeNotAllowedForAsMethod()
{
   new AutofacCompile()
        .Body(
        @"
            var containerBuilder = new ContainerBuilder();
                
            containerBuilder.Register(c => new DerivedClass()).As<BaseClass?>();
        ")
        .AssertWarningEndsWith("Nullability of type argument 'Autofac.Test.Compilation.BaseClass?' doesn't match 'notnull' constraint.");
}
```

This approach works, I can pass/fail tests based on compiler results, and I feel like it's a pretty descriptive way to say 'this is how we want Autofac to respond to users trying to do this'.

I already found that I needed to add constraints to the ``As`` methods when my expectations weren't being met.

However, I also worry that I'm using a sledgehammer to crack a nut, and this is going to cause annoyance and difficulty somehow. 

Thoughts?

alistairjevans:
Interesting thing going on with registering enum types; while attempting to further constrain types that can be registered down to 'class', I found a test that specifies that enums can be registered for property auto-wiring.

I can somewhat see why someone might want to do that, to inject a global enum value that represents configuration? The test was added a long time ago, so can't find the reason for the requirement.

If I leave the constraints at notnull, then you can use Enum registrations, but then behaviour can be slightly odd where you can register non-null value types but not nullable ones.

tillig:
I'm sort of torn on the tests. I think it's interesting to see how the compiler reacts to things, but I've had bad luck in the past asserting on literal string messages from things that can be localized (like exception messages or compiler warnings). If we keep them, the best we could do would likely be to assert the message contains `int?` and `notnull` or something - keywords we expect that aren't localizable. Otherwise tests fail when compiled by someone with a different culture setup.

We don't have documentation on why previous requirements exist; unless it's a really large project that started internal and went open source (like, say, Chromium) I've generally never seen such things. I've been working with Autofac since like 2009 and it was around long before me, so... yeah. The test is the spec.

If we have a test that shows we can register enums... we need to keep supporting that or it's a breaking change. I think it's probably reasonable to keep that functionality.

I got ahold of @alexmg and he agrees this is probably a good thing to pull into 5.0 so if we can get things working reasonably with tests passing let's go for it. I'm going to let you use your best judgment on whether it's "ready" - There's a lot here and, honestly, I'm not going to have enough time to literally pore over every line of changes and vet it all. In general it looks really good and I trust you.

Looks like there's a conflict to resolve, probably something to do with the package icon in the .csproj.

Once this is in, we probably need to do a pull into the immutable container branch to keep it up to date. Unclear if we'll actually get that back into develop at the moment; the perf hit is too big and we haven't had enough work done to fix it up yet. Better to ship without immutable containers than get further killed on perf. But that's another discussion entirely.

test/Autofac.Test.Compilation/NullableReferenceTests.cs
```diff
@@ -0,0 +1,102 @@
...
+            var containerBuilder = new ContainerBuilder();
+
+            new AutofacCompile()
+                .Body(
+                @"
+                   var containerBuilder = new ContainerBuilder();
+                
+                   containerBuilder.Register(c => new DerivedClass()).As<BaseClass?>();
+                ")
+                .AssertWarningEndsWith("Nullability of type argument 'Autofac.Test.Compilation.BaseClass?' doesn't match 'notnull' constraint.");
```
tillig: If we keep these tests and assert on the warning, we should only assert on things that aren't localizable like the names of types and constraints. Otherwise the tests will fail on a machine that has a different culture setting.

I'm OK if we don't keep these, or if we just keep it limited to one or two like this so things can be generally tested out. I can see a lot of these might become painful to maintain.

alistairjevans: I've amended the assertions to look for keywords only, that is a good tip (I did consider trying to force the compiler to invariant culture, but decided that really wasn't a great solution at all).

alistairjevans:
Git is so infuriating sometimes.

I've done the rebase onto latest develop to resolve the conflicts.

The compilation tests are something I see as testing the 'broad strokes' of the solution, and to verify that a change in compiler version doesn't change the behaviour we expect to see. I won't be adding tests for every method that has nullable behaviour, just the different categories of method.

tillig:
I can't tell you how many times I've been, like, forget this, I'll just manually merge everything. Thanks for nothing, Git! ;)

I think the tests look great now. You ready for it to be merged?

alistairjevans:
Something additional relating to clear support for value type services, that is a clear difference from before; if you use generic TryResolve or ResolveOptional methods (say for an int), and they fail to resolve, the return value will be ``Nullable<int>``, rather than an int with the default value, as it is now.

I think it is probably more appropriate anyway for these methods to return explicit null, but it is a breaking change.

As an example (and to round out how value type resolution works):

```csharp
containerBuilder.Register<MyEnum>(); // compiles
containerBuilder.Register<MyEnum?>(); // error

scope.ResolveOptional<MyEnum?>; // error
scope.ResolveOptional<MyEnum>; // return type is Nullable<MyEnum>

scope.TryResolve<MyEnum>(out var enumVal); // enumVal is Nullable<MyEnum>
```

Just rounding out some last nullability hints and we should be good to go.

alistairjevans:
One second, just want to check a notnull constraint on a Resolve method...

tillig:
No hurry, I'm ready when you are. If it's not today, no worries. Take all the time you need.

alistairjevans:
All good, just wanted to make sure all the Resolve methods were consistent with each other.

I think we might be ready to ``nullify`` this thing (I feel it needs the code quotes...)

alexmg:
Great work @alistairjevans!

Issue #963, opened 2019-03-07
Partially decorating service implementing multiple interfaces throws


blalonde:
Autofac 4.9.1

When a service implements 2 interfaces and only one is decorated, it fails.

```
public interface IDecorable
{
	
}

public interface IService
{

}

public class Service : IDecorable, IService
{

}

public class Decorator : IDecorable
{
	public Decorator(IDecorable service)
	{
	}
}

[Fact]
public void Test()
{
	var builder = new ContainerBuilder();
	builder.RegisterType<Service>().As<IDecorable>().As<IService>();
	builder.RegisterDecorator<Decorator, IDecorable>();

	var container = builder.Build();
	var decoratedService = container.Resolve<IDecorable>();
	var nondecoratedService = container.Resolve<IService>(); // fails Unable to cast object of type 'Decorator' to type 'IService'.
}
```
Failure : throws exception in comment

I assumed getting IDecorable would retrieve Service decorated by Decorator and retrieving IService would get same instance of Service not decorated

tillig:
When you say "fails," is it an exception or is it failing to decorate the interface? What do you see vs. what do you expect?

tillig:
Hmmm, this is an interesting case. I think your expectation sounds right, that you should be able to decorate a particular interface but not have the system assume every interface on the registration is decorated.

As a quick workaround, you could probably separate the registration:

```c#
builder.RegisterType<Service>().As<IDecorable>();
builder.RegisterType<Service>().As<IService>();
builder.RegisterDecorator<Decorator, IDecorable>();
```

But that doesn't solve the problem, it just works around it. We'll have to see what we can do about this.

jonathansapp:
I'm also having an issue that I believe is related.

I have a class that closes an interface implementation twice. When wrapped with a generic decorator, only one of the two closed interfaces can be resolved:

```csharp
public class UnitTest1
{
    [Fact]
    public void CanResolveUseCaseImplementations()
    {
        // Arrange
        var builder = new ContainerBuilder();
        builder.RegisterType<MyServiceImpl>().AsImplementedInterfaces();
        var container = builder.Build();

        // Act
        container.Resolve<IUseCase<MyServiceImpl.ConfirmRequest>>();
        container.Resolve<IUseCase<MyServiceImpl.Request>>();
    }

    [Fact]
    public void CanResolveFirstDecoratedService()
    {
        // Arrange
        var builder = new ContainerBuilder();
[...]

PR #1072, opened 2020-01-26
Fix Issue#963: Partially decorating service implementing multiple interfaces throws

RaymondHuy:
Fix issue #963: Add target component to be decorated to the decorating process so that we can get the InstanceSharing.


src/Autofac/Core/Lifetime/LifetimeScope.cs
```diff
@@ -316,6 +316,41 @@ public object GetOrCreateAndShare(Guid id, Func<object> creator)
             return result;
         }
 
+        /// <summary>
+        /// Try to create an instance with a GUID key.
+        /// </summary>
+        /// <param name="id">Key.</param>
+        /// <param name="creator">Creation function.</param>
+        /// <returns>An instance.</returns>
+        public object Create(Guid id, Func<object> creator)
```
tillig: It appears the `GetOrCreateAndShare` method was broken up into two pieces. Is there anything else calling the single unified version of the method? If not, and if the interface is already breaking, perhaps removing the old one is interesting so we're not maintaining it.

src/Autofac/Core/Resolving/InstanceLookup.cs
```diff
@@ -81,19 +83,43 @@ public object Execute()
             _executed = true;
 
             object? decoratorTarget = null;
-            object instance = ComponentRegistration.Sharing == InstanceSharing.None
-                ? Activate(Parameters, out decoratorTarget)
-                : _activationScope.GetOrCreateAndShare(ComponentRegistration.Id, () => Activate(Parameters, out decoratorTarget));
+
```
tillig: I'll need someone else to take a look at this. I haven't spent much time in the decorator stuff and could miss something important. Ideally @alexmg but @alistairjevans or @alsami may have some insight as well?

src/Autofac/Core/Resolving/InstanceLookup.cs
```diff
@@ -81,19 +83,43 @@ public object Execute()
...
             handler?.Invoke(this, new InstanceLookupEndingEventArgs(this, NewInstanceActivated));
 
             StartStartableComponent(decoratorTarget);
 
-            return instance;
+            return _newInstance;
         }
 
-        private void StartStartableComponent(object? instance)
+        private void StartStartableComponent(object instance)
```
tillig: I think the `object?` was part of the nullable object support recently added. The warnings only show up in a few cases, though I don't remember what those cases are. If this thing can take in a null, I think the `object?` needs to stay in place.

test/Autofac.Test/Features/Decorators/DecoratorTests.cs
```diff
@@ -128,10 +128,10 @@ public void DecoratedInstancePerLifetimeScopeRegistrationCanIncludeOtherServices
             var decoratedServiceInstance = container.Resolve<IDecoratedService>();
             Assert.IsType<DecoratorA>(decoratedServiceInstance);
 
-            Assert.Same(serviceInstance, decoratedServiceInstance);
+            Assert.Same(serviceInstance, decoratedServiceInstance.Decorated);
```
tillig: Man, I did a horrible job at getting these tests in here. :( Thanks for fixing them.

src/Autofac/Core/Resolving/InstanceLookup.cs
alistairjevans: In the original code, Roslyn couldn't work out something always assigned to decoratorTarget in the Execute method, so the nullable marker was necessary. Looks like Roslyn can figure it out with the new method, so there's no need for the nullable marker in StartStartableComponent.

src/Autofac/Core/ISharingLifetimeScope.cs
```diff
@@ -50,5 +50,20 @@ public interface ISharingLifetimeScope : ILifetimeScope
         /// <param name="creator">Creation function.</param>
         /// <returns>An instance.</returns>
         object GetOrCreateAndShare(Guid id, Func<object> creator);
+
+        /// <summary>
+        /// Try to retrieve an instance based on a GUID key.
+        /// </summary>
+        /// <param name="id">Key to look up.</param>
+        /// <returns>An instance.</returns>
+        object Get(Guid id);
```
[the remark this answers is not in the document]
RaymondHuy: Hi @tillig I've just added 2 methods in this interface. I didn't modify the logic or remove `GetOrCreateAndShare` method so I wonder why it's considered as a breaking change :)

src/Autofac/Core/Lifetime/LifetimeScope.cs
RaymondHuy: Hi @tillig  I have just reviewed the code, the `GetOrCreateAndShare` method is not used anymore but maybe someone using Autofac library is calling it. In my opinion the `Get` method must be added (it does only "get" responsibility so I don't think it's a breaking change or has side-effect to anything else), the `Create` method maybe replaced by re-using `GetOrCreateAndShare` method if you want ;)

src/Autofac/Core/Resolving/InstanceLookup.cs
alsami: @tillig I haven't worked with that code so far but I could dig into it if no one has time that also already has a better understanding of it.

src/Autofac/Core/ISharingLifetimeScope.cs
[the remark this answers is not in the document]
tillig: If you add, remove, or change a method on an interface and someone has an implementation of that interface, their code will no longer compile - that's why it's breaking.

That said, I think it's pretty low risk that anyone outside of Autofac has implemented their own `ISharingLifetimeScope` so, while it's technically breaking, we could probably release it as a 0.1.x increment. It'd still be nice to see if it's possible to fix without changing the interface.

src/Autofac/Core/Resolving/InstanceLookup.cs
tillig: I'd welcome any available help on this. I know I won't have time to dive deeper.

alexmg: I'll take a look at this one.

alexmg: The fix looks good and does require the `GetOrCreateAndShare` method to be split up. I don't think there is any point leaving the old method behind and unused either. There is no mechanism in place for you to swap out the `ISharingLifetimeScope` implementation and internally we create instances of `LifetimeScope` and pass them through internal classes. I don't see a reason why anyone would be implementing this interface given the constraints.

This is definitely a breaking change but I don't think it is one that will have much of an impact due to it not being an extensibility point, but these things can certainly play out in unexpected ways. I would be fine with trying to get away with a minor release for this one.

A few suggestions:

- Remove the `GetOrCreateAndShare` method from `ISharingLifetimeScope` and implementation in `LifetimeScope` (not used and we are breaking things anyway)
- Rename `Create` method to `CreateSharedInstance`
- Rename `Get` to `TryGetSharedInstance` with an out parameter and return a `bool` instead of `null` when not found

It would be good to consider how we can actually make these methods internal and remove them from the public API but that's probably a job for a separate issue.

RaymondHuy: Thank for your suggestions @alexmg I have updated the code. :)

alexmg:
Thanks for the help on this one @RaymondHuy! 👍

PR #1076, opened 2020-01-30
Fix Issue #1073: Synchronize read access

MaratFaskhiev:
Fix issue #1073: Synchronize read access. We should disallow to read during write. Otherwise it is possible to read inconsistent state.

### DIFF
--- a/test/Autofac.Test/Concurrency/ConcurrencyTests.cs
+++ b/test/Autofac.Test/Concurrency/ConcurrencyTests.cs
@@ -134,7 +134,7 @@ namespace Autofac.Test.Concurrency
         }
 
         [Fact]
-        public void WhenTwoThreadsResolvesNotAlreadyRegisteredType_DoesNotThrow()
+        public void WhenSeveralThreadsResolveNotAlreadyRegisteredType_DoesNotThrow()
         {
             for (int i = 0; i < 10000; i++)
             {
@@ -142,6 +142,12 @@ namespace Autofac.Test.Concurrency
                 builder.RegisterSource(new AnyConcreteTypeNotAlreadyRegisteredSource());
                 var container = builder.Build();
                 Parallel.Invoke(
+                    () => container.Resolve<A>(),
+                    () => container.Resolve<A>(),
+                    () => container.Resolve<A>(),
+                    () => container.Resolve<A>(),
+                    () => container.Resolve<A>(),
+                    () => container.Resolve<A>(),
                     () => container.Resolve<A>(),
                     () => container.Resolve<A>());
             }
--- a/test/Autofac.Test/Concurrency/ConcurrencyTests.cs
+++ b/test/Autofac.Test/Concurrency/ConcurrencyTests.cs
@@ -3,6 +3,7 @@ using System.Collections.Concurrent;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using Autofac.Features.ResolveAnything;
 using Xunit;
 
 namespace Autofac.Test.Concurrency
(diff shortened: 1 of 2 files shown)


tillig:
Thanks for the PR! We'll need to take a little time to analyze the perf changes you mentioned, but if it's reasonably minimal it'll be good to be more thread safe.

test/Autofac.Test/Concurrency/ConcurrencyTests.cs
```diff
@@ -132,6 +133,20 @@ public void WhenTwoThreadsInitialiseASharedInstanceSimultaneouslyViaChildLifetim
...
 
+        [Fact]
+        public void WhenTwoThreadsResolvesNotAlreadyRegisteredType_DoesNotThrow()
+        {
+            for (int i = 0; i < 10000; i++)
+            {
+                var builder = new ContainerBuilder();
+                builder.RegisterSource(new AnyConcreteTypeNotAlreadyRegisteredSource());
+                var container = builder.Build();
+                Parallel.Invoke(
```
botinko: maybe it's better to ad more `() => container.Resolve<A>()`  to fail with higher probability?

MaratFaskhiev: done

alexmg:
Thanks for the PR @MaratFaskhiev!

It definitely looks like the issue is with the ACTNAR source not being available even though the `ServiceRegistrationInfo` is considered initialized.

Unfortunately, it looks like the `ReaderWriterLockSlim` impacts concurrency considerably more than expected.

I have run some benchmarks to see compare the original, current and proposed locking implementations.

Current implementation:

```
|                          Method | ResolveTaskCount | ResolvesPerTask |     Mean |    Error |   StdDev |    Gen 0 |   Gen 1 | Gen 2 | Allocated |
|-------------------------------- |----------------- |---------------- |---------:|---------:|---------:|---------:|--------:|------:|----------:|
| MultipleResolvesOnMultipleTasks |              100 |             100 | 536.2 us | 10.19 us | 10.91 us | 635.7422 | 35.1563 |     - |   5.05 MB |
```

Proposed `ReaderWriterLockSlim` implementation:

```
|                          Method | ResolveTaskCount | ResolvesPerTask |     Mean |     Error |    StdDev |    Gen 0 |   Gen 1 | Gen 2 | Allocated |
|-------------------------------- |----------------- |---------------- |---------:|----------:|----------:|---------:|--------:|------:|----------:|
| MultipleResolvesOnMultipleTasks |              100 |             100 | 8.257 ms | 0.1570 ms | 0.3099 ms | 625.0000 | 15.6250 |     - |   5.05 MB |
```

Original coarse grain locking implementation:

```
|                          Method | ResolveTaskCount | ResolvesPerTask |     Mean |     Error |    StdDev |    Gen 0 |   Gen 1 | Gen 2 | Allocated |
|-------------------------------- |----------------- |---------------- |---------:|----------:|----------:|---------:|--------:|------:|----------:|
| MultipleResolvesOnMultipleTasks |              100 |             100 | 1.862 ms | 0.0195 ms | 0.0183 ms | 636.7188 | 27.3438 |     - |   5.05 MB |
```

Another way to fix this is to consider the `ServiceRegistrationInfo` class to be invalid until it is both initialized and contains implementations. If a thread catches the instance at a point where it is initialized but does not yet have any implementations, it will be forced to move forward into the lock. If there really wasn't any implementations this will be confirmed once the instance has been retrieved from within the lock.

This is a rather targeted solution but will resolve the issue while keeping the concurrency close to the current implementation. I think the concurrency benefit makes taking this approach worthwhile.

Additional check for implementations outside lock:

```
|                          Method | ResolveTaskCount | ResolvesPerTask |     Mean |    Error |   StdDev |    Gen 0 |   Gen 1 | Gen 2 | Allocated |
|-------------------------------- |----------------- |---------------- |---------:|---------:|---------:|---------:|--------:|------:|----------:|
| MultipleResolvesOnMultipleTasks |              100 |             100 | 640.4 us | 12.26 us | 15.05 us | 730.4688 | 38.0859 |     - |   5.81 MB |
```

I have pushed a commit with a fix for this.

alexmg:
Fixed with commit https://github.com/autofac/Autofac/commit/cf3674a77cfd1de04ed198fd3760ae4d8b13a571

MaratFaskhiev:
@alexmg Thank you.

alexmg:
This fixed was later updated in f55da6287966167fd11cb2969ff3521af216428a to use the original coarse grain locking due to issue #1073.

PR #1090, opened 2020-03-07
Fix #1089: Owned<T> asynchronous disposal inconsistency with ILifetim…

### LINKED ISSUE #1089: Owned<T> asynchronous disposal inconsistency with ILifetimeScope
It seems that if you call DisposeAsync on instance of Owned<T> all the dependant instances under that scope still get disposed with Dispose method instead of DisposeAsync when it's available.

rafalkwol:
I think this should fix #1089 issue - I'm not sure if I did all the fork, branch and pull request process correctly as it's my fist time. I think I should also link this to the exisiting issue, but I'm not sure how to do it as I can't find any option for it.

### DIFF
--- a/src/Autofac/Features/OwnedInstances/Owned.cs
+++ b/src/Autofac/Features/OwnedInstances/Owned.cs
@@ -26,6 +26,7 @@
 using System;
 using System.Diagnostics.CodeAnalysis;
 using System.Threading;
+using System.Threading.Tasks;
 using Autofac.Core;
 using Autofac.Util;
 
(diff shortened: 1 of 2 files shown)


alistairjevans:
@rafalkwol, this looks good to me; thanks so much for the contribution!

You can link a pull request by putting some text in your description like:
```
Fixes #1089
```

The word 'Fixes' is one of the [special words GitHub looks for](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword). You can also put this in when editing the description and it will pick it up.

Minor annoyance about ``Owned<T>`` (nothing to do with your changes); the constructor for ``Owned<T>`` takes an ``IDisposable`` as the lifetime object, even though in Autofac code it's always a ``ILifetimeScope``; hence the need to pattern match for ``IAsyncDisposable`` in your new ``DisposeAsync`` override. Perhaps there's a reason Owned takes an IDisposable rather than ILifetimeScope, but it would be a breaking change now to change it.

@tillig / @alexmg, this looks ready to merge from my perspective.

rafalkwol:
@alistairjevans Thanks for your input and help. I changed my description a bit to do the linking trick you mentioned. About the IDisposable I also noticed it, but indeed it seems like a change that may have broke something somewhere else if I changed it so I decided to go for the pattern matching. I'm also not sure why IDisposable is in the constructor, but I guess it may be a relic from a distant past so it's better not to change it.

PR #1091, opened 2020-03-08
Enhance autofac internal to fix issue#27 in DynamicProxy

RaymondHuy:
Reference: https://github.com/autofac/Autofac.Extras.DynamicProxy/issues/27
Add Service info to events raised from `IComponentRegistration`

### DIFF
--- a/src/Autofac/Core/IComponentRegistration.cs
+++ b/src/Autofac/Core/IComponentRegistration.cs
@@ -91,11 +91,11 @@ namespace Autofac.Core
         /// Called by the container when an instance is required.
         /// </summary>
         /// <param name="context">The context in which the instance will be activated.</param>
-        /// <param name="parameters">Parameters for activation. These may be modified by the event handler.</param>
         /// <param name="service">The service being resolved.</param>
+        /// <param name="parameters">Parameters for activation. These may be modified by the event handler.</param>
         [SuppressMessage("Microsoft.Design", "CA1045:DoNotPassTypesByReference", MessageId = "1#", Justification = "The method may change the backing store of the parameter collection.")]
         [SuppressMessage("Microsoft.Design", "CA1030:UseEventsWhereAppropriate", Justification = "This is the method that would raise the event.")]
-        void RaisePreparing(IComponentContext context, ref IEnumerable<Parameter> parameters, Service service);
+        void RaisePreparing(IComponentContext context, Service service, ref IEnumerable<Parameter> parameters);
 
         /// <summary>
         /// Fired when a new instance is being activated. The instance can be
@@ -109,12 +109,12 @@ namespace Autofac.Core
         /// </summary>
         /// <param name="context">The context in which the instance was activated.</param>
         /// <param name="parameters">The parameters supplied to the activator.</param>
-        /// <param name="instance">The instance.</param>
         /// <param name="service">The service being resolved.</param>
+        /// <param name="instance">The instance.</param>
         [SuppressMessage("Microsoft.Design", "CA1007:UseGenericsWhereAppropriate")]
         [SuppressMessage("Microsoft.Design", "CA1045:DoNotPassTypesByReference", MessageId = "2#", Justification = "The method may change the object as part of activation.")]
         [SuppressMessage("Microsoft.Design", "CA1030:UseEventsWhereAppropriate", Justification = "This is the method that would raise the event.")]
-        void RaiseActivating(IComponentContext context, IEnumerable<Parameter> parameters, ref object instance, Service service);
+        void RaiseActivating(IComponentContext context, IEnumerable<Parameter> parameters, Service service, ref object instance);
 
         /// <summary>
         /// Fired when the activation process for a new instance is complete.
@@ -127,9 +127,9 @@ namespace Autofac.Core
         /// </summary>
         /// <param name="context">The context in which the instance was activated.</param>
         /// <param name="parameters">The parameters supplied to the activator.</param>
-        /// <param name="instance">The instance.</param>
         /// <param name="service">The service being resolved.</param>
+        /// <param name="instance">The instance.</param>
         [SuppressMessage("Microsoft.Design", "CA1030:UseEventsWhereAppropriate", Justification = "This is the method that would raise the event.")]
-        void RaiseActivated(IComponentContext context, IEnumerable<Parameter> parameters, object instance, Service service);
+        void RaiseActivated(IComponentContext context, IEnumerable<Parameter> parameters, Service service, object instance);
     }
 }
--- a/src/Autofac/Core/IComponentRegistration.cs
+++ b/src/Autofac/Core/IComponentRegistration.cs
@@ -1,4 +1,4 @@
-// This software is part of the Autofac IoC container
+﻿// This software is part of the Autofac IoC container
 // Copyright © 2011 Autofac Contributors
 // https://autofac.org
 //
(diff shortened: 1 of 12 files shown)


alistairjevans:
This was discussed on the original issue, https://github.com/autofac/Autofac.Extras.DynamicProxy/issues/27, but just to note that these are breaking changes due to the modifications to ``IActivatedEventArgs``, ``IActivatingEventArgs`` and the constructors on the applicable implementations, all of which are public.

So this looks like a 6.0 change (that may not be released for a while, since 5.0 only just went out), unless we can think of a way to make the changes non-breaking. I thought about a constructor overload, but it's the interface that is difficult to get around.

Might be worth bringing this into a develop-60 branch for breaking changes such as this one.

alsami:
We could also make that parameter for the service take null by default and do not check for null rather set it to some constant which states that it is being unknown.

```csharp
public ActivatedEventArgs(
            IComponentContext context,
            IComponentRegistration component,
            IEnumerable<Parameter> parameters,
            T instance,
            Service service = null)
{
  // ...
}
```

This wouldn't make it a breaking-change but has the downside that you always have to check for null-values.

alistairjevans:
@alsami, adding optional parameters change the signature of the method, so it is technically still a breaking change (based on my understanding of 'non-breaking', in that the library can be patched without rebuilding any code).

The bigger problem though is the interfaces, which must be implemented by anyone who has defined their own (for testing, perhaps). 

Perhaps we could temporarily sub-class the ``ActivatingEventArgs`` with a new ``ServiceAwareActivatingEventArgs``, and raise that one in 5.0, then in 6 we could lift the Service Property up? Until 6.0, extensions would need to check ``if (args is ServiceAwareActivatingEventArgs)``.

alsami:
> adding optional parameters change the signature of the method, so it is technically still a breaking change (based on my understanding of 'non-breaking', in that the library can be patched without rebuilding any code).

It does change the signature but it does not introduce a breaking-change, when the new parameter is being added in the and and is being marked as optional. Whenever you upgrade a library you projects will rebuild so I am not sure if that should be considered a breaking-change.

Removing something from a signature or changing an existing type of a signature would introduce a breaking-change. When we add this to let's say 5.2 to and people update their libraries they won't need to apply any changes in their code.

> The bigger problem though is the interfaces, which must be implemented by anyone who has defined their own (for testing, perhaps).

That is indeed a problem though. In this case, as far as I saw, it would only be adding some properties. Maybe we could leverage the default-interface implementation of C# 8 here? Something like this
`ServiceType serviceType { get; } = default`

> Perhaps we could temporarily sub-class the `ActivatingEventArgs` with a new `ServiceAwareActivatingEventArgs`, and raise that one in 5.0, then in 6 we could lift the Service Property up? Until 6.0, extensions would need to check `if (args is ServiceAwareActivatingEventArgs)`.

That is a good option as well. Might be a little more work than the other way. Anyway I don't mind which one it is, just seeing that it's possible to do

alistairjevans:
> It does change the signature but it does not introduce a breaking-change, when the new parameter is being added in the end and is being marked as optional. Whenever you upgrade a library your projects will need rebuilding so I am not sure if that should be considered a breaking-change.
> 
> Removing something from a signature or changing an existing type of a signature would introduce a breaking-change. When we add this to let's say 5.2 to and people update their libraries they won't need to apply any changes in their code.

That is only true for a direct dependency on Autofac. If there is another package (say, MyAutofacExt) installed that depends on Autofac, and the user updates Autofac themselves, MyAutofacExt may no longer be compatible, even though it has a constraint of ``[5.0.0,6.0.0)``, intended to prevent breaking changes from causing issues in MyAutofacExt.

> Maybe we could leverage the default-interface implementation of C# 8 here? Something like this
``ServiceType serviceType { get; } = default``

I mean, this is pretty much exactly what default interface implementations are for. My only concern is that it can constrain consumers of the library as to what version of the compiler they can use when **consuming** Autofac? Any target framework restrictions? It feels like a feature the runtime would need to be involved in, but am struggling to find explicit supported framework info for it.

alsami:
> MyAutofacExt may no longer be compatible

Are you speaking about cases where you would get runtime exceptions like `MethodNotFoundException` because of the signature?
Yeah, that could happen, but only if you reference a higher version of `Autofac` in the project that consumes `MyAutofacExt`.

> I mean, this is pretty much exactly what default interface implementations are for. My only concern is that it can constrain consumers of the library as to what version of the compiler they can use when **consuming** Autofac? Any target framework restrictions? It feels like a feature the runtime would need to be involved in, but am struggling to find explicit supported framework info for it.

Sure, it might not be available for everyone. Not sure how it behaves on a .NET-Framework project.

Anyway, not trying to argue for or against any of it 👍

alistairjevans:
Yeah, ``MethodNotFoundException``.  Not arguing either way here either, just musing on breaking change impacts. Equally, we could add a constructor overload instead of an optional value, that would stop any constructor breakages.

I feel like the subclass approach _might_ be more straightforward/less-breaking, but can't be sure. @RaymondHuy, do you have any thoughts? Is this something you think can wait for a 6.0?

tillig:
Adding an optional parameter is a breaking change. I have, unfortunately, learned that the hard way having to roll back a change like that. :(

It's okay to add a second method overload, though. Leave the first in place and have that act like the default version of the method; add the second overload and don't make the parameter optional. Less clean, but always works.

RaymondHuy:
Hi @alistairjevans , this PR is only used for that issue so I think we can wait until 6.0

src/Autofac/Core/IComponentRegistration.cs
```diff
@@ -109,10 +110,11 @@ public interface IComponentRegistration : IDisposable
         /// <param name="context">The context in which the instance was activated.</param>
         /// <param name="parameters">The parameters supplied to the activator.</param>
         /// <param name="instance">The instance.</param>
+        /// <param name="service">The service being resolved.</param>
         [SuppressMessage("Microsoft.Design", "CA1007:UseGenericsWhereAppropriate")]
         [SuppressMessage("Microsoft.Design", "CA1045:DoNotPassTypesByReference", MessageId = "2#", Justification = "The method may change the object as part of activation.")]
         [SuppressMessage("Microsoft.Design", "CA1030:UseEventsWhereAppropriate", Justification = "This is the method that would raise the event.")]
-        void RaiseActivating(IComponentContext context, IEnumerable<Parameter> parameters, ref object instance);
+        void RaiseActivating(IComponentContext context, IEnumerable<Parameter> parameters, ref object instance, Service service);
```
alistairjevans: Same as with RaisePreparing, the ref parameter should probably be at the end.

src/Autofac/Core/IComponentRegistration.cs
```diff
@@ -92,9 +92,10 @@ public interface IComponentRegistration : IDisposable
         /// </summary>
         /// <param name="context">The context in which the instance will be activated.</param>
         /// <param name="parameters">Parameters for activation. These may be modified by the event handler.</param>
+        /// <param name="service">The service being resolved.</param>
         [SuppressMessage("Microsoft.Design", "CA1045:DoNotPassTypesByReference", MessageId = "1#", Justification = "The method may change the backing store of the parameter collection.")]
         [SuppressMessage("Microsoft.Design", "CA1030:UseEventsWhereAppropriate", Justification = "This is the method that would raise the event.")]
-        void RaisePreparing(IComponentContext context, ref IEnumerable<Parameter> parameters);
+        void RaisePreparing(IComponentContext context, ref IEnumerable<Parameter> parameters, Service service);
```
alistairjevans: Since we're making breaking changes, can I suggest that we move the ``service`` argument to before ``parameters``? Normally ref and out parameters would be last in the parameter list.

src/Autofac/Core/IComponentRegistration.cs
```diff
@@ -126,7 +128,8 @@ public interface IComponentRegistration : IDisposable
         /// <param name="context">The context in which the instance was activated.</param>
         /// <param name="parameters">The parameters supplied to the activator.</param>
         /// <param name="instance">The instance.</param>
+        /// <param name="service">The service being resolved.</param>
         [SuppressMessage("Microsoft.Design", "CA1030:UseEventsWhereAppropriate", Justification = "This is the method that would raise the event.")]
-        void RaiseActivated(IComponentContext context, IEnumerable<Parameter> parameters, object instance);
+        void RaiseActivated(IComponentContext context, IEnumerable<Parameter> parameters, object instance, Service service);
```
alistairjevans: If we change RaiseActivating order, RaiseActivated should match.

alistairjevans:
Once the small additional changes are done, unless anyone disagrees, I'll create a v6.0 branch and merge this into there.

tillig:
I think a v6 branch is fine, but it may require ongoing maintenance. Not that leaving a PR out there is much better, of course.

src/Autofac/Core/IComponentRegistration.cs
RaymondHuy: Thanks for your suggestion. I have updated it. ;)

Issue #925, opened 2018-07-06
ACTNARS fails to resolve closed generics with abstract type argument


Codelet:
I still have problem with #855 in cases of `Foo<Bar>` where `Bar` is abstract class.
simply doing
```
public class Foo<T> { }
public abstract class Bar {}
```
and trying to resolve `Foo<Bar>` with ACTNARS gives me exception that type is not registered

tillig:
Do you have a small repro we can try in unit test form? Which version of Autofac are you using?

Codelet:
Sure, here is the demo code
https://github.com/Codelet/AutofacACTNARS

tillig:
For reference, here's the repro code from the repo:

```csharp
namespace AutofacACTNARS
{
  using Autofac;
  using Autofac.Features.ResolveAnything;

  public class Foo<T>
  {
  }

  public class Bar1
  {
  }

  public abstract class Bar2
  {
  }

  public class Program
  {
    public static void Main(string[] args)
    {
      var builder = new ContainerBuilder();

      builder.RegisterSource(new AnyConcreteTypeNotAlreadyRegisteredSource());

      var container = builder.Build();

      var thisWorks = container.Resolve<Foo<Bar1>>(); // OK, Bar1 is not abstract
      var thisDoesnt = container.Resolve<Foo<Bar2>>(); // Throws, Bar2 is abstract, but there is no code that would use it
    }
  }
}
```

Based on that, it appears the challenge is around _abstract_ types.

On a side note, I think you'd have to define the interface as `Foo<out T>` to allow it to be used. I doubt that'd make a difference here, but it's something to be aware of.

tillig:
It looks like it comes down to this in ACTNARS:

```csharp
            if (typeInfo.IsGenericType)
            {
                foreach (var typeParameter in typeInfo.GenericTypeArguments)
                {
                    // Issue #855: If the generic type argument doesn't match the filter don't look for a registration.
                    if (_predicate(typeParameter) && !registrationAccessor(new TypedService(typeParameter)).Any())
                    {
                        return Enumerable.Empty<IComponentRegistration>();
                    }
                }
            }
```

Basically, "If it hasn't been filtered out and we can find a registration for the type, let it through."

But ACTNARS by default doesn't provide registrations for abstract types:

```csharp
            if (!typeInfo.IsClass ||
                typeInfo.IsSubclassOf(typeof(Delegate)) ||
                typeInfo.IsAbstract ||
[...]

Issue #855, opened 2017-06-19
ACTNARS fails to resolve closed generics


droyad:
We are using `AnyConcreteTypeNotAlreadyRegisteredSource`, and it's behaviour has changed in 4.2.0 due to [this](https://github.com/autofac/Autofac/commit/1aca241fc0e7d5535fdff23c833dcbfdaa0f02b7) change stemming from #495.

For example we want to resolve `Foo<Bar>`  where we have an implementation `Foo<T>`. The `ACTNARS` delegate is `t => t.Name.StartsWith("Foo")`. `Foo<T>` does not take `Bar` in the constructor.

`ACTNARS` currently (`4.6.0`) fails to resolve this type as it checks that `Bar` is registered ([here](https://github.com/autofac/Autofac/blob/v4.6.0/src/Autofac/Features/ResolveAnything/AnyConcreteTypeNotAlreadyRegisteredSource.cs#L98)

```
     [Fact]
        public void ConstructableOpenGenericsCanBeResolved()
        {
            var cb = new ContainerBuilder();
            cb.RegisterSource(new AnyConcreteTypeNotAlreadyRegisteredSource(t => t.Name.StartsWith("Progress")));
            var container = cb.Build();
            Assert.True(container.IsRegistered<Progress<Exception>>());
            Assert.NotNull(container.Resolve<Progress<Exception>>());
        }
```

tillig:
Good find. We'll have to think about how best to solve this. `AnyConcreteTypeNotRegisteredSource` and `ContravariantRegistrationSource` are two great, flexible, pains in my behind. In #495 we found ACTNARS interfering with other registration sources but we really can't put special cases around `Meta<T>` or other source "known types." Maybe I fixed that wrong, or maybe there's more to check? Hmmm. Will have to dig in.

alexmg:
This fix is included in 4.6.1 which has been pushed to NuGet.

PR #1099, opened 2020-03-31
Resolve Issue #925

robjmenna:
I made an update to try and allow ACTNARS to resolve generic types with abstract type arguments. Essentially, it just checks the constructor of the concrete type to see if the concrete type is dependent on the abstract type arguments. If not, then it provides the registration. I did have to add a special case to make sure it didn't return any Lazy<> registrations that might override the default Lazy<> behavior.

### DIFF
--- a/src/Autofac/Features/ResolveAnything/AnyConcreteTypeNotAlreadyRegisteredSource.cs
+++ b/src/Autofac/Features/ResolveAnything/AnyConcreteTypeNotAlreadyRegisteredSource.cs
@@ -58,6 +58,39 @@ namespace Autofac.Features.ResolveAnything
         }
 
         /// <summary>
+        /// Takes a generic type in <paramref name="typeInfo"/> and checks to see if it has any abstract type parameters that have not been registered.
+        /// If so, the method checks to see if the generic type has any constructors that can be used without depending on the abstract types.
+        /// </summary>
+        /// <param name="typeInfo">The <see cref="System.Reflection.TypeInfo"/> of the generic type to inspect.</param>
+        /// <param name="registrationAccessor">A function that will return existing registrations for a service.</param>
+        /// <returns>Returns true if <paramref name="typeInfo"/> has a constructor that does not depend on a unregistered abstract type.</returns>
+        private bool IsGenericTypeAValidRegistration(TypeInfo typeInfo, Func<Service, IEnumerable<IComponentRegistration>> registrationAccessor)
+        {
+            var typeParameters = typeInfo.GenericTypeArguments
+                    .Where(t => t.IsAbstract &&
+                        !registrationAccessor(new TypedService(t)).Any());
+
+            var isValidConstructorFound = false;
+            foreach (var constructor in typeInfo.GetConstructors())
+            {
+                var parameterTypes = constructor.GetParameters()
+                    .Select(p => p.ParameterType);
+
+                var isTypeParameterInConstructor = typeParameters
+                    .Where(t => parameterTypes.Contains(t))
+                    .Any();
+
+                if (!isTypeParameterInConstructor)
+                {
+                    isValidConstructorFound = true;
+                    break;
+                }
+            }
+
+            return isValidConstructorFound;
+        }
+
+        /// <summary>
         /// Retrieve registrations for an unregistered service, to be used
         /// by the container.
         /// </summary>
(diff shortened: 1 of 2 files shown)


src/Autofac/Features/ResolveAnything/AnyConcreteTypeNotAlreadyRegisteredSource.cs
```diff
@@ -91,16 +91,39 @@ public AnyConcreteTypeNotAlreadyRegisteredSource(Func<Type, bool> predicate)
...
             }
 
+            // Check to resolve issue #925
             if (typeInfo.IsGenericType)
             {
-                foreach (var typeParameter in typeInfo.GenericTypeArguments)
+                if (typeInfo.GetGenericTypeDefinition() == typeof(Lazy<>))
                 {
-                    // Issue #855: If the generic type argument doesn't match the filter don't look for a registration.
-                    if (_predicate(typeParameter) && !registrationAccessor(new TypedService(typeParameter)).Any())
```
tillig: It looks like the predicate check here got removed in the new code. I recognize that if the tests are passing that we potentially don't have tests around that, though the comment here indicates the predicate was added for a reason.

Do we no longer need that? Are we missing a test?

src/Autofac/Features/ResolveAnything/AnyConcreteTypeNotAlreadyRegisteredSource.cs
```diff
@@ -91,16 +91,39 @@ public AnyConcreteTypeNotAlreadyRegisteredSource(Func<Type, bool> predicate)
                 return Enumerable.Empty<IComponentRegistration>();
             }
 
+            // Check to resolve issue #925
```
tillig: Check what to resolve issue #925? Historically, we've actually moved from Google Code to GitHub and we've transferred issues around so just an issue number isn't enough context. Ideally comments here will explain not only the issue but _why_ we're doing what we're doing.

For example, the comment here probably doesn't make sense. This `if` check didn't change for #925.

However, just below where we exclude `Lazy<>` - why? A comment there explaining `Lazy<T>` has a default no-parameter constructor and potentially interferes with the `Lazy<T>` handling provided elsewhere... that's important. It also sort of explains why `Func<>` is _not_ excluded... but maybe it should be?

src/Autofac/Features/ResolveAnything/AnyConcreteTypeNotAlreadyRegisteredSource.cs
```diff
@@ -91,16 +91,39 @@ public AnyConcreteTypeNotAlreadyRegisteredSource(Func<Type, bool> predicate)
...
+                        !registrationAccessor(new TypedService(t)).Any());
+
+                bool isValidConstructorFound = false;
+                foreach (var constructor in typeInfo.GetConstructors())
+                {
+                    var parameterTypes = constructor.GetParameters()
+                        .Select(p => p.ParameterType);
+
+                    var isTypeParameterInConstructor = typeParameters
+                        .Where(t => parameterTypes.Contains(t))
```
tillig: When `parameterTypes.Contains` is called, does it re-run the `Select` above? (I have a feeling it might.) In which case, maybe there should be, say, a `ToArray()` above to force it to evaluate once and be done?

src/Autofac/Features/ResolveAnything/AnyConcreteTypeNotAlreadyRegisteredSource.cs
robjmenna: I broke out the main body of the code I added into its own private method so that I could document it more clearly. I added some single line comments that I hope are helpful also.

As for Func<>, I don't think I need to exclude it explicitly since ACTNARS already excludes all sub-types of Delegate.

src/Autofac/Features/ResolveAnything/AnyConcreteTypeNotAlreadyRegisteredSource.cs
robjmenna: Good point, I'll add `ToArray()` like you said.

alistairjevans: Minor perf point, ToList will be more efficient than ToArray in this case.

src/Autofac/Features/ResolveAnything/AnyConcreteTypeNotAlreadyRegisteredSource.cs
robjmenna: So I think the predicate was originally added to address [this issue](https://github.com/autofac/Autofac/issues/855). When I remove the predicate from the original code, `ConstructableOpenGenericsWithGenericTypeArgumentNotMatchingFilterCanBeResolved()` fails so it at least has one test. 

I don't think its needed anymore, but maybe I'm missing something. The test would originally fail because ACTNARS checked each type parameter for a registration when resolving a generic type. Since the `Exception` type did not satisfy the predicate, the whole `Progress<Exception>` type would fail to register without the predicate check. 

The new ACTNARS behavior passes the test because `Progress<Exception>` has a default constructor and it doesn't need a registration for `Exception`.

alistairjevans:
I had a more detailed look at this this morning, and I have a few thoughts. Let me know if I've missed something.

- By checking whether there is a valid constructor for a generic type before we provide a registration, we're duplicating behaviour that already exists elsewhere in Autofac, that throws the appropriate exception if such a constructor does not exist. This feels a bit odd. Why would we not just let ACTNARS say "yes, here's the registration you're looking for, let me know if you need any constructor parameters."?
Based on the work on the Moq Integration, which had to do a lot of ACTNARS-ish behaviour, I've created an alternate set of changes which simply excludes closed generic types based on whether they are one of the implicit relationship types.  Take a look here: https://github.com/autofac/Autofac/pull/1100.
It may be that I've missed something major, so do let me know.

- It strikes me as very odd that we would apply a predicate designed to filter the services that can be exposed by ACTNARS to also constrain the set of generic type parameters?

tillig:
I like the fairly minimal changes in #1100.

The predicate for filtering has been there for a while but was added to the generic type arguments as a fix for #855. I gather the intent is that if someone doesn't want type `Foo` auto-registered, then they also don't want `Generic<Foo>` auto-registered. However, I'm not sure that's totally sound logic and if the tests pass without the filter on the parameters... well, maybe we'll need someone to file an issue with a test that fails based on that missing predicate filter.

robjmenna:
I'll close this pull request then in favor of #1100.

As far as the predicate filter, I think the intention of the filter is actually the complete opposite. If the user doesn't want to register `Foo`, then the check makes sure that `Generic<Foo>` WILL still be registered. (Very confusing...)

For example, this test is already in ResolveAnythingTests.cs
```
[Fact]
public void ConstructableOpenGenericsWithGenericTypeArgumentNotMatchingFilterCanBeResolved()
{
    cb.RegisterSource(new AnyConcreteTypeNotAlreadyRegisteredSource(t => t.Name.StartsWith("Progress")));
    var container = cb.Build();
    Assert.True(container.IsRegistered<Progress<Exception>>());
    Assert.NotNull(container.Resolve<Progress<Exception>>());
}
```
Because the ACTNARS wants to see that every type parameter is registered, it would normally fail to register `Progress<Exception>` because `Exception` does not satisfy the predicate. By adding the predicate check, ACTNARS is basically saying "its impossible for this generic type parameter to be registered by me, so just return the closed generic type registration anyway".

PR #1100, opened 2020-04-01
Alternate ACTNARS generic approach

alistairjevans:
Alternate approach for https://github.com/autofac/Autofac/pull/1099.


test/Autofac.Test/Features/ResolveAnything/ResolveAnythingTests.cs
```diff
@@ -257,6 +256,16 @@ private static IContainer CreateResolveAnythingContainer()
             return cb.Build();
         }
 
+        public class GenericType<T>
```
tillig, line 259: Where is this getting used in the tests?

alistairjevans: It's not, my bad; left over from when I looked at bringing in the tests from the original PR. Removed.

alistairjevans:
Are we considering these breaking changes because of the predicate removal?

tillig:
No, the predicate is still running on the type before it gets to the generic part, so it still works. The tests are passing, too, so I'm unsure whether having it in the generic type parameter part ever really mattered. I feel like this is sort of 0.0.1 "bug fix" stuff right now.

alistairjevans:
Ok, cool, this is ready to go then. I'll just link the issue.

PR #1104, opened 2020-04-08
Fixes #1102 - AutoActivate Instance Registrations if they have Activation/ed handlers

### LINKED ISSUE #1102: OnRelease not triggered if Provided Instance service has never been resolved
@ale7canna and I noticed a weird behavior regarding the disposal of components registered with `RegisterInstance()` in combination with the use of `OnRelease()`.

We know from the documentation about [Provided Instances](https://autofaccn.readthedocs.io/en/latest/lifetime/disposal.html#provided-instances) that

> If you provide an instance registration to Autofac, Autofac will assume ownership of that instance and will handle its disposal.

This is confirmed by the next test, which shows how Autofac automatically disposes of the `MyClass` instance during the container disposal:

```csharp
internal class MyClass : IDisposable
{
    public static Boolean Disposed = false;

    public void Dispose() => Disposed = true;
}

[Fact]
void autofac_takes_care_of_the_disposal_of_instances_registered_with_RegisterInstance()
{
    builder.RegisterInstance(new MyClass());

    using (var container = builder.Build())
    {
        using (var scope = container.BeginLifetimeScope())
        {
            var service = scope.Resolve<MyClass>();
        }
    }

    MyClass.Disposed.Should().Be(true);
}
```

This works even if `MyClass` is actually never resolved, that is if the line

```csharp
var service = scope.Resolve<MyClass>();
```
is commented out.

So far so good. 
We know from [OnRelease](https://autofaccn.readthedocs.io/en/latest/lifetime/events.html#onrelease) that:

> OnRelease event replaces the standard cleanup behavior for a component.

So, if we register `MyClass` instance providing the registration with an `OnRelease` callback, Autofac will stop invoking `Dispose` on the Container disposal. 
In the next test, we use `OnRelease` to manually invoke `Dispose` when `OnRelease` is triggered.

```csharp
[Fact]
void autofac_does_does_not_invoke_OnRelease_if_service_is_never_resolved()
{

    builder
        .RegisterInstance(new MyClass())
        .OnRelease(s => s.Dispose());

    using (var container = builder.Build())
    {
        using (var scope = container.BeginLifetimeScope())
        {
            var service = scope.Resolve<MyClass>();
        }
    }

    MyClass.Disposed.Should().Be(true);
}
```

Strangely, and differently from the previous case, if the service is never resolved, `Dispose` is in fact not called. The following test fails:

```csharp
[Fact]
void autofac_does_does_not_invoke_OnRelease_if_service_is_never_resolved()
{

    builder
        .RegisterInstance(new MyClass())
        .OnRelease(s => s.Dispose());

    using (var container = builder.Build())
    {
        using (var scope = container.BeginLifetimeScope())
        {
//            var service = scope.Resolve<MyClass>();
        }
    }

    MyClass.Disposed.Should().Be(true);
}
```

This means that the clean up of `MyClass` instance is never performed:

* `Dispose` is not called, because the automatic disposal of the provided instance has been disabled  by the use of `OnRelease` and replaced by the provided callback;
* but `OnRelease` is never triggered, because no instance of `MyClass` service has never been requested.

Consequently, the instance is never cleaned up, potentially leading to a resource leak.
We are not sure this is a bug, but it sounds a bit confusing. Both v4 and v5 have the same behavior.

alistairjevans:
I considered other more complicated options to fix this that didn't use AutoActivate, but simplicity won out here.

I'll put a reference to the original issue on the pipeline event issue, to be considered when we get there.

### DIFF
--- a/test/Autofac.Specification.Test/Lifetime/LifetimeEventTests.cs
+++ b/test/Autofac.Specification.Test/Lifetime/LifetimeEventTests.cs
@@ -153,6 +153,36 @@ namespace Autofac.Specification.Test.Lifetime
             Assert.False(dt.IsDisposed);
         }
 
+        [Fact]
+        public void OnReleaseForSingletonStillFiresIfNotResolved()
+        {
+            var builder = new ContainerBuilder();
+
+            var instance = new ReleasingClass();
+
+            builder.RegisterInstance(instance)
+                   .OnRelease(s => s.Dispose());
+
+            using (var container = builder.Build())
+            {
+                using (var scope = container.BeginLifetimeScope())
+                {
+                }
+            }
+
+            Assert.True(instance.Disposed);
+        }
+
+        private class ReleasingClass : IDisposable
+        {
+            public bool Disposed { get; set; }
+
+            public void Dispose()
+            {
+                Disposed = true;
+            }
+        }
+
         private class MethodInjection
         {
             public int Param { get; private set; }
--- a/test/Autofac.Specification.Test/Lifetime/LifetimeEventTests.cs
+++ b/test/Autofac.Specification.Test/Lifetime/LifetimeEventTests.cs
@@ -161,7 +161,7 @@ namespace Autofac.Specification.Test.Lifetime
             var instance = new ReleasingClass();
 
             builder.RegisterInstance(instance)
-                   .OnRelease(s => s.Dispose());
+                   .OnRelease(s => s.Release());
 
             using (var container = builder.Build())
             {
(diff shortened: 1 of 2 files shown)


test/Autofac.Specification.Test/Lifetime/LifetimeEventTests.cs
```diff
@@ -153,6 +153,36 @@ public void ReleaseStopsAutomaticDisposal()
...
+            {
+                using (var scope = container.BeginLifetimeScope())
+                {
+                }
+            }
+
+            Assert.True(instance.Disposed);
+        }
+
+        private class ReleasingClass : IDisposable
```
tillig, line 176:
```suggestion
        private class ReleasingClass
```

Since we have a test [that shows disposal for provided `IDisposable` even when not resolved](https://github.com/autofac/Autofac/blob/f75c2276d44c599afa84fe1b383cab78402f8164/test/Autofac.Specification.Test/Lifetime/ProvidedInstanceTests.cs#L113), this class, for clarity, should probably _not_ implement `IDisposable` to ensure it's really `OnRelease`  doing the work and not some bug that might get introduced in the future around `IDisposable` handling.

(Also, if you need something `IDisposable` that tracks disposal, the `Util` namespace in the test project has a `DisposalTracker` class for that purpose.)

alistairjevans: Excellent point (and I'd forgotten about DisposalTracker).

Updated.

Issue #1107, opened 2020-04-21
TypeExtensions.TypeIsCompatibleWithGenericParameterConstraints doesn't handle new() constraint correctly


EikeStein:
### Summary
In a large project I'm working on we encoutered a hard to debug exception (Generic contraint violation). Due to the complexity of that project I had to narrow it down to create a usable bug report. The Problem appears to be in TypeExtensions.TypeIsCompatibleWithGenericParameterConstraints.

### Reproduction

``` CSharp
private class A<T> where T : class, new()
{ }

private class B
{
    internal B()
    { }
}

[Test]
public void ThisTestFails()
{
    // A<B> --> compiler: B must be a non-abstract type with a public parameterless constructor in order to use it as parameter 'T' in the generic type or method A<T>

    var type = typeof(Autofac.Util.AssemblyExtensions).Assembly.GetTypes().Single(t => t.FullName == "Autofac.Util.TypeExtensions");
    var methodInfo = type.GetMethod("IsCompatibleWithGenericParameterConstraints", BindingFlags.Static|BindingFlags.Public);
    var result = (bool?)methodInfo?.Invoke(null, new object[]
    {
        typeof(A<>),
        new [] { typeof(B) }
    });  
    Assert.IsFalse(result); // Assertion fails
}
```
### Version
[Current Source](https://github.com/autofac/Autofac/commit/ed92a3d8f3153ae73f840a4a4d918ff8497c0310) as well as 4.6.2

### Details

I our project we dynamically register services with generics. One construtor depends on a lot of other types and if I am not mistaken that makes AutoFac check for possible insertions and generic substitutes. The exception is thrown in OpenGenericServiceBinder.TryBindServiceType at the line 
`var constructedImplementationTypeTmp = openGenericImplementationType.MakeGenericType(implementorGenericArguments);`
because the check for compatibility via `IsCompatibleWithGenericParameterConstraints` mistakenly returns true, although the "new()" contraint is only fulfilled by an internal constructor.

### Suggestion:
Add check for parameterless and public instead of just parameterless cunstructor.

tillig:
Is there a way to illustrate the problem without reflecting into the internals?

What's the actual problem, that the exception you got was wrong, or just that it was hard to understand? Bear in mind we're not the compiler, so if you're doing some crazy detailed low level reflection stuff to register types and violating constraints, that my not always be something we could (or should) try enforcing in Autofac. C# changes over time and it's not feasible to try to catch everything.

[...]

PR #1109, opened 2020-04-22
#1107 Fixed generic contraint violation exception being thrown

EikeStein:
#1107 Fixed generic contraint violation exception being thrown by including an additional check to search for public and parameterless constructors, instead of just parameterless constructors when faced with a "where T : new()" contraint.


test/Autofac.Test/Core/Resolving/ResolveWithGenericTypeConstraintTests.cs
```diff
@@ -0,0 +1,21 @@
+﻿using Autofac.Test.Scenarios.GenericContraints;
+using Xunit;
+
+namespace Autofac.Test.Core.Resolving
+{
+    public class ResolveWithGenericTypeConstraintTests
```
tillig, line 6: Could you put this test in [this fixture](https://github.com/autofac/Autofac/blob/develop/test/Autofac.Specification.Test/Registration/OpenGenericTests.cs)?

It's not obvious, but we're trying to create tests that act against the public API (the "specification tests") and keep them separate from the tests that use internals. That way as we change internals, we know the spec tests should continue to pass. We're still trying to refactor those out of the original gigantic test assembly.

Maybe create a folder/namespace under that "Graph1" to follow suit with [these complex graph resolution test stubs](https://github.com/autofac/Autofac/tree/develop/test/Autofac.Specification.Test/Resolution/Graph1) and avoid clutter in the location with the other tests.

EikeStein: Should be done

PR #1110, opened 2020-04-25
Fixes #1108 - Stop secondary invocation of OnActivating() handlers with decorated instance

### LINKED ISSUE #1108: OnActivating()/OnActivated() not playing nicely with RegisterDecorator()
First of all, thank you for the decorator support added in 4.9.1 and its subsequent refinements! It's a far cry better than the clunky approaches we'd tried implementing on our own or event the older RegisterDecorator() support from earlier versions.

This seems to be at least somewhat related to #860, in terms of being in a similar or same area of code.

Using Autofac 5.1.2, I noticed that the lifetime events OnActivating() and OnActivated() don't quite work the way that might make the most sense when also using RegisterDecorator().

Here are some unit tests with basic comments on success/failure, and I'll describe more below them:

```csharp
using Autofac;
using NUnit.Framework;

namespace NextIT.ActiveAgent.Ioc.Test
{
    [TestFixture]
    public class TempTest
    {
        private interface ISample
        {
            string Injectable { get; set; }
        }

        private class Sample : ISample
        {
            public string Injectable { get; set; }
        }

        private class SampleDecorator : ISample
        {
            public readonly ISample Decoratee;

            public string Injectable
            {
                get => Decoratee.Injectable;
                set => Decoratee.Injectable = value;
            }

            public SampleDecorator(ISample decoratee)
            {
                Decoratee = decoratee;
            }
        }

        [Test]
        public void RegisterDecorator_with_OnActivating_decorates_all_relevant_registered_types()
        {
            builder.RegisterType<Sample>()
                   .As<ISample>()
                   .OnActivating(args => args.Instance.Injectable = "String time!");
            builder.RegisterDecorator<SampleDecorator, ISample>();

            //Throws cast exception: Unable to cast object of type 'SampleDecorator' to type 'Sample'
            Assert.That(() => container.Resolve<ISample>(), Throws.Nothing);
            var result = container.Resolve<ISample>();
            Assert.That(result, Is.TypeOf<SampleDecorator>());
        }

        [Test]
        public void RegisterDecorator_plus_OnActivating_interface_workaround_decorates_all_relevant_registered_types()
        {
            builder.RegisterType<Sample>().AsSelf();
            builder.Register<ISample>(c => c.Resolve<Sample>())
                   .OnActivating(args => args.Instance.Injectable = "String time!");
            builder.RegisterDecorator<SampleDecorator, ISample>();


            Assert.That(() => container.Resolve<ISample>(), Throws.Nothing);
            var result = container.Resolve<ISample>();
            Assert.That(result, Is.TypeOf<SampleDecorator>());
            //Success!
        }

        [Test]
        public void RegisterDecorator_with_OnActivated_receives_decorated_instance()
        {
            object activatedInstance = null;
            builder.RegisterType<Sample>()
                   .As<ISample>()
                   .OnActivated(args => activatedInstance = args.Instance);
            builder.RegisterDecorator<SampleDecorator, ISample>();

            Assert.That(() => container.Resolve<ISample>(), Throws.Nothing);
            var result = container.Resolve<ISample>();
            Assert.That(result, Is.TypeOf<SampleDecorator>());
            //Assertion Exception: activatedInstance type = Sample, result type = SampleDecorator
            Assert.That(result, Is.SameAs(activatedInstance));
        }

        [Test]
        public void RegisterDecorator_plus_OnActivated_interface_workaround_receives_decorated_instance()
        {
            object activatedInstance = null;
            builder.RegisterType<Sample>().AsSelf();
            builder.Register<ISample>(c => c.Resolve<Sample>())
                   .OnActivated(args => activatedInstance = args.Instance);
            builder.RegisterDecorator<SampleDecorator, ISample>();


            Assert.That(() => container.Resolve<ISample>(), Throws.Nothing);
            var result = container.Resolve<ISample>();
            Assert.That(result, Is.TypeOf<SampleDecorator>());
            //Assertion Exception: activatedInstance type = Sample, result type = SampleDecorator
            Assert.That(result, Is.SameAs(activatedInstance));
        }
    }
}
```

With regards to OnActivated(), I would have thought that at least the workaround using Register<ISample>() would have invoked the handler with the final decorator, as that is the completed instance that has gone through all of the machinations. I can understand how the OnActivated() against Register<Sample>() could end up being invoked as soon as the concrete instance is constructed, since the relative context is based around that type, though would create an inconsistency compared to the perspective of ISample. I don't know that any of this behavior is a *bug*, but may shed light on an area of enhancement to allow for performing "post-processing" on an instance after it's been decorated, such as resolving observers to register their event handlers for events exposed by the class/interface being resolved.

The scenario with OnActivating(), on the other hand, only functions when using the Register<ISample>(c => c.Resolve<Sample>()) workaround that had been [identified a number of years ago](https://groups.google.com/forum/#!topic/autofac/BPdp3RpGbBw) and added to the [wiki](https://autofaccn.readthedocs.io/en/latest/lifetime/events.html#onactivating) (by a former coworker of mine, no less!). It does look like the OnActivating() handler gets called twice: once with an instance of Sample, and again with an instance of SampleDecorator, which seems potentially undesirable and unexpected. I can see why it would behave that way though, since the container sees SampleDecorator as another ISample that is "activating" though it's part of the same resolution chain/graph...

In the case of Register<Sample>().As<ISample>().OnActivating(...), if you register a decorator for ISample, resolving ISample results in this exception:
```
Autofac.Core.DependencyResolutionException: An exception was thrown while executing a resolve operation. See the InnerException for details. ---> System.InvalidCastException: Unable to cast object of type 'SampleDecorator' to type 'Sample'.
   at Autofac.Builder.RegistrationBuilder`3.<>c__DisplayClass32_0.<OnActivating>b__0(Object s, ActivatingEventArgs`1 e)
   at Autofac.Core.Registration.ComponentRegistration.RaiseActivating(IComponentContext context, IEnumerable`1 parameters, Object& instance)
   at Autofac.Core.Resolving.InstanceLookup.Execute()
   at Autofac.Core.Resolving.ResolveOperation.GetOrCreateInstance(ISharingLifetimeScope currentOperationScope, ResolveRequest request)
   at Autofac.Core.Resolving.ResolveOperation.Execute(ResolveRequest request)
   --- End of inner exception stack trace ---
   at Autofac.Core.Resolving.ResolveOperation.Execute(ResolveRequest request)
   at Autofac.ResolutionExtensions.TryResolveService(IComponentContext context, Service service, IEnumerable`1 parameters, Object& instance)
   at Autofac.ResolutionExtensions.ResolveService(IComponentContext context, Service service, IEnumerable`1 parameters)
   at Autofac.ResolutionExtensions.Resolve[TService](IComponentContext context, IEnumerable`1 parameters)
   at TempTest.<>c__DisplayClass3_0.<RegisterDecorator_with_OnActivating_decorates_all_relevant_registered_types>b__1() in TempTest.cs:line 49
```
In that unit test, the OnActivating() handler is called with an instance of Sample, which succeeds, but the same behavior of the handler being invoked with the decorator instance next fails with the invalid cast exception.

I'll do some more investigation and see if I can scare up any more details, but wanted to get this info out for the sake of gathering opinions/feedback/corrections. My primary concern is in the fact that I work on a product that uses Autofac to facilitate a plugin architecture, where plugins might only have knowledge about some common interfaces in the product itself, and their registrations could indirectly impact each other in unexpected ways and create some difficult debugging adventures. We can close off some doors, as needed to protect the consumers, but it's nice to be able to provide a robust set of capabilities.

- Assembly version information
```
<ItemGroup>
    <PackageReference Include="Autofac" Version="5.1.2" />
    <PackageReference Include="NUnit" Version="3.12.0" />
</ItemGroup>
```

VonOgre:
Fixes #1108, specifically the most critical part, as it stops the casting exception from occurring or unexpectedly calling an OnActivating() delegate multiple times.

It also brings OnActivating() back into alignment with the other lifetime events when decorators are in play

### DIFF
--- a/src/Autofac/Core/Resolving/InstanceLookup.cs
+++ b/src/Autofac/Core/Resolving/InstanceLookup.cs
@@ -102,9 +102,6 @@ namespace Autofac.Core.Resolving
                 _activationScope,
                 resolveParameters);
 
-            if (instance != decoratorTarget)
-                ComponentRegistration.RaiseActivating(this, resolveParameters, ref instance);
-
             var handler = InstanceLookupEnding;
             handler?.Invoke(this, new InstanceLookupEndingEventArgs(this, NewInstanceActivated));
 
--- a/test/Autofac.Specification.Test/Features/DecoratorTests.cs
+++ b/test/Autofac.Specification.Test/Features/DecoratorTests.cs
@@ -749,6 +749,92 @@ namespace Autofac.Specification.Test.Features
         }
 
         [Fact]
+        public void ResolvesDecoratedServiceWhenTargetHasOnActivatingHandlerOnType()
+        {
+            var activatingInstances = new List<object>();
+
+            var builder = new ContainerBuilder();
+
+            builder.RegisterType<ImplementorA>().As<IDecoratedService>()
+                .OnActivating(args => activatingInstances.Add(args.Instance));
+            builder.RegisterDecorator<DecoratorA, IDecoratedService>();
+            var container = builder.Build();
+
+            var instance = container.Resolve<IDecoratedService>();
+
+            Assert.IsType<DecoratorA>(instance);
+            Assert.IsType<ImplementorA>(instance.Decorated);
+
+            Assert.Single(activatingInstances);
+            Assert.IsType<ImplementorA>(activatingInstances[0]);
+        }
+
+        [Fact]
+        public void ResolvesDecoratedServiceWhenTargetHasOnActivatingHandlerOnInterface()
+        {
+            var activatingInstances = new List<object>();
+
+            var builder = new ContainerBuilder();
+
+            builder.RegisterType<ImplementorA>().AsSelf();
+            builder.Register<IDecoratedService>(c => c.Resolve<ImplementorA>())
+                .OnActivating(args => activatingInstances.Add(args.Instance));
+            builder.RegisterDecorator<DecoratorA, IDecoratedService>();
+            var container = builder.Build();
+
+            var instance = container.Resolve<IDecoratedService>();
+
+            Assert.IsType<DecoratorA>(instance);
+            Assert.IsType<ImplementorA>(instance.Decorated);
+
+            Assert.Single(activatingInstances);
+            Assert.IsType<ImplementorA>(activatingInstances[0]);
+        }
+
+        [Fact]
+        public void ResolvesDecoratedServiceWhenTargetHasOnActivatedHandlerOnType()
+        {
+            var activatedInstances = new List<object>();
+
+            var builder = new ContainerBuilder();
+
+            builder.RegisterType<ImplementorA>().As<IDecoratedService>()
+                .OnActivated(args => activatedInstances.Add(args.Instance));
+            builder.RegisterDecorator<DecoratorA, IDecoratedService>();
+            var container = builder.Build();
+
+            var instance = container.Resolve<IDecoratedService>();
+
+            Assert.IsType<DecoratorA>(instance);
+            Assert.IsType<ImplementorA>(instance.Decorated);
+
+            Assert.Single(activatedInstances);
+            Assert.IsType<ImplementorA>(activatedInstances[0]);
+        }
+
+        [Fact]
+        public void ResolvesDecoratedServiceWhenTargetHasOnActivatedHandlerOnInterface()
+        {
+            var activatedInstances = new List<object>();
+
+            var builder = new ContainerBuilder();
+
+            builder.RegisterType<ImplementorA>().AsSelf();
+            builder.Register<IDecoratedService>(c => c.Resolve<ImplementorA>())
+                .OnActivated(args => activatedInstances.Add(args.Instance));
+            builder.RegisterDecorator<DecoratorA, IDecoratedService>();
+            var container = builder.Build();
+
+            var instance = container.Resolve<IDecoratedService>();
+
+            Assert.IsType<DecoratorA>(instance);
+            Assert.IsType<ImplementorA>(instance.Decorated);
+
+            Assert.Single(activatedInstances);
+            Assert.IsType<ImplementorA>(activatedInstances[0]);
+        }
+
+        [Fact]
         public void StartableTypesCanBeDecorated()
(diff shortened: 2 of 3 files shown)


src/Autofac/Core/Resolving/InstanceLookup.cs
```diff
@@ -102,9 +102,6 @@ public object Execute()
                 _activationScope,
                 resolveParameters);
 
-            if (instance != decoratorTarget)
```
VonOgre, line 105: Unless I'm missing something specific, you wouldn't actually want to raise the Activating event again for the same registration, but specify the decorated instance. CreateInstance() already raises Activating for the original, so this block that's removed is a secondary occurrence only happening with decorators in play.

VonOgre:
@tillig - I can definitely dig into the first one and shore up some additional tests, for the sake of adding robustness, but the second ask gives me pause, since that seems to contradict the fix that I added. Alternatively, are you asking that appropriate tests be added to ensure ```Activating``` events are raised for decorating classes that have their own ```OnActivating()``` handlers added?

I don't know that decorating classes can have their own ```OnActivating()``` handlers, because all of the ```RegisterDecorator()``` overloads are ```void```, . At best, you might be able to have something akin to:
```
builder.Register<MyDecorator>().OnActivating(...);
builder.RegisterDecorator<MyDecorator, IService>();
```
where you might have an ```Activating``` event that works, at that point?

I'll do some more investigation, but wanted to make sure I understood the ask!

tillig:
Oh, no, you're right; I was thinking about the way the decorators automatically share the lifetime of the component to which they're attached and incorrectly extrapolated that there are also the same lifetime events for each decorator. Could have sworn that was a thing. Hmmm.

Anyway, skip that second bullet. If we have some tests that just ensure we haven't incorrectly removed events or doubled events where they already were supposed to be, we'll be good. Sorry for the confusion. It's a hell of a day today. :)

VonOgre:
I'll investigate and reinforce tests as best I can in the next day or so.

No apologies needed, this isn't exactly an easy topic to visualize or
articulate! Plus it's Monday, which dooms us all ;)

PR #1114, opened 2020-05-02
WIP: Fixes #1113

### LINKED ISSUE #1113: RegisterDecorator corrupts resolution when a target registration has greater lifetime than per-dependency and IEnumerable<> is resolved
Reproduced in Autofac trunk DecoratorTests, but also noted in Autofac 5.1.2.

If one or more services (e.g. IService) are registered with a specific lifetime (InstancePerLifetimeScope, SingleInstance), resolving IEnumerable<IService> will return instances of all of  the types as normal, but adding a decorator into the mix breaks the IEnumerable<IService> resolution, replacing all instances with the first "reusable" instance. Similarly, once IEnumerable<IService> is resolved, trying to resolve IService no longer returns the last registration, but instead also returns the "reusable" decorated instance noted in the IEnumerable<IService> case

```
[Fact]
        public void CanResolveMultipleDecoratedServicesSingleInstance()
        {
            builder.RegisterType<ImplementorA>().As<IDecoratedService>();
            builder.RegisterType<ImplementorB>().As<IDecoratedService>().SingleInstance();
            builder.RegisterDecorator<DecoratorA, IDecoratedService>();

            var services = container.Resolve<IEnumerable<IDecoratedService>>();

            Assert.Collection(
                services,
                s =>
                {
                    Assert.IsType<DecoratorA>(s);
                    Assert.IsType<ImplementorA>(s.Decorated);
                },
                s =>
                {
                    Assert.IsType<DecoratorA>(s);
                    Assert.IsType<ImplementorB>(s.Decorated);
                });

            services = container.Resolve<IEnumerable<IDecoratedService>>();

            Assert.Collection(
                services,
                s =>
                {
                    Assert.IsType<DecoratorA>(s);
                    //Fails because s.Decorated is of type ImplementorB
                    Assert.IsType<ImplementorA>(s.Decorated);
                },
                s =>
                {
                    Assert.IsType<DecoratorA>(s);
                    Assert.IsType<ImplementorB>(s.Decorated);
                });
        }

[Fact]
        public void CanResolveSingleServiceAfterMultipleDecoratedServicesSingleInstance()
        {
            builder.RegisterType<ImplementorA>().As<IDecoratedService>().SingleInstance();
            builder.RegisterType<ImplementorB>().As<IDecoratedService>();
            builder.RegisterDecorator<DecoratorA, IDecoratedService>();

            container.Resolve<IEnumerable<IDecoratedService>>();
            var service = container.Resolve<IDecoratedService>();

            Assert.IsType<DecoratorA>(service);
            //Fails because s.Decorated is of type ImplementorA
            Assert.IsType<ImplementorB>(service.Decorated);
        }
```

VonOgre:
Introduced class that can generate predictable GUIDS, per RFC 4122 subsection 4.3, and used in InstanceLookup to generate a sort of "composite" guid to store shared instances of decorators in isolation.

Raising this as a WIP to get feedback on approach. If approved, I'll add unit tests for the DeterministicGuid class.

### DIFF
--- a/src/Autofac/Core/Resolving/InstanceLookup.cs
+++ b/src/Autofac/Core/Resolving/InstanceLookup.cs
@@ -30,6 +30,7 @@ using System.Globalization;
 using System.Linq;
 using System.Text;
 using Autofac.Builder;
+using Autofac.Core.Lifetime;
 using Autofac.Features.Decorators;
 
 namespace Autofac.Core.Resolving
@@ -86,10 +87,14 @@ namespace Autofac.Core.Resolving
 
             var resolveParameters = Parameters as Parameter[] ?? Parameters.ToArray();
 
-            if (!_activationScope.TryGetSharedInstance(ComponentRegistration.Id, out var instance))
+            var instanceKey = _decoratorTargetComponent != null ?
+                new SharedInstanceKey(new[] { ComponentRegistration.Id, _decoratorTargetComponent.Id }) :
+                new SharedInstanceKey(ComponentRegistration.Id);
+
+            if (!_activationScope.TryGetSharedInstance(instanceKey, out var instance))
             {
                 instance = sharing == InstanceSharing.Shared
-                    ? _activationScope.CreateSharedInstance(ComponentRegistration.Id, () => CreateInstance(Parameters))
+                    ? _activationScope.CreateSharedInstance(instanceKey, () => CreateInstance(Parameters))
                     : CreateInstance(Parameters);
             }
 
--- a/test/Autofac.Specification.Test/Features/DecoratorTests.cs
+++ b/test/Autofac.Specification.Test/Features/DecoratorTests.cs
@@ -209,6 +209,341 @@ namespace Autofac.Specification.Test.Features
         }
 
         [Fact]
+        public void CanResolveMultipleDecoratedServicesWithMultipleDecorators()
+        {
+            var builder = new ContainerBuilder();
+            builder.RegisterType<ImplementorA>().As<IDecoratedService>();
+            builder.RegisterType<ImplementorB>().As<IDecoratedService>();
+            builder.RegisterDecorator<DecoratorA, IDecoratedService>();
+            builder.RegisterDecorator<DecoratorB, IDecoratedService>();
+            var container = builder.Build();
+
+            var services = container.Resolve<IEnumerable<IDecoratedService>>();
+
+            Assert.Collection(
+                services,
+                s =>
+                {
+                    Assert.IsType<DecoratorB>(s);
+                    Assert.IsType<DecoratorA>(s.Decorated);
+                    Assert.IsType<ImplementorA>(s.Decorated.Decorated);
+                },
+                s =>
+                {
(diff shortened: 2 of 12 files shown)


src/Autofac/Core/Resolving/InstanceLookup.cs
```diff
@@ -83,13 +84,21 @@ public object Execute()
             _executed = true;
 
             var sharing = _decoratorTargetComponent?.Sharing ?? ComponentRegistration.Sharing;
+            var sharedId = ComponentRegistration.Id;
+            if (_decoratorTargetComponent != null)
+            {
+                // Issue #1113
+                // Need to generate a GUID that represents the composite of the decorator and target registrations
+                // to retain multiple shared instances for IEnumerable<T>
+                sharedId = DeterministicGuid.Create(_decoratorTargetComponent.Id, sharedId.ToByteArray());
```
VonOgre, line 93: Since it is possible for multiple instances of a given decorator to be generated via IEnumerable<T>, the ID used to store said instances can't all be the ID for the decorator registration, since that just stomps all over itself. "Namespacing" the decorator registration ID with the registration ID of the target creates uniqueness with a very low chance of collision.

src/Autofac/Util/DeterministicGuid.cs
```diff
@@ -0,0 +1,135 @@
...
+using System.Security.Cryptography;
+using System.Text;
+#pragma warning disable CA5350 // Do Not Use Weak Cryptographic Algorithms - SHA1 required by specification
+
+namespace Autofac.Util
+{
+    /// <summary>
+    /// Utility class for generating predictable <see cref="Guid"/>s, per RFC 4122 §4.3.
+    /// </summary>
+    internal static class DeterministicGuid
```
VonOgre, line 12: This class is based on things I've done for my day job, which was derived from some stackoverflow posts. I wasn't sure what the practice was for comments regarding attribution/credit, but I can definitely add the SO answer for posterity.

test/Autofac.Specification.Test/Features/DecoratorTests.cs
```diff
@@ -208,6 +208,341 @@ public void CanResolveMultipleDecoratedServices()
                 });
         }
 
+        [Fact]
+        public void CanResolveMultipleDecoratedServicesWithMultipleDecorators()
```
VonOgre, line 212: Most of these new tests fail if the change to InstanceLookup is reverted.

alistairjevans:
My gut says that the extra work to generate a deterministic guid is going to be pretty expensive; the ``DeterministicGuid.Create`` function does a lot of work, with a number of allocations, and a SHA1 operation in the mix.

We don't actually have any benchmarks for decorators attached to components with a shared lifetime, and we would definitely need them for this level of change.

If you wanted to add those benchmarks (perhaps in a separate PR), we can use them to assess the impact of any changes to the sharing Guid behaviour?

VonOgre:
I'm going to abandon this pull request and separate benchmarks into one and the fix for #1113 into another

PR #1128, opened 2020-05-28
Seal container builder

### LINKED ISSUE #1120: Seal ContainerBuilder
The `ContainerBuilder` has no virtual methods and is intended to be extended using custom extension methods rather than derive-and-override (or derive-and-new, as the case may be). Some recent conversations indicate there may be a small number of folks who have incorrectly created their own `ContainerBuilder`, which is sort of an unforeseen and not-really-supported method of extension.

I'm thinking if we seal `ContainerBuilder` we can keep folks on the path to success.

Any reason we _shouldn't_ do that?

If so, this should _probably_ be a 6.0 change, but... I dunno, I could be convinced it's small enough impact it could be a 0.1.0 increment in 5.x.

alistairjevans:
Nice and easy; fixes #1120.


alistairjevans:
Whoops, accidentally included another change. I'll try to split it out.

PR #1137, opened 2020-05-31
Replaced linq Any calls with Length or Count

DamirAinullin:
to improve perfomance


DamirAinullin:
@alistairjevans Is it OK now?

PR #1138, opened 2020-05-31
Replaced linq first calls with [0]

DamirAinullin:
to improve perfomance


alexmg:
Thanks for the PR @DamirAinullin

PR #1140, opened 2020-06-04
Micro Perf - Store a value for 'trace is enabled'

alistairjevans:
The 'pipeline progression' function that implements the actual next() call is an extremely hot path, and checks for a Tracer to see if it exists in each call. 

This change is a small optimisation to store a 'TracingEnabled' flag at the start of the request, to speed up that check slightly.

### DeepGraphBenchmark

this PR

| Resolve | 17.55 us | 0.221 us | 0.196 us | 4.0894 |     - |     - |  16.71 KB |

v6

| Resolve | 17.83 us | 0.265 us | 0.222 us | 4.0894 |     - |     - |  16.71 KB |

### DIFF
--- a/src/Autofac/Core/Resolving/Pipeline/ResolvePipelineBuilder.cs
+++ b/src/Autofac/Core/Resolving/Pipeline/ResolvePipelineBuilder.cs
@@ -269,14 +269,9 @@ namespace Autofac.Core.Resolving.Pipeline
                 return (ctxt) =>
                 {
                     // Optimise the path depending on whether a tracer is attached.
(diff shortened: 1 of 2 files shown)


tillig:
>Dango is a Japanese dumpling and sweet made from mochiko, related to mochi. It is often served with green tea. Dango is eaten year-round, but the different varieties are traditionally eaten in given seasons. Three to five dango are often served on a skewer.

alistairjevans:
I think the unicode consortium team have been given too many bytes to put emojis in...

PR #1141, opened 2020-06-04
Enabled SA1600 (Elements must have docs) - Added required comments

alistairjevans:
I've turned on S1600 in the analyzer ruleset and added the missing comments. There were only about 200 of them...

I've also added a suppression file for the test projects so they don't need docs, because the ruleset can't vary between projects.

### DIFF
--- a/build/Analyzers.ruleset
+++ b/build/Analyzers.ruleset
@@ -1,61 +1,33 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<RuleSet Name="Autofac Analyzer Rules" Description="Analyzer rules for Autofac assemblies." ToolsVersion="14.0">
-  <!-- https://github.com/dotnet/roslyn/blob/master/docs/compilers/Rule%20Set%20Format.md -->
+<RuleSet Name="Autofac Analyzer Rules" Description="Analyzer rules for Autofac assemblies." ToolsVersion="16.0">
   <IncludeAll Action="Warning" />
   <Rules AnalyzerId="Microsoft.Usage" RuleNamespace="Microsoft.Usage">
-    <!-- Implement standard exception constructors - not all of the exception constructors (e.g., parameterless) are desired in our system. -->
     <Rule Id="CA1032" Action="None" />
-    <!-- Change names to avoid reserved word overlaps (e.g., Delegate, GetType, etc.) - too many of these in the public API, we'd break if we fixed it. -->
     <Rule Id="CA1716" Action="None" />
-    <!-- Implement serialization constructors - false positive when building .NET Core -->
     <Rule Id="CA2229" Action="None" />
-    <!-- Mark ISerializable types with SerializableAttribute - false positive when building .NET Core -->
     <Rule Id="CA2237" Action="None" />
   </Rules>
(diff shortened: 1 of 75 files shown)


alistairjevans:
If I had a dollar for every time I've absent-mindedly removed a method because CodeLens tells me it's unused, for everything to break horribly.

build/Analyzers.ruleset
```diff
@@ -1,61 +1,33 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<RuleSet Name="Autofac Analyzer Rules" Description="Analyzer rules for Autofac assemblies." ToolsVersion="14.0">
-  <!-- https://github.com/dotnet/roslyn/blob/master/docs/compilers/Rule%20Set%20Format.md -->
+<RuleSet Name="Autofac Analyzer Rules" Description="Analyzer rules for Autofac assemblies." ToolsVersion="16.0">
```
tillig, line 2: I think the comments about the rules were helpful. Can we revert this?

tillig: (Keep the updated ToolsVersion, but put the comments back, I mean.)

src/Autofac/Features/GeneratedFactories/GeneratedFactoryRegistrationSource.cs
```diff
@@ -32,6 +32,9 @@
 
 namespace Autofac.Features.GeneratedFactories
 {
+    /// <summary>
+    /// Registration source for generated factory methods (i.e. when resolving <![CDATA[Func<TService>]]> or some variant).
```
tillig, line 36:
```suggestion
    /// Registration source for generated factory methods (i.e. when resolving <see cref="Func{T}"/> or some variant).
```

test/TestSuppressions.cs
```diff
@@ -0,0 +1,5 @@
+﻿// Suppressions file for test projects.
```
tillig, line 1: Is this better than `NoWarn` on SA1600? (I don't have a real preference, just curious if you find one way to be better than the other.)

In other projects I've done, I actually have two different rulesets - one for production code, one for tests. The test one is less stringent. Maybe something to consider in the future.

alistairjevans: I just did it like this so I could share it, but perhaps it's better to put the NoWarn in a Directory.props file in the tests folder?

build/Analyzers.ruleset
alistairjevans: Yeah, I can do that; Visual Studio must have borked them.

test/TestSuppressions.cs
tillig: Meh. It's fine for now. Some later point I might put two different ruleset files in there for better control, but I'm not going to be a stickler over it.

src/Autofac/Features/GeneratedFactories/GeneratedFactoryRegistrationSource.cs
alistairjevans: Yeah, I don't why I didn't just do that...

test/TestSuppressions.cs
alistairjevans: Whoops, too late!

PR #1144, opened 2020-06-05
Add implicit registration source helper class

### LINKED ISSUE #1142: Add RegistrationSource base class to simplify custom implicit relationships
I recently needed to implement my own implicit relationship and found it a bit more challenging than I expected. The challenges I hit were:

- Metadata order is not easily retrieved with public APIs
- Some of the type helpers are not available

I was able to work around these, but I think there is an opportunity for a simple, extensible API for simple use cases like this. This would be to create implicit relationships similar to `Lazy<T>`, `Owned<T>`, and others.

Here's an example API:

```csharp
  public abstract class RegistrationSource : IRegistrationSource
    {
        private static readonly MethodInfo CreateRegistrationMethod = typeof(RegistrationSource).GetTypeInfo().GetDeclaredMethod(nameof(CreateRegistrationInternal));

        private readonly Type _type;

        protected RegistrationSource(Type type)
        {
            _type = type;
        }

        public abstract bool IsAdapterForIndividualComponents { get; }

        public abstract string Description { get; }

        protected abstract object CreateRegistration<T>(IComponentContext ctx, ResolveRequest request)  where T : notnull;

        protected virtual IRegistrationBuilder<object, SimpleActivatorData, SingleRegistrationStyle> BuildRegistration(IRegistrationBuilder<object, SimpleActivatorData, SingleRegistrationStyle> registration);
    }
```

This is modeled after the `LazyRegistrationSource`. I have a working example at https://github.com/twsouthwick/Autofac/tree/simple-registration-source where `Lazy<T>`, `Meta<T>`, and `Owned<T>` all use this instead of duplicating code themselves.

I may have missed some aspect of this, but would like to start a discussion to add this API.

twsouthwick:
Fixes #1142


src/Autofac/Features/ImplicitRegistrationSource.cs
```diff
@@ -0,0 +1,110 @@
+﻿// This software is part of the Autofac IoC container
```
alistairjevans, line 1: We don't currently have any types directly under Features; I think this class needs to go under Autofac.Core, at the same level as `IRegistrationSource`.

src/Autofac/Features/ImplicitRegistrationSource.cs
```diff
@@ -0,0 +1,110 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Autofac.Builder;
+using Autofac.Core;
+using Autofac.Util;
+
+namespace Autofac.Features
+{
+    public abstract class ImplicitRegistrationSource : IRegistrationSource
```
alistairjevans, line 36: This class should really have XML comments for the public-facing members (we're turning on the analyzer warning in v6, but it's not there yet in v5).

src/Autofac/Features/ImplicitRegistrationSource.cs
```diff
@@ -0,0 +1,110 @@
...
+
+            if (!(service is IServiceWithType swt) || !swt.ServiceType.IsGenericTypeDefinedBy(_type))
+            {
+                return Enumerable.Empty<IComponentRegistration>();
+            }
+
+            var valueType = swt.ServiceType.GetTypeInfo().GenericTypeArguments[0];
+            var valueService = swt.ChangeType(valueType);
+
+            var registrationCreator = CreateRegistrationMethod.MakeGenericMethod(valueType);
```
alistairjevans, line 72: This is perhaps not for now (this problem already existed); but might it be worth caching these generic methods when we generate them, to speed up nested scope benchmarks? Perhaps something to consider for v6.

src/Autofac/Features/ImplicitRegistrationSource.cs
twsouthwick: Done

src/Autofac/Features/ImplicitRegistrationSource.cs
twsouthwick: Done.

src/Autofac/Features/ImplicitRegistrationSource.cs
twsouthwick: Yes, I'd like to take a look at that, but probably in a separate PR. Looks like there's a bunch of opportunities for caching method/generic instantiations

PR #1145, opened 2020-06-06
Update a few build settings

twsouthwick:
- Allow rolling forward for global.json so later features versions are allowed
- Add Directory.Build.[props|targets] for solution level settings
- Moved compile as warning to error to solution level but only turned on for release builds. This helps so that while prototyping stuff, it does not fail to build on style stuff that can be cleaned up before the final commit.

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -1,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
+</Project>
--- a/global.json
+++ b/global.json
@@ -1,5 +1,6 @@
 {
   "sdk": {
-    "version": "3.1.200"
+    "version": "3.1.200",
+    "rollForward": "latestFeature"
   }
 }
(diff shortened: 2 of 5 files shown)


global.json
```diff
@@ -1,5 +1,6 @@
 {
   "sdk": {
-    "version": "3.1.200"
+    "version": "3.1.200",
+    "rollForward": "latestFeature"
```
alistairjevans, line 4: I didn't know about this, this will make our lives so much easier!

Directory.Build.targets
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
```
alistairjevans, line 1: Do we need an empty targets file?

twsouthwick: It's more for completion. It's not necessary, but by habit I added it for symmetry with the props file. It's useful, for instance, if you want to add analyzers for the whole project. I can remove it if you'd like.

alistairjevans: No, that's fine, it makes sense; I know that the directory targets files are super useful.

Issue #1127, opened 2020-05-28
Performance for v6.0


alistairjevans:
There's a few things left to do from a performance perspective following the pipelines changes.

- [ ] Run profiler on all the various benchmark cases to see if there are additional optimizations
- [x] `CircularDependencyMiddleware.Execute` stack sees PushWithResize a bit, might be worth allocating some memory up front
- [x] `ResolveOperationBase.ResetSuccessfulRequests` resets the list with new, but would it be better to Clear() or something else?

alistairjevans:
Addressed the `ResetSuccessfulRequests` item in #1131.

alistairjevans:
Addressed the Circular Dependency item in #1148.

alistairjevans:
Run the full benchmark set against `develop` vs `v6`; used the benchmark comparer tool used in the dotnet runtime; high-level view of the current state of play is below.

Priority next is to figure out the concurrency problem, then I can look at the others.

summary:
better: 42, geomean: 1.062
worse: 14, geomean: 1.052
total diff: 56

| Slower                                                                           | diff/base | Base Median (ns) | Diff Median (ns) | Modality|
| -------------------------------------------------------------------------------- | ---------:| ----------------:| ----------------:| --------:|
| Autofac.Benchmarks.ConcurrencyBenchmark.MultipleResolvesOnMultipleTasks(ResolveT |      1.10 |       2035909.77 |       2242665.63 |         |
| Autofac.Benchmarks.Decorators.KeylessSimpleSharedInstanceBenchmark.ResolveEnumer |      1.06 |          5138.37 |          5456.89 |         |
| Autofac.Benchmarks.Decorators.KeyedNestedBenchmark.ResolveEnumerableT(repetition |      1.05 |          4653.39 |          4907.33 |         |
| Autofac.Benchmarks.Decorators.KeylessSimpleLambdaBenchmark.ResolveEnumerableT(re |      1.05 |          5125.31 |          5388.90 |         |
| Autofac.Benchmarks.Decorators.KeylessNestedSharedInstanceLambdaBenchmark.Resolve |      1.05 |          4667.69 |          4904.86 |         |
| Autofac.Benchmarks.Decorators.KeylessSimpleBenchmark.ResolveEnumerableT(repetiti |      1.05 |          3025.03 |          3162.98 |         |
| Autofac.Benchmarks.EnumerableResolveBenchmark.ResolveArray                       |      1.04 |          2336.60 |          2438.25 |         |
[...]

PR #1148, opened 2020-06-09
Circular Dependency Changes

alistairjevans:
I know how much @tillig likes looking at a PR every day, so.....

I've made two changes to the way circular dependencies are detected.

### Max Depth
The first is to the 'max depth' constraint, which currently has a limit of 50. Based on @VonOgre's suggestion of `RuntimeHelpers.EnsureSufficientExecutionStack()` in #575 (👍), there is another method, `RuntimeHelpers.TryEnsureSufficientExecutionStack()`. On netstandard2.1, when we hit the normal 50 limit, rather than stop immediately, we keep going until `TryEnsureSufficientExecutionStack` returns false. 

These methods do not exist prior to netstandard2.1, so the existing '50' limit is applied.

This should please anyone with insanely deep service graphs; on my testing, I hit 891 resolve requests before it stopped. 

> Side note: tracing does still work on in this new style of stack-checking. It generates a *significant* amount of text though... 

### Request Stack Change - Depth and Segments

We talked already about pre-allocating the request stack in #1127 ; that has been done (default size of 16). That's actually pretty deep for 90% of users I'd guess; it'll then double each resize after that.

In addition, I wanted to fix an annoying problem with decorators, where they had to reach out and execute a top-level `ResolveOperation` instead of just resolving the decorator in the same `IComponentContext` resulting in a whole other allocation of the `ResolveOperation`, and generally some sad code. They did this because they needed a fresh circular dependency detection stack.

To counter this, I've replaced the use of `Stack<T>` with a new `SegmentedStack<T>`. This type behaves almost exactly like a regular stack, except it allows you define a point in the stack to start a new 'segment'. The enumerator for the stack then only operates within the segment. 

In this way we can mark a point in the request stack beyond which we do not look for any further dependencies.

In the `DecoratorMiddleware`:

```csharp
using (context.Operation.EnterNewDependencyDetectionBlock())
{
    var decoratedInstance = context.ResolveComponent(resolveRequest);

    context.Instance = decoratedInstance;

    context.DecoratorContext = context.DecoratorContext.UpdateContext(decoratedInstance);
}
```

This gives us some memory and performance gains on resolving decorators (KeylessGenericBenchmark):

**v6**
|             Method | repetitions |        Mean |    Error |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------- |------------ |------------:|---------:|-------:|------:|------:|----------:|
|           Baseline |           ? |    540.1 ns |  5.71 ns | 0.5102 |     - |     - |   2.09 KB |
|                    |             |             |          |        |       |       |           |
| ResolveEnumerableT |           1 |  3,931.6 ns | 40.18 ns | 1.7395 |     - |     - |   7.14 KB |
|           ResolveT |           1 |  1,973.3 ns | 24.02 ns | 1.0872 |     - |     - |   4.45 KB |
|                    |             |             |          |        |       |       |           |
| ResolveEnumerableT |           2 |  7,380.6 ns | 54.92 ns | 2.9831 |     - |     - |   12.2 KB |
|           ResolveT |           2 |  3,381.3 ns | 24.34 ns | 1.6670 |     - |     - |   6.82 KB |
|                    |             |             |          |        |       |       |           |
| ResolveEnumerableT |           3 | 10,589.0 ns | 76.76 ns | 4.2114 |     - |     - |  17.25 KB |
|           ResolveT |           3 |  4,688.6 ns | 21.50 ns | 2.2430 |     - |     - |   9.19 KB |

**request-stack**
|             Method | repetitions |        Mean |     Error |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------- |------------ |------------:|----------:|-------:|------:|------:|----------:|
|           Baseline |           ? |    537.8 ns |   4.15 ns | 0.5102 |     - |     - |   2.09 KB |
|                    |             |             |           |        |       |       |           |
| ResolveEnumerableT |           1 |  3,841.2 ns |  25.83 ns | 1.5793 |     - |     - |   6.47 KB |
|           ResolveT |           1 |  1,892.1 ns |  10.53 ns | 1.0090 |     - |     - |   4.13 KB |
|                    |             |             |           |        |       |       |           |
| ResolveEnumerableT |           2 |  7,234.6 ns |  25.49 ns | 2.6550 |     - |     - |  10.85 KB |
|           ResolveT |           2 |  3,091.5 ns |  19.88 ns | 1.5068 |     - |     - |   6.16 KB |
|                    |             |             |           |        |       |       |           |
| ResolveEnumerableT |           3 | 10,198.3 ns | 131.48 ns | 3.7231 |     - |     - |  15.23 KB |
|           ResolveT |           3 |  4,326.6 ns |  33.11 ns | 2.0065 |     - |     - |    8.2 KB |


tillig:
This was definitely a hole in my life, to be sure.

Looks like some good improvements, though, so... I guess I'll dig in. 😆

src/Autofac/Core/Resolving/Pipeline/ResolveRequestContextBase.cs
```diff
@@ -162,21 +162,10 @@ public void ChangeParameters(IEnumerable<Parameter> newParameters)
...
-        /// if you need to resolve a component with a completely separate operation and circular dependency verification stack.
-        /// </summary>
-        /// <param name="request">The resolve request.</param>
-        /// <returns>
-        /// The component instance.
-        /// </returns>
-        /// <exception cref="ComponentNotRegisteredException"/>
-        /// <exception cref="DependencyResolutionException"/>
-        public abstract object ResolveComponentWithNewOperation(ResolveRequest request);
+        public object ResolveComponent(ResolveRequest request)
```
tillig, line 165: Given this the base class should this be `virtual`?

src/Autofac/Core/Resolving/SegmentedStack.cs
```diff
@@ -0,0 +1,239 @@
...
+        }
+
+        /// <summary>
+        /// Push an item onto the stack.
+        /// </summary>
+        /// <param name="item">The item.</param>
+        public void Push(T item)
+        {
+            // No null check for item here; internally called method only, known to never be null, and is a very hot path.
+            var next = _next;
```
tillig, line 62: Is there a reason to assign local variables to the instance variables? (Like, is there some memory or perf optimization I'm unaware of? Not pointed - legit question.)

src/Autofac/Core/Resolving/ResolveOperationBase.cs
```diff
@@ -124,7 +124,17 @@ protected ResolveOperationBase(ISharingLifetimeScope mostNestedLifetimeScope, IR
...
+        internal SegmentedStack<ResolveRequestContextBase> RequestStack { get; } = new SegmentedStack<ResolveRequestContextBase>();
+
+        /// <summary>
+        /// Enter a new dependency chain block where subsequent requests inside the operation are allowed to repeat
+        /// registrations from before the block.
+        /// </summary>
+        /// <returns>A disposable that should be disposed to exit the block.</returns>
+        public IDisposable EnterNewDependencyDetectionBlock()
+        {
+            return RequestStack.EnterSegment();
```
tillig, line 136: I'm noticing `RequestDepth` is getting tracked separately and used in, say, `GetOrCreateInstance`. Does this new stack mechanism make that separate tracking no longer required?

test/Autofac.Specification.Test/Features/CircularDependencyTests.cs
```diff
@@ -94,6 +102,23 @@ public void InstancePerDependencyDoesNotAllowCircularDependencies_PropertyOwnerR
...
+        [Fact]
+        public void InstancePerDependencyDoesNotAllowCircularDependencies_PropertyOwnerResolved_WithTracerAttached()
+        {
+            var cb = new ContainerBuilder();
+            cb.RegisterType<DependsByCtor>();
+            cb.RegisterType<DependsByProp>().PropertiesAutowired(PropertyWiringOptions.AllowCircularDependencies);
+
+            var c = cb.Build();
+
+            c.AttachTrace((req, trace) =>
```
tillig, line 114: I was looking at the build output for the PR and didn't see a trace here; is this for some other debugging? Is it dumping out to the .trx file? (Do we need it? I'm OK if it is helpful.)

src/Autofac/Core/Resolving/Pipeline/ResolveRequestContextBase.cs
alistairjevans: Yes, that makes sense; will sort that out tomorrow.

src/Autofac/Core/Resolving/SegmentedStack.cs
alistairjevans: I believe (but don't know all the detail tbh), that reading local variables on the stack can be faster than accessing heap variables multiple times, and can be worth it in these super-hot paths. Local variables can be cached more readily at the CPU, because the compiler understands the entire flow of the variable access (can't be changed by another thread or anything). Like I said, this may be off-base.

I have seen this in some of the internal collection methods in the dotnet runtime though, so it can't be too far off-base.

Fun fact, you'll see the odd uint casts in a couple of places, which allows the jitter to bypass array bounds checking I believe (other trick from the runtime).

test/Autofac.Specification.Test/Features/CircularDependencyTests.cs
alistairjevans: The trace capture here is to actually test that we can generate a trace capture in a call-stack depth exception scenario (to make sure that the remaining stack space detection leaves enough room for any tracing/exception catching.

Might update the asserts though to make that explicit.

src/Autofac/Core/Resolving/ResolveOperationBase.cs
alistairjevans: Possibly yes, although I would need to add a `GlobalCount` to the `SegmentStack`, because `Count` only gives the size of the current segment.

Might be worth it to avoid the extra increment/decrement instructions.

test/Autofac.Specification.Test/Features/CircularDependencyTests.cs
alistairjevans: Not certain where the actual trace content ends up in CI when using `ITestOutputHelper`; it mostly only shows up in the VS Test Explorer.

src/Autofac/Core/Resolving/ResolveOperationBase.cs
alistairjevans: Ah, I remember now why it's tracked separately; property injection that permits circular dependencies will always have an empty request stack, because it runs in Complete, but still needs the depth detection counted separately. Otherwise we get a StackOverflowException.

PR #1154, opened 2020-06-25
Remove calls to GetTypeInfo

alistairjevans:
These changes remove all calls to `GetTypeInfo`, preferring to use the actual methods on `Type`. 

A couple of extension methods on `Type` have been added for consistency with the code that was already there.

The primary motivation around these changes is one of style preference, to conform with the fact that `GetTypeInfo` is a mistake the .net team now regrets, and to eke out a minor performance gain.

There are some very minor performance gains on some of the long benchmarks (like 100ns on DeepGraph and ChildScope), but that's a slight bonus extra only.

### DIFF
--- a/src/Autofac/Builder/RegistrationBuilder.cs
+++ b/src/Autofac/Builder/RegistrationBuilder.cs
@@ -198,7 +198,7 @@ namespace Autofac.Builder
                         continue;
                     }
 
-                    if (!asServiceWithType.ServiceType.GetTypeInfo().IsAssignableFrom(limitType.GetTypeInfo()))
+                    if (!asServiceWithType.ServiceType.IsAssignableFrom(limitType))
                     {
                         throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, RegistrationBuilderResources.ComponentDoesNotSupportService, limitType, ts));
                     }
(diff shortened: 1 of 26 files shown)


tillig:
It might be worth taking a pull from develop into v6 if that hasn't been done recently. At the very least, it'll get the new issue templates, but I think we might have had a couple of other PRs come through since the v6 branch started.

alistairjevans:
I've been doing it pretty regularly, I think it's already up to date; I just pushed those merges directly rather than PRing it.

tillig:
I was thinking about turning on the StyleCop rules we'd talked about but then was like... if it happens in develop, it's going to be merge conflicts in v6; if it happens in v6 it's going to be hard to pull from develop if we need to... meh. Better wait until we've got v6 in develop.

alistairjevans:
Agreed; my gut says that we aren't too far off being ready to merge though. Got some more perf analysis to do, but no major changes left from my perspective. I might go through the issue list and check if there's anything we can pull in.

I think it'd be good to try this version out on some of the integrations; is that easier on develop you think?

Also, documentation updates needed for pipelines and the new composite feature.

tillig:
It's definitely easier to get integration work started once we get things into `develop`; it would mean we can also do the integration work in those `develop` branches rather than having a `v6` for every package. (I guess this is why the .NET Core folks went for the complex build process + monorepo.)

PR #1157, opened 2020-06-30
Split RegistrationExtensions into partial classes

RaymondHuy:


tillig:
It looks like there was maybe a one-click fix or something applied here and a lot of unrelated code got reformatted. Can you undo all but the relevant changes? Thanks!

tillig:
Or maybe your v6 branch hadn't been pulled in a while and you need to rebase?

PR #1159, opened 2020-07-01
Concurrency - Overhaul Registry Locking Mechanism

alistairjevans:
I've been looking how to improve our concurrency performance, and fundamentally it comes down the locks inside `DefaultRegisteredServicesTracker`. 

There's been a number of iterations and attempts to improve this, with varying success. We've also had to roll back some of our more fine-grained locking recently to be more global due to #1073.

This PR is my attempt at improving the concurrency behaviour (which I feel is sort of a rite-of-passage for Autofac contributors).

These changes mean that retrieving an Initialized service's registrations is lock-free.

I know that concurrency like this can be a thorny issue, so it's entirely possible I have missed something, or need additional tests. Please let me know if that is the case.

### Changes Overview

The main design behaviour that caused #1073, and prevents us doing the fine-grained locking, is because we call `CompleteInitialization` on the `ServiceRegistrationInfo` inside recursive calls from registration sources, meaning that an info block indicates it has completed initalisation before the ACTNARS (for example) registration is added to the implementation list.

- Introduced an `InitializationDepth` property to the `ServiceRegistrationInfo`. This value is incremented and decremented manually in `DefaultRegisteredServicesTracker` to capture the knowledge that we are inside a nested registration source call.

- Only when `InitializationDepth` has returned to 0 after processing all the sources will we call `CompleteInitialization`.

- A Monitor.Enter/Exit on the `ServiceRegistrationInfo` protects the section that manipulates InitializationDepth, so only one thread can be recursing through it (or indeed incrementing/decrementing the number). The reason I used Enter/Exit explicitly rather than the lock statement is because I wanted to share the finally block with other code.

- Registration Sources can now access the set of implementations for a service **before** initialization has completed, but **only** if that `InitializationDepth` counter is greater than 0, meaning we are doing initialisation work. Because the counter is protected by the Monitor block, only the current thread can access the implementations in this way. This gives us more safety than just letting anyone access implementations at any time.

- Removed the `Lazy` set of implementations for a `ServiceRegistrationInfo`, and instead use `yield return` in the `Implementations` property to yield the full set of implementations, to ensure that we don't have to regenerate the list if sources need to access implementations during initialization.

### Benchmarks

Concurrency benchmarks are greatly improved:

**develop**
|                          Method | ResolveTaskCount | ResolvesPerTask |     Mean |     Error |    StdDev |    Gen 0 |   Gen 1 | Gen 2 | Allocated |
|-------------------------------- |----------------- |---------------- |---------:|----------:|----------:|---------:|--------:|------:|----------:|
| MultipleResolvesOnMultipleTasks |              100 |             100 | 2.084 ms | 0.0408 ms | 0.0635 ms | 617.1875 | 27.3438 |     - |    4.9 MB |
| NestedScope.MultipleResolvesOnMultipleTasks |                100 |          10 | 1.688 ms | 0.0413 ms | 0.0948 ms | 1101.5625 | 253.9063 |     - |   8.71 MB |

**concurrency**
|                          Method | ResolveTaskCount | ResolvesPerTask |     Mean |    Error |   StdDev |    Gen 0 |   Gen 1 | Gen 2 | Allocated |
|-------------------------------- |----------------- |---------------- |---------:|---------:|---------:|---------:|--------:|------:|----------:|
| MultipleResolvesOnMultipleTasks |              100 |             100 | 593.2 μs | 11.81 μs | 20.68 μs | 977.5391 | 75.1953 |     - |   7.64 MB |
| NestedScope.MultipleResolvesOnMultipleTasks |                100 |          10 | 1.154 ms | 0.0224 ms | 0.0249 ms | 1187.5000 | 267.5781 |     - |   9.44 MB |

### DIFF
--- a/src/Autofac/Core/Registration/DefaultRegisteredServicesTracker.cs
+++ b/src/Autofac/Core/Registration/DefaultRegisteredServicesTracker.cs
@@ -238,11 +238,11 @@ namespace Autofac.Core.Registration
             }
 
             var succeeded = false;
-
-            Monitor.Enter(info);
-
+            var lockTaken = false;
             try
             {
+                Monitor.Enter(info, ref lockTaken);
+
                 if (info.IsInitialized)
                 {
                     return info;
@@ -292,7 +292,10 @@ namespace Autofac.Core.Registration
                     info.CompleteInitialization();
                 }
 
-                Monitor.Exit(info);
+                if (lockTaken)
+                {
+                    Monitor.Exit(info);
+                }
             }
 
             return info;
--- a/src/Autofac/Core/Registration/DefaultRegisteredServicesTracker.cs
+++ b/src/Autofac/Core/Registration/DefaultRegisteredServicesTracker.cs
@@ -24,10 +24,12 @@
 // OTHER DEALINGS IN THE SOFTWARE.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Runtime.CompilerServices;
+using System.Threading;
 using Autofac.Core.Resolving.Pipeline;
 using Autofac.Util;
 
@@ -38,12 +40,14 @@ namespace Autofac.Core.Registration
     /// </summary>
     internal class DefaultRegisteredServicesTracker : Disposable, IRegisteredServicesTracker
     {
+        private static readonly Func<Service, ServiceRegistrationInfo> regInfoFactory = srv => new ServiceRegistrationInfo(srv);
+
         private readonly Func<Service, IEnumerable<ServiceRegistration>> _registrationAccessor;
 
         /// <summary>
         /// Keeps track of the status of registered services.
         /// </summary>
-        private readonly Dictionary<Service, ServiceRegistrationInfo> _serviceInfo = new Dictionary<Service, ServiceRegistrationInfo>();
+        private readonly ConcurrentDictionary<Service, ServiceRegistrationInfo> _serviceInfo = new ConcurrentDictionary<Service, ServiceRegistrationInfo>();
 
         /// <summary>
         /// External registration sources.
@@ -53,14 +57,9 @@ namespace Autofac.Core.Registration
         /// <summary>
         /// All registrations.
         /// </summary>
-        private readonly List<IComponentRegistration> _registrations = new List<IComponentRegistration>();
+        private readonly ConcurrentBag<IComponentRegistration> _registrations = new ConcurrentBag<IComponentRegistration>();
 
         private readonly List<IServiceMiddlewareSource> _servicePipelineSources = new List<IServiceMiddlewareSource>();
-
-        /// <summary>
-        /// Protects instance variables from concurrent access.
-        /// </summary>
-        private readonly object _synchRoot = new object();
 
         /// <summary>
         /// Initializes a new instance of the <see cref="DefaultRegisteredServicesTracker"/> class.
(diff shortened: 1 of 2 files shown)


src/Autofac/Core/Registration/DefaultRegisteredServicesTracker.cs
```diff
@@ -253,52 +233,80 @@ private ServiceRegistrationInfo GetInitializedServiceInfo(Service service)
...
-            if (!info.IsInitializing)
             {
-                BeginServiceInfoInitialization(service, info, _dynamicRegistrationSources);
+                return info;
             }
 
-            while (info.HasSourcesToQuery)
+            var succeeded = false;
+
+            Monitor.Enter(info);
```
dadhi, line 242: @alistairjevans  Hi and sorry for intruding,
The more robust `lock` pattern is using the overload with `Monitor.Enter(info, ref lockTaken)`, see details here http://www.albahari.com/threading/part2.aspx#_MonitorEnter_and_MonitorExit

tillig:
It'd be good to hear from @alexmg if possible, he has messed with this a bunch in the past.

src/Autofac/Core/Registration/DefaultRegisteredServicesTracker.cs
alistairjevans: No worries, thanks for the tip! Change is in.

tillig:
I haven't forgotten this, just haven't had a chance to look yet. Wife and kid are camping with the in-laws so I'm running the 3D printer, CNC router, and my Silhouette Cameo _all at the same time_ making stuff, getting things done that I normally need to do in one-hour increments. 😆

alistairjevans:
No sweat; make the most of this time!

src/Autofac/Core/Registration/DefaultRegisteredServicesTracker.cs
```diff
@@ -199,42 +188,33 @@ public bool IsRegistered(Service service)
...
         {
             if (service == null) throw new ArgumentNullException(nameof(service));
 
-            lock (_synchRoot)
-            {
-                var info = GetInitializedServiceInfo(service);
-                return info.Implementations.ToList();
-            }
+            var info = GetInitializedServiceInfo(service);
+            return info.Implementations.ToList();
```
tillig, line 200: I'm thinking about how `Implementations` is accessed here and in a couple of other spots. Ostensibly once the `ServiceRegistrationInfo` is initialized, that's a fixed list, right? Could `Implementations` be cached as a generated list and avoid the allocation during `RegistrationsFor`?

I may not be totally following things through, so I could be wrong.

tillig: Also, it's OK if that's a separate PR. It doesn't have to block this one.

alistairjevans: Possibly, although the `RegistrationsFor` method is now not really used at runtime (in favour of `ServiceRegistrationsFor`).

I can take a look at the caching of the ServiceRegistration list. but perhaps do that in a separate PR, because it would be another set of fairly substantial changes unrelated to concurrency.

tillig: Sounds good!

alexmg:
Sorry it has taken so long to chime in here. It looks like this addresses the main issue with the nested registration source call and the improved concurrency is going to be a win for server side use. It appears the allocations are up a little but the benchmark would inflate those compared to typical mobile usage scenarios where it is a concern. It's definitely a balancing act, and if I had to pick a direction to tip the scales, it would be towards server side usage in web servers and the like. Nice work as always @alistairjevans! 👏

PR #1163, opened 2020-07-03
feat: build using Ubuntu image

alsami:
* Update `appveyor.yml` to use Ubuntu image and re-order some stuff a bit
* Do not install dotnet-sdk manually on linux

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -165,3 +165,6 @@ $RECYCLE.BIN/
 
 # Mac crap
 .DS_Store
+
+# JetBrains Rider
+.idea
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,13 +1,13 @@
-version: '5.2.0.{build}'
+image: Ubuntu
+
+version: '6.0.0.{build}'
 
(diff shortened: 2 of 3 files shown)


alsami:
@tillig @alistairjevans 

Just tried it out a bit, some stuff breaks because of case-sensitivity now (Build vs build for instance).
Also I see some stuff is redundant and can be simplified when using bash instead of powershell. 

You guys want to continue using powershell or would bash be okay?

tillig:
I think leaving the build as PowerShell has a couple of benefits: it can be used on both Windows or Mac/Linux (PowerShell Core); and we'll still be able to share some of the build script bits with the other builds which may not be entirely Linux-buildable, like the WCF or MVC extensions.

alsami:
Okay great. I really love being able to edit this solution now without the need to put up a virtual machine!

Build should be fine, so ready for review

alsami:
It's actually slower than building on windows. The last successful build on windows took 2min and 40secs, whereas the Ubuntu build took 2min and 57secs.

Issue #1164, opened 2020-07-09
Improve internal object relationship graph for easier testing


tillig:
## Problem Statement

In creating (internal) unit tests for new features you sometimes need to create new instances of things like `RequestContext` (or `RequestContextBase`), `ResolveOperation`, and so on.

Unfortunately...

- The base/abstract classes don't have zero-parameter constructors so you can't use Moq or other mocking frameworks to create them.
- The required constructor parameters themselves are also not easy to mock or set up.

This makes setup for tests somewhat fragile. 

## Desired Solution

I think we should analyze the relationships between things like:

- ResolveOperation (and ResolveOperationBase)
- ResolveRequest
- ResolveRequestContext (and ResolveRequestContextBase)
- Associated interfaces involved in the resolve pipeline

...and see if there can be some cleanup/simplification of the relationships between these things.

## Alternatives You've Considered

As noted, I've tried using Moq to mock these things but they don't have zero-parameter constructors. The only other option is to try to create all of these things.

## Additional Context

Here's what it looks like to try and create a `ResolveOperation`, a `ResolveRequestContext`, and an `IResolveMiddleware` for testing some of the diagnostics:

```c#
using var container = new ContainerBuilder().Build();
using var scope = container.BeginLifetimeScope() as ISharingLifetimeScope;
var operation = new ResolveOperation(scope);
var request = new ResolveRequest(
  new TypedService(typeof(string)),
  new ServiceRegistration(Mock.Of<IResolvePipeline>(), Mock.Of<IComponentRegistration>()),
  Enumerable.Empty<Parameter>());
var context = new ResolveRequestContext(operation, request, scope);
diagnostics.MiddlewareStart(operation, context, Mock.Of<IResolveMiddleware>());
```

I think the `ResolveOperation` construction is reasonable. However, I see that `ResolveRequestContext` requires three things in the constructor and... maybe it needs a zero-parameter constructor and settable properties.

By way of comparison, if you want to test something in ASP.NET Core, [`HttpContext` (the abstract class)](https://github.com/dotnet/aspnetcore/blob/master/src/Http/Http.Abstractions/src/HttpContext.cs) has a zero-parameter constructor so it can be mocked and [`DefaultHttpContext` can be newed-up](https://github.com/dotnet/aspnetcore/blob/master/src/Http/Http/src/DefaultHttpContext.cs) so if you need something concrete without mocking you can do that.

[...]

PR #1165, opened 2020-07-09
chore: small code optimizations

alsami:
* Remove redundant constructor in `ResolveOperationBase` and make `IPipelineTracer` default to `null` 
* Mark some fields as readonly in `ResolveOperationBase`
* Make setters of some properties private, since they are not set from outside
* Make some public properties private, since they are not set from outside
* Use new pattern matching method `x is { }` rather than `x is object`

### DIFF
--- a/src/Autofac/Core/Resolving/Pipeline/ResolveRequestContextBase.cs
+++ b/src/Autofac/Core/Resolving/Pipeline/ResolveRequestContextBase.cs
@@ -162,10 +162,8 @@ namespace Autofac.Core.Resolving.Pipeline
         }
 
         /// <inheritdoc />
-        public virtual object ResolveComponent(ResolveRequest request)
-        {
-            return Operation.GetOrCreateInstance(ActivationScope, request);
-        }
+        public virtual object ResolveComponent(ResolveRequest request) =>
+            Operation.GetOrCreateInstance(ActivationScope, request);
 
         /// <summary>
         /// Complete the request, raising any appropriate events.
--- a/src/Autofac/Core/Resolving/ResolveOperationBase.cs
+++ b/src/Autofac/Core/Resolving/ResolveOperationBase.cs
@@ -39,8 +39,8 @@ namespace Autofac.Core.Resolving
         private const int SuccessListInitialCapacity = 32;
 
         private bool _ended;
-        private IResolvePipelineTracer? _pipelineTracer;
-        private List<ResolveRequestContext> _successfulRequests = new List<ResolveRequestContext>(SuccessListInitialCapacity);
+        private readonly IResolvePipelineTracer? _pipelineTracer;
+        private readonly List<ResolveRequestContext> _successfulRequests = new List<ResolveRequestContext>(SuccessListInitialCapacity);
         private int _nextCompleteSuccessfulRequestStartPos = 0;
 
         /// <summary>
@@ -48,23 +48,13 @@ namespace Autofac.Core.Resolving
         /// </summary>
         /// <param name="mostNestedLifetimeScope">The most nested scope in which to begin the operation. The operation
         /// can move upward to less nested scopes as components with wider sharing scopes are activated.</param>
-        protected ResolveOperationBase(ISharingLifetimeScope mostNestedLifetimeScope)
-            : this(mostNestedLifetimeScope, null)
-        {
-        }
-
-        /// <summary>
-        /// Initializes a new instance of the <see cref="ResolveOperationBase"/> class.
-        /// </summary>
-        /// <param name="mostNestedLifetimeScope">The most nested scope in which to begin the operation. The operation
-        /// can move upward to less nested scopes as components with wider sharing scopes are activated.</param>
         /// <param name="pipelineTracer">A pipeline tracer for the operation.</param>
-        protected ResolveOperationBase(ISharingLifetimeScope mostNestedLifetimeScope, IResolvePipelineTracer? pipelineTracer)
+        protected ResolveOperationBase(ISharingLifetimeScope mostNestedLifetimeScope, IResolvePipelineTracer? pipelineTracer = null)
         {
+            _pipelineTracer = pipelineTracer;
             TracingId = this;
             IsTopLevelOperation = true;
             CurrentScope = mostNestedLifetimeScope;
-            _pipelineTracer = pipelineTracer;
             IsTopLevelOperation = true;
         }
 
@@ -81,6 +71,15 @@ namespace Autofac.Core.Resolving
             TracingId = tracingId;
             IsTopLevelOperation = false;
         }
+
+        /// <summary>
+        /// Gets the modifiable active request stack.
+        /// </summary>
+        /// <remarks>
+        /// Don't want this exposed to the outside world, but we do want it available in the <see cref="CircularDependencyDetectorMiddleware"/>,
+        /// hence it's internal.
+        /// </remarks>
+        internal SegmentedStack<ResolveRequestContextBase> RequestStack { get; } = new SegmentedStack<ResolveRequestContextBase>();
 
         /// <summary>
         /// Gets the active resolve request.
@@ -103,9 +102,9 @@ namespace Autofac.Core.Resolving
         public ITracingIdentifer TracingId { get; }
 
         /// <summary>
-        /// Gets or sets the current request depth.
+        /// Gets the current request depth.
         /// </summary>
-        public int RequestDepth { get; protected set; }
+        public int RequestDepth { get; private set; }
 
         /// <summary>
         /// Gets a value indicating whether this operation is a top-level operation (as opposed to one initiated from inside an existing operation).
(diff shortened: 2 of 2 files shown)


src/Autofac/Core/Resolving/Pipeline/ResolveRequestContextBase.cs
```diff
@@ -59,7 +59,7 @@ public abstract class ResolveRequestContextBase : IComponentContext
             Parameters = request.Parameters;
             PhaseReached = PipelinePhase.ResolveRequestStart;
             Tracer = tracer;
-            TracingEnabled = tracer is object;
```
alsami, line 62: @alistairjevans I remember that you've introduced this here for some performance reason. Is there a specific thing that I might miss, on why you have used `is object` instead of `tracer != null`?

tillig: Tracer is changing a lot in my current work, switching to `System.Diagnostics.DiagnosticSource`. I wouldn't worry about this too much.

tillig: (But I don't think there's a difference; `is object` is readability syntactic sugar, I believe.)

alsami: Okay cool.

alistairjevans: Storing the bool `TracingEnabled` is an optimisation to prevent checking it repeatedly. 

The `is object` is a test for non-null that does not look at custom operators. I generally prefer it now that pattern matching performance is optimised for that case.

https://stackoverflow.com/questions/58505313/proper-null-check-on-service-result-in-c-sharp-net/58506420#58506420

alsami: So basically `!(x is null)` would generate less IL code as well? Anyway, adding it back!

alsami: Apparently this is also possible now
`TracingEnabled = tracer is { };`

:man_shrugging:

alsami:
Good to go @tillig @alistairjevans

src/Autofac/Core/Resolving/Pipeline/ResolveRequestContextBase.cs
alistairjevans: Yeah, pattern matching is kinda crazy.

I think the `is object` and `is null` terms have specific JIT optimisation to make them as fast as the original null comparison. Not sure about the `is {}`.

src/Autofac/Core/Resolving/Pipeline/ResolveRequestContextBase.cs
```diff
@@ -59,7 +59,7 @@ public abstract class ResolveRequestContextBase : IComponentContext
...
+            TracingEnabled = tracer is { };
```
alistairjevans, line 62: I feel like `is { }` is semantically more confusing than `is object`, so I'd  prefer to keep the original.

That syntax is really intended for checking for specific properties on the type, rather than just a null check.

alsami: Okay, reverted that change in both places.

alistairjevans:
General comment; some of the public properties here are public because I thought they'd be useful in Autofac extensions/integrations, even if they're  not used in Autofac.

alsami:
> General comment; some of the public properties here are public because I thought they'd be useful in Autofac extensions/integrations, even if they're not used in Autofac.

Ah okay, mind stating which? I'd add some dummy tests referencing those, so it's easier to understand which were intentionally exposed.

alistairjevans:
That's a good point.

I'll take a proper look and let you know.

src/Autofac/Core/Resolving/ResolveOperationBase.cs
```diff
@@ -83,48 +73,43 @@ protected ResolveOperationBase(ISharingLifetimeScope mostNestedLifetimeScope, IR
         }
 
         /// <summary>
-        /// Gets the active resolve request.
+        /// Gets or sets the active resolve request.
         /// </summary>
-        public ResolveRequestContextBase? ActiveRequestContext { get; private set; }
```
alistairjevans, line 88: Could be useful for any code that only has access to the operation to know which Resolve is currently running. I'm picturing more monitoring/diagnostic extensions as much as anything else.

src/Autofac/Core/Resolving/ResolveOperationBase.cs
```diff
@@ -83,48 +73,43 @@ protected ResolveOperationBase(ISharingLifetimeScope mostNestedLifetimeScope, IR
...
+        private ResolveRequestContextBase? ActiveRequestContext { get; set; }
 
         /// <summary>
-        /// Gets the current lifetime scope of the operation; based on the most recently executed request.
+        /// Gets or sets the current lifetime scope of the operation; based on the most recently executed request.
         /// </summary>
-        public ISharingLifetimeScope CurrentScope { get; private set; }
```
alistairjevans, line 93: The `CurrentScope` of the `ResolveOperation` is technically different from the `ActivationScope` of a `ResolveRequestContext`, because it represents the scope at the **start** of the Resolve, rather than one selected by the `ScopeSelectionMiddleware` during the resolve.

src/Autofac/Core/Resolving/ResolveOperationBase.cs
```diff
@@ -83,48 +73,43 @@ protected ResolveOperationBase(ISharingLifetimeScope mostNestedLifetimeScope, IR
...
+        private ISharingLifetimeScope CurrentScope { get; set; }
 
         /// <summary>
-        /// Gets the set of all in-progress requests on the request stack.
+        /// Gets the modifiable active request stack.
         /// </summary>
-        public IEnumerable<ResolveRequestContextBase> InProgressRequests => RequestStack;
```
alistairjevans, line 98: If an extension wanted to know the in-progress requests to do anything different with the Circular Dependency Analysis, this is how they get a read-only view of the current request stack.

src/Autofac/Core/Resolving/ResolveOperationBase.cs
```diff
@@ -39,32 +39,22 @@ public abstract class ResolveOperationBase : IResolveOperation, ITracingIdentife
         private const int SuccessListInitialCapacity = 32;
 
         private bool _ended;
-        private IResolvePipelineTracer? _pipelineTracer;
-        private List<ResolveRequestContext> _successfulRequests = new List<ResolveRequestContext>(SuccessListInitialCapacity);
+        private readonly IResolvePipelineTracer? _pipelineTracer;
+        private readonly List<ResolveRequestContext> _successfulRequests = new List<ResolveRequestContext>(SuccessListInitialCapacity);
         private int _nextCompleteSuccessfulRequestStartPos = 0;
 
-        /// <summary>
```
alistairjevans, line 46: I have mixed feelings about the removal of the constructor that has no tracer. I get that you can just do the optional parameter, but I also like the semantics of there being two different 'modes' of tracing.

I think this may change with @tillig's DiagnosticSource changes anyway, so we can revisit that when those changes are ready.

tillig: With DiagnosticSource, it's sort of a global singleton at the container level. It gets passed down from lifetime scope to lifetime scope, but it's not as flexible as having a tracer that only sits at a single scope level. On the other hand, it's super efficient and you can subscribe to individual events pretty easy for things like perf counters if you want.

Net: the constructor parameter with the tracer goes away and the DiagnosticSource on the scope gets used. I wouldn't worry about messing with this too much.

alsami: What's the advantage of the constructor that I've removed, when it also went and called `this(mostNestedLifetimeScope, null)` anyway?

alsami:
> I like the idea of adding simple tests to assert these properties are working, but I think they're better off being available.

I could add them with this PR if you don't mind.

src/Autofac/Core/Resolving/ResolveOperationBase.cs
alistairjevans: It's mostly just a mild style preference; it's sometimes nicer to just have separate constructors, especially if you might want different summary comments (which I didn't actually avail myself of here).

I don't feel strongly about it either way, and it sounds like it'll be changed anyway with the DiagnosticSource updates.

alistairjevans:
If you're happy to, that would be great @alsami, thanks.

Or if you see anything else that you don't think has enough coverage! 😀

alsami:
> Or if you see anything else that you don't think has enough coverage!

For this PR I'd just create a dummy derived type, that would access given properties and add some comment, for the next the dev to not think "oh, this is unused, let's remove it". :laughing:

alistairjevans:
👍
Might also be a good experiment in figuring out easier unit testing of these structures, as per #1164.

alsami:
Closing this. Gonna expirement with tests for now.

Issue #788, opened 2016-08-19
Feature request: dot-graph (graphviz) generation, from a running Autofac container


ph-hs:
(co creator of PicoContainer here)

Obviously as a optional plugin for Autofac, could there be a means to instrument injections. Only produce the dotfile output, as the user can run that through the executables to make pics.

Ref - http://picocontainer.com/Old_JIRA_Issues/PICO/285/ and https://github.com/picocontainer/PicoContainer2/blob/master/pico/gems/src/java/org/picocontainer/gems/monitors/DotDependencyGraphComponentMonitor.java

You would not run this in a production system, unless you were intending to make some diagram that that showed exhaustive injections for the system, and counts for each. More likely you'd have a Selenium test suite for the whole app, and extract the dot diagram before tearing down the server.

tillig:
Instrumentation is definitely an area in which we need to improve. We've tried previously to add text based logs or other similarly simple diagnostics with some limited success. An actual graphic view would be pretty cool.

Looks like it may be time to revive some of the instrumentation and perf efforts now that the .NET Core madness is mostly over.

ph-hs:
With PicoContainer we had a a 'ComponentMonitor' (default NullComponentMonitor) :- https://github.com/picocontainer/PicoContainer2/blob/master/pico/container/src/java/org/picocontainer/ComponentMonitor.java

Methods: instantiating, instantiated, instantiationFailed, invoking (for lifecycle: start/stop/dispose), invoked, invocationFailed, lifecycleInvocationFailed, noComponentFound, newInjector, newBehavior.

tillig:
A lot of time has passed on this; I'm curious if we could make use of `DiagnosticSource` now to raise events that could be handled by instrumentation.

tillig:
I was thinking about picking this up now that v6 has some good instrumentation. I'm curious about how we would expect it to work. I figure there are two logical ways:

1. The graph is built for the whole container. The graph generator keeps an abbreviated in-memory tree of what depends on what for the entire run of the container. The "root" of the tree is the container itself. The graph is rendered on dispose of the container or on manual execution of some sort of "render" method.
2. One graph is built per resolve operation. Similar to the existing trace mechanism, where each trace scope is a single resolve through the complete dependency chain, that's what happens here, just with a dot-graph format.

I think the first option is _interesting_ but not _valuable_. The amount of data in that graph would likely be overwhelming and confusing, plus it gets really complicated when you get into things like per-request and named lifetime scopes.

[...]

PR #1169, opened 2020-07-17
Switch pipeline trace to DiagnosticSource

tillig:
This PR is part of #788 and is the first step in getting the Autofac.Diagnostics.DotGraph package out.

It updates the diagnostic mechanism to use `System.Diagnostics.DiagnosticSource` to emit events and moves the diagnostics classes into an `Autofac.Diagnostics` namespace.

Subscribing to events in this system means creating a `System.Diagnostics.DiagnosticListener` and attaching it to the container's `DiagnosticSource`. Internally this follows an observer pattern that receives a stream of events, keyed by string IDs, with which the listener may do... whatever they want. There are convenience methods for subscribing to the container's diagnostics, like:

```c#
container.SubscribeToDiagnostics<MyListener>();
```

The default diagnostic tracer was updated to be a diagnostic listener in this format. Rather than paste an example in here, check that out.

I tried to follow patterns I'd seen in areas like ASP.NET Core with respect to things like:

- Checking to see if there's a listener subscribed prior to emitting an event.
- Keeping some things internal, like the strings that are the list of event IDs.

That latter one was sort of interesting. I wondered why it was internal but realized a couple of things:

1. Constants are faster because they get compiled into your code, but... they get compiled into your code. So changing the constant later doesn't actually take effect in consuming code outside the assembly. You'd have to use static readonly, but that's not as fast access-wise and this needs to be super low overhead.
2. Adding, removing, or changing events means adding, removing, or changing these event IDs. If there's not a public constants class, it means we can do that pretty freely (within reason) and the most "breaking" it will be for a consumer is that they won't get that event anymore. If it's part of the public API, it actually breaks in a more fatal way, and that's not cool.

The history is pretty messy because I started with the DOT graph tracer in this PR and then it moved out to a separate package. I tried to squash/rebase it but that, too, got kinda dicey since I'd been pulling changes from the v6 branch and... yeah.


tillig:
A few extra things did sneak in here:

- The .editorconfig got updated based on styles we're hoping to make uniform. They do provide some "info" level suggestions in VS Code; we'll want to look at fixing some of that up once v6 is in develop.
- We have a .github repo now with things like the bug/issue templates, the code of conduct, and so on. We need to remove them from here to allow that central repo to work.
- I found a couple of PSScriptAnalyzer warnings I fixed in the build.ps1 script.
- The `MetadataKeys` file had a TODO that'd been in there for quite some time and was actually obsolete. I removed the associated "TODO: delete this stuff" stuff. I did that because I was trying to use TODO to track the stuff I was working on and... my OCD took over.

I didn't run the perf metrics against it to see if things changed, though I can't imagine it's significant enough to make a huge impact. I've been wrong.

One thing that _is_ different is that the diagnostic source is shared from the container down. You attach at the container level, not at any arbitrary lifetime scope level. I'm not sure if this reduction in functionality matters too much, but it's important to be aware - that's different than the pipeline tracer approach from before.

src/Autofac/Diagnostics/DiagnosticTracerBase.cs
```diff
@@ -0,0 +1,367 @@
...
+    /// to Autofac.
+    /// </para>
+    /// </remarks>
+    /// <seealso cref="DefaultDiagnosticTracer"/>
+    public abstract class DiagnosticTracerBase : IObserver<KeyValuePair<string, object>>
+    {
+        /// <summary>
+        /// The list of event names to which this observer is subscribed.
+        /// </summary>
+        private readonly List<string> _subscriptions = new List<string>();
```
alistairjevans, line 36: Just thought I'd mention that I decided to benchmark List vs HashSet for the size of set we're working with here, just to check; I kinda assumed that a list would be faster for small sets, which it sort of is when searching the beginning of the List, but not the end.

List is wayyy faster when it's empty though, which is our most common case.

|             Method |       Mean |     Error |    StdDev |
|------------------- |-----------:|----------:|----------:|
|          ListEmpty |  0.3399 ns | 0.0134 ns | 0.0119 ns |
|     ListFirstAdded | 13.8114 ns | 0.1138 ns | 0.1008 ns |
|    ListMiddleAdded | 26.4125 ns | 0.1709 ns | 0.1598 ns |
|      ListLastAdded | 53.4440 ns | 0.2396 ns | 0.2242 ns |
|       HashSetEmpty |  2.3426 ns | 0.0317 ns | 0.0265 ns |
|  HashSetFirstAdded | 38.4255 ns | 0.2956 ns | 0.2765 ns |
| HashSetMiddleAdded | 34.3019 ns | 0.1730 ns | 0.1350 ns |
|   HashSetLastAdded | 34.3787 ns | 0.1868 ns | 0.1656 ns |

src/Autofac/Diagnostics/DiagnosticTracerBase.cs
```diff
@@ -0,0 +1,367 @@
...
+        /// opt-in to any events you wish to handle. This method helps you
+        /// to quickly subscribe to all the Autofac events at once. You may use
+        /// <see cref="Enable"/> to subscribe to individual events.
+        /// </para>
+        /// </remarks>
+        /// <seealso cref="Enable"/>
+        /// <seealso cref="Disable"/>
+        public void EnableAll()
+        {
+            _subscriptions.Add(DiagnosticEventKeys.MiddlewareStart);
```
alistairjevans, line 82: For a micro-perf gain while capturing diagnostics, I'd suggest re-ordering these events so they are ordered by 'most frequently invoked'.

So I'd put all the failures at the end, then put the request items before the operation ones.

src/Autofac/Diagnostics/DiagnosticTracerBase.cs
```diff
@@ -0,0 +1,367 @@
...
+        /// <param name="data">
+        /// Diagnostic data associated with the event.
+        /// </param>
+        /// <remarks>
+        /// <para>
+        /// Derived classes can override this method and perform actions based
+        /// on the event. By default, the base class does nothing.
+        /// </para>
+        /// </remarks>
+        public virtual void OnMiddlewareFailure(MiddlewareDiagnosticData data)
```
alistairjevans, line 183: It feels like all of the `On` methods should be protected, because they don't mean anything to someone that news up a `DefaultDiagnosticTracer` (and could actually cause some interesting errors).

src/Autofac/Core/Lifetime/LifetimeScope.cs
```diff
@@ -101,10 +91,12 @@ protected LifetimeScope(IComponentRegistry componentRegistry, LifetimeScope pare
         /// <param name="componentRegistry">Components used in the scope.</param>
         public LifetimeScope(IComponentRegistry componentRegistry, object tag)
         {
-            _sharedInstances[SelfRegistrationId] = this;
             ComponentRegistry = componentRegistry ?? throw new ArgumentNullException(nameof(componentRegistry));
             Tag = tag ?? throw new ArgumentNullException(nameof(tag));
+
+            _sharedInstances[SelfRegistrationId] = this;
             RootLifetimeScope = this;
+            this.DiagnosticSource = new DiagnosticListener("Autofac");
```
alistairjevans, line 99: Remaining `this.`

src/Autofac/Core/Lifetime/LifetimeScope.cs
```diff
@@ -75,23 +74,14 @@ public class LifetimeScope : Disposable, ISharingLifetimeScope, IServiceProvider
...
-            : this(componentRegistry, tag)
-        {
-            _tracer = tracer;
+            ComponentRegistry = componentRegistry ?? throw new ArgumentNullException(nameof(componentRegistry));
+            Tag = tag ?? throw new ArgumentNullException(nameof(tag));
             _parentScope = parent ?? throw new ArgumentNullException(nameof(parent));
+
+            _sharedInstances[SelfRegistrationId] = this;
             RootLifetimeScope = _parentScope.RootLifetimeScope;
+            this.DiagnosticSource = _parentScope.DiagnosticSource;
```
alistairjevans, line 84: Remaining `this.`

src/Autofac/Autofac.csproj
```diff
@@ -57,6 +57,7 @@
     <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.113">
       <PrivateAssets>All</PrivateAssets>
     </PackageReference>
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="4.7.1" />
```
alistairjevans, line 60: Incredibly small nit based on personal preference, can we put the 'public' package reference before the private (analyser) ones? Maybe even in a separate ItemGroup?

I'm so used to using the csproj directly these days it would be nice to group them.

src/Autofac/Core/Container.cs
```diff
@@ -100,10 +100,7 @@ public ILifetimeScope BeginLifetimeScope(object tag, Action<ContainerBuilder> co
         }
 
         /// <inheritdoc/>
-        public void AttachTrace(IResolvePipelineTracer tracer)
-        {
-            _rootLifetimeScope.AttachTrace(tracer);
-        }
+        public DiagnosticListener DiagnosticSource => this._rootLifetimeScope.DiagnosticSource;
```
alistairjevans, line 103: The `DiagnosticListener` implements `IDisposable`. Should we dispose it when the `Container` is disposed?

src/Autofac/Core/Resolving/ResolveOperationBase.cs
```diff
@@ -212,7 +196,10 @@ public object GetOrCreateInstance(ISharingLifetimeScope currentOperationScope, R
 
             try
             {
-                _pipelineTracer?.RequestStart(this, requestContext);
+                if (DiagnosticSource.RequestDiagnosticsEnabled())
```
alistairjevans, line 199: Suggest saving this value as a boolean to avoid the additional test when the request succeeds/fails.

src/Autofac/Core/Resolving/ResolveOperationBase.cs
```diff
@@ -155,26 +122,40 @@ protected object ExecuteOperation(ResolveRequest request)
             {
                 InitiatingRequest = request;
 
-                _pipelineTracer?.OperationStart(this, request);
+                if (DiagnosticSource.OperationDiagnosticsEnabled())
```
alistairjevans, line 125: Suggest saving this as a boolean so we don't need to re-run the check at the end of the operation.

src/Autofac/ILifetimeScope.cs
```diff
@@ -119,15 +119,10 @@ public interface ILifetimeScope : IComponentContext, IDisposable, IAsyncDisposab
...
+        /// Gets the <see cref="System.Diagnostics.DiagnosticListener"/> to which
+        /// trace events should be written.
         /// </summary>
-        /// <param name="tracer">The <see cref="IResolvePipelineTracer"/> implementation.</param>
-        /// <remarks>
-        /// Only one tracer is supported at once, so calling this will replace any prior tracer on this scope. Nested scopes
-        /// will continue to retain the same trace behaviour.
-        /// </remarks>
-        void AttachTrace(IResolvePipelineTracer tracer);
+        DiagnosticListener DiagnosticSource { get; }
```
alistairjevans, line 125: If someone subscribes to events on the `DiagnosticListener` on ILifetime Scope rather than the Container, they might expect to only capture events for that scope. I know we aren't going to support that pattern, but perhaps we should enforce it by only making the `DiagnosticListener` publicly available on `IContainer` with the `DiagnosticSource` only cropping back up when we start executing resolves?

I can't picture a scenario where someone would need access to write messages to **our** diagnostic source outside of a pipeline.

src/Autofac/Diagnostics/DiagnosticTracerBase.cs
```diff
@@ -0,0 +1,367 @@
...
+        /// <para>
+        /// By default the observer is not subscribed to any events. You must
+        /// opt-in to any events you wish to handle. You may use
+        /// <see cref="EnableAll"/> to subscribe to all Autofac events at once,
+        /// or <see cref="Enable"/> to subscribe to individual events.
+        /// </para>
+        /// </remarks>
+        /// <seealso cref="EnableAll"/>
+        /// <seealso cref="Enable"/>
+        public void Disable(string diagnosticName)
```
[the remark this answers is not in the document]
tillig: I had them virtual, but you shouldn't call virtual methods from a constructor due to issues with order of constructor execution and how it could lead to calling the method in a class not fully constructed.

I could separate Enable and Disable into public virtual and protected non-virtual methods such that the ctor and default public methods call the protected methods.

src/Autofac/Diagnostics/DiagnosticTracerBase.cs
```diff
@@ -0,0 +1,367 @@
...
+        /// strongly-typed object that can be handled by the other methods
+        /// in this observer.
+        /// </summary>
+        /// <param name="diagnosticName">
+        /// The name of the event that was raised. Diagnostic names are case-sensitive.
+        /// </param>
+        /// <param name="data">
+        /// The diagnostic data associated with the event.
+        /// </param>
+        public virtual void Write(string diagnosticName, object data)
```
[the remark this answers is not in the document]
tillig: Torn on that because it isn't one of the Autofac events, it's sort of a pseudo-event. Adding "On" makes it seem like it's something that'd be raised by middleware or an operation. {I stole "Write" from the ASP.NET Core mechanism.)

src/Autofac/ILifetimeScope.cs
tillig: I'll see what I can do here. If I recall correctly, in some cases we only pass the interface around and I need access to the parent source to hand it down to the child. Which is to say, this isn't necessarily only for the consumer; I think I need it to make the hand-off work. But I'll look.

src/Autofac/Autofac.csproj
tillig: Conversely, I find it easier to locate packages when they're sorted in alpha order by package name. That's also how `dotnet add` adds the references so there's no fighting the tool. Do you feel that strongly about it?

src/Autofac/Core/Resolving/Pipeline/ResolvePipelineBuilder.cs
```diff
@@ -271,10 +270,10 @@ Action<ResolveRequestContextBase> Chain(Action<ResolveRequestContextBase> next,
 
                 return (ctxt) =>
                 {
-                    // Optimise the path depending on whether a tracer is attached.
-                    if (ctxt.TracingEnabled)
+                    // Optimise the path depending on whether diagnostics are enabled.
+                    if (ctxt.DiagnosticSource.MiddlewareDiagnosticsEnabled())
```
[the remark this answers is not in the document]
tillig: This is tough because the thing being checked is looking to see if any subscribers have this enabled, right? And there could be multiple subscribers, which could be added at any time.

I could simplify the checks - check for literally the exact event being subscribed instead of one of a logical set. Right now I check for start or success or fail to determine if any subscribers exist, but in the majority case I'm guessing that's three "false" checks every time. I could switch it to check only for the specific event being issued. It'll change the flow a bit since you could subscribe to a failure event without subscribing to start or success, but maybe that's okay.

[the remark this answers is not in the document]
alistairjevans: I think it's probably justified for performance reasons to say "If you change diagnostic subscriptions to middleware events, that change  may not take effect until the next resolve operation.

Especially since the 99% use case is going to be turning it on once, for all events, and never changing it.

src/Autofac/Diagnostics/DiagnosticTracerBase.cs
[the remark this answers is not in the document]
alistairjevans: Understood, don't feel strongly about it. This becomes less of an issue if Write became protected.

src/Autofac/Autofac.csproj
alistairjevans: I don't feel strongly about it at all, like I said, personal preference.

src/Autofac/Core/Resolving/Pipeline/ResolvePipelineBuilder.cs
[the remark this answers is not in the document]
tillig: So evaluate all the subscriptions at the operation level and pass that all down through the stack? I think that might add some complexity. Let me see about making the checks more precise and see where that leaves us. I also have a not-totally-formed idea about using flags internally to do bitwise checks and speed up comparison.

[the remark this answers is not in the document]
alistairjevans: Honestly I was just thinking that if someone changes the set of suscribers or subscribed events such that our checking method returns a different value, we...might not raise some middleware events during that transition. But then if someone did change that set during a request, they would have missed related operation and request messages anyway, which could introduce different problems.

Is this sort of thing what the DiagnosticSource `Activity` functionality is for?

Fundamentally it's whatever gets us the perf saving with no subscribers, so whatever approach works.

Are we anticipating encouraging the use of diagnostics in production? If so, do we need to bench the diagnostic source capture?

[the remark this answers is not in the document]
tillig: I thought about using Activity but it didn't seem worth the allocations.

I'm less concerned about changing subscriptions in the middle of the request than I am about trying to be too smart in the caching mechanism such that we stop responding correctly to observables or it becomes hard to add a new event in here. I could see wanting diagnostic events for all the regular lifetime events, too.

I don't think diagnostics in prod would be recommended. More like "turn this on if you find a problem and turn it back off" or something.

src/Autofac/Core/Lifetime/LifetimeScope.cs
tillig: Removed all `this.` from the src/** set of files.

src/Autofac/Core/Lifetime/LifetimeScope.cs
tillig: Removed all `this.` from the src/** set of files.

src/Autofac/Autofac.csproj
tillig: Sounds good, I'll leave this one, then.

tillig:
First push today fixes the `this` references. Diving into other stuff now.

src/Autofac/Diagnostics/DiagnosticTracerBase.cs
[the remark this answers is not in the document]
tillig: I split these into `public virtual Enable` and `protected EnableBase` such that constructors and things can call the non-virtual versions but we can override them to "disable" subscription management. I did that on the default tracer - it now throws a `NotSupportedException` if you try to `Enable` or `Disable` a subscription.

src/Autofac/Diagnostics/DiagnosticTracerBase.cs
tillig: I reordered these by:
- Middleware events (most often - start, then success)
- Request events (second most often - start, then success)
- Operation events (third most often - start, then success)
- Failures by most to least frequent (middleware, request, operation)

Hopefully that helps a bit.

src/Autofac/Diagnostics/DiagnosticTracerBase.cs
tillig: All the trace event `OnWhatever` methods as well as `Write` are now protected.

src/Autofac/ILifetimeScope.cs
tillig: Looks like I can add some new constructor parameters to things like `ResolveOperationBase` to allow me to pass in the diagnostic source and this should work.

tillig: Pushed changes such that only `IContainer` has a diagnostic source property. Everything else is internal. The source gets propagated through constructor parameters as needed.

src/Autofac/Core/Container.cs
tillig: Updated the `LifetimeScope` constructor such that if it is the thing creating the `DiagnosticSource` it also adds the source to its `DisposeTracker` so when the scope is disposed, the diagnostic source will also be disposed.

src/Autofac/Diagnostics/DiagnosticTracerBase.cs
```diff
@@ -0,0 +1,367 @@
...
+                    OnRequestFailure((RequestFailureDiagnosticData)data);
+                    break;
+                case DiagnosticEventKeys.RequestStart:
+                    OnRequestStart((RequestDiagnosticData)data);
+                    break;
+                case DiagnosticEventKeys.RequestSuccess:
+                    OnRequestSuccess((RequestDiagnosticData)data);
+                    break;
+                default:
+                    break;
```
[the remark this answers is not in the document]
tillig: Added a raft of commentary to the top of the class and a reference to that doc in the `Write` documentation.

src/Autofac/Core/Resolving/Pipeline/ResolveRequestContextBase.cs
```diff
@@ -47,19 +48,16 @@ public abstract class ResolveRequestContextBase : IComponentContext
...
-            IResolvePipelineTracer? tracer)
+            ISharingLifetimeScope scope)
         {
             Operation = owningOperation;
             ActivationScope = scope;
             Parameters = request.Parameters;
             PhaseReached = PipelinePhase.ResolveRequestStart;
-            Tracer = tracer;
-            TracingEnabled = tracer is object;
+            this.DiagnosticSource = scope.DiagnosticSource;
```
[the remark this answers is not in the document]
tillig: Removed all `this.` from the src/** set of files.

tillig:
A few more pushes address

- `this` usage - removed all the `this` from the src/** files
- documentation updates
- switching things so only `IContainer` exposes `DiagnosticSource`
- disposal of `DiagnosticSource` when the container is disposed
- updating the event handling methods of the base tracer to be protected

src/Autofac/Diagnostics/DiagnosticTracerBase.cs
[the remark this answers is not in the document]
tillig: Left it `Write`, made it protected.

src/Autofac/Diagnostics/OperationDiagnosticTracerBase.cs
```diff
@@ -0,0 +1,75 @@
...
+    /// Base class for tracers that require all operations for logical operation tracing.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// Derived classes will be subscribed to all Autofac diagnostic events
+    /// and will raise an <see cref="OperationDiagnosticTracerBase.OperationCompleted"/>
+    /// event when a logical operation has finished and trace data is available.
+    /// </para>
+    /// </remarks>
+    public abstract class OperationDiagnosticTracerBase : DiagnosticTracerBase
```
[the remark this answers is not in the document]
tillig: Updated to be generic.

src/Autofac/Core/Resolving/Pipeline/ResolvePipelineBuilder.cs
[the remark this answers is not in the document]
tillig: Realized now that I was basically checking enabled _twice_ for any given write.

First time in `if(source.MiddlewareDiagnosticsEnabled())` and later in the actual call to write the event. Same for operation and request.

I've removed the aggregate checks and just leave it to the call to write the event. This should remove some of the overhead. **Let's see where this puts us perf-wise and I can try to tweak some more if it looks like it still needs help.**

I'm trying to not cart too much additional context around if possible, and while a single Boolean isn't that big a deal, the orchestration of keeping the cache up to date with the actual settings seems like it could be a challenge long-term, similar to how container rebuilding got to be a bit sticky and unpredictable.

I can see where there is a use case for success or failure instrumentation, like perf counters or something, to track stats. That'd mean subscribing to just success/failure events but maybe not start events. The ability to have that granularity seems interesting to me, rather than just "all on" or "all off" for a whole request.

I can also see that we may want to support (at some point) the [ability to conditionally determine if you're 'enabled' based on the data itself](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.diagnosticlistener.isenabled?view=netcore-3.1#System_Diagnostics_DiagnosticListener_IsEnabled_System_String_System_Object_System_Object_) which wouldn't be possible if we try to pre-cache things too much. Like, "I want to see events that involve resolving a certain type" to diagnose an issue would be a use case for being enabled based on certain data.

src/Autofac/Diagnostics/DiagnosticTracerBase.cs
tillig: Hmmm, interesting. Keep in mind it's the `DiagnosticListener` / event provider that determines if `IsEnabled` comes out right by iterating across all subscribers (our tracers).

Which is to say, either there _aren't any subscribers_, so it's never hitting our implementation of `IsEnabled` anyway; or there _are subscribers_ (one of our tracers? some other `IObservable<KVP<>>` implementation?) and the subscribers are likely subscribed to at least one event.

I don't think the majority case for our tracers actually will be that they're attached but have no subscriptions. I do think the majority case in general is that the container's `DiagnosticListener` instance won't have any subscribers.

I did [just find that `DiagnosticListener` can do a shortcut call with `IsEnabled()`](https://github.com/dotnet/runtime/blob/e0d20d37592d52d672ff89d5838d0e77e7ee580e/src/libraries/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/DiagnosticListener.cs#L225) such that if there aren't any subscribers then it's a fast null check. Let me add that to all the events, should speed things up.

tillig: Just pushed that. Benchmark the crap out if it, it should still basically boil down to a check for `subscriptions != null` before doing anything, which should be just as fast as what we had before. Granted, it's going to do that null check for each event rather than just one time, but that still should be pretty fast.

src/Autofac/Core/Resolving/Pipeline/ResolvePipelineBuilder.cs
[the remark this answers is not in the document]
tillig: Just pushed an update to have the event source `DiagnosticListener` check for `IsEnabled()` before checking if the specific named subscription is enabled. Benchmark the crap out if it, it should still basically boil down to a check for subscriptions != null before doing anything, which should be just as fast as what we had before. Granted, it's going to do that null check for each event rather than just one time, but that still should be pretty fast.

src/Autofac/Core/Resolving/ResolveOperationBase.cs
tillig: Each diagnostic write now does `if(DiagnosticSource.IsEnabled() && DiagnosticSource.IsEnabled("eventname"))` before writing, which is basically a check to see if there are _any_ subscribers (a null check) prior to checking if a specific event is enabled. This should be roughly as fast as the null check we had before.

src/Autofac/Core/Resolving/ResolveOperationBase.cs
tillig: Each diagnostic write now does `if(DiagnosticSource.IsEnabled() && DiagnosticSource.IsEnabled("eventname"))` before writing, which is basically a check to see if there are _any_ subscribers (a null check) prior to checking if a specific event is enabled. This should be roughly as fast as the null check we had before.

tillig:
Pushed an update to the way things get checked for subscription enablement to try to speed that up.

I think it's ready for another review and benchmarking.

alsami:
> I think it's ready for another review and benchmarking.

Just ran the Benchmarks on Linux, hope that's fine.

|  DeepGraphResolve |     Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|v6|16.67 us | 0.206 us | 0.183 us | 0.4883 |     - |     - |  16.92 KB |
| issue-788| 17.50 us | 0.106 us | 0.099 us | 0.4883 |     - |     - |  16.91 KB |

|  ChildScopeResolveBenchmark |     Mean |    Error |   StdDev |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|-------- |---------:|---------:|---------:|-------:|-------:|------:|----------:|
| v6 | 57.70 us | 0.163 us | 0.127 us | 1.0986 | 0.5493 |     - |  37.62 KB |
| issue-788| 58.22 us | 0.277 us | 0.259 us | 1.0986 | 0.5493 |     - |  37.61 KB |

src/Autofac/Core/Resolving/Pipeline/ResolvePipelineBuilder.cs
```diff
@@ -270,34 +270,25 @@ Action<ResolveRequestContextBase> Chain(Action<ResolveRequestContextBase> next,
 
                 return (ctxt) =>
                 {
-                    // Optimise the path depending on whether diagnostics are enabled.
-                    if (ctxt.DiagnosticSource.MiddlewareDiagnosticsEnabled())
+                    ctxt.DiagnosticSource.MiddlewareStart(ctxt, stage);
```
alistairjevans, line 273: If there is a shortcut here for 'are any subscribers attached', using `IsEnabled` without the name parameter, can we use that to conditionally enter the entire diagnostic block vs the simple execution block, as per the original `MiddlewareDiagnosticsEnabled` check?  Just to save on the additional method calls we may not need to make.  

Unless you're thinking someone might add a diagnostic subscriber inside the middleware itself, and expect the Success/Failure events to then fire for that same middleware?

src/Autofac/Core/Resolving/ResolveOperationBase.cs
```diff
@@ -70,9 +73,9 @@ protected ResolveOperationBase(ISharingLifetimeScope mostNestedLifetimeScope)
         public IEnumerable<ResolveRequestContextBase> InProgressRequests => RequestStack;
 
         /// <summary>
-        /// Gets the <see cref="System.Diagnostics.DiagnosticSource"/> for the operation.
+        /// Gets the <see cref="System.Diagnostics.DiagnosticListener"/> for the operation.
         /// </summary>
-        public DiagnosticSource DiagnosticSource { get; }
+        public DiagnosticListener DiagnosticSource { get; }
```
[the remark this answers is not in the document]
tillig: `DiagnosticListener` is what exposes the `IsEnabled()` fast null check; `DiagnosticSource` (the base class) only exposes `IsEnabled(string)` without the faster check. I figured since this was internal, passing around the strong type rather than casting each time made sense.

src/Autofac/Core/Resolving/Pipeline/ResolvePipelineBuilder.cs
tillig: Let me have one more look at this. It was pretty iterative to get where I got so I can see I missed that.

tillig: I updated this to be closer to what we had to begin with:

- Hot path surrounded by `DiagnosticSource.IsEnabled()` (the null check) with one "version" of the pipeline being instrumented and the other being just the stage execution.
- Same thing in `ResolveOperationBase.GetOrCreateInstance` - but since there were more than a couple of lines I made the main pipeline invocation there an aggressively inlined method to avoid accidental copy/paste errors.
- Removed all the `IsEnabled()` checks from the extensions that write the diagnostics. Every write is now surrounded by a call to `IsEnabled()` and, where possible, that's only called one time around a block of code instead of at every potential write.

I can't guarantee it'll be _identical_ to the v6 perf but it should be... within tolerances? I can't imagine what else we could do to optimize the flow - it's boiling down to null checks like it was doing before.

tillig: Would you like to run benchmarks again or should we just call it good?

alistairjevans: I'll run the benchmarks shortly and post back here, they only take a few moments to run.

src/Autofac/Core/Resolving/ResolveOperationBase.cs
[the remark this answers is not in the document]
alistairjevans: 👍

src/Autofac/Core/Resolving/Pipeline/ResolvePipelineBuilder.cs
alistairjevans: New Benchmarks. Still a ~0.5μs difference, but that might be fine.

**DeepGraphResolve - ~0.6μs / 6% slower**

|  Method |     Mean |     Error |    StdDev |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|-------- |---------:|----------:|----------:|-------:|-------:|------:|----------:|
| v6 | 9.539 us | 0.0693 us | 0.0615 us | 2.0599 | 0.0458 |     - |  16.92 KB |
| issue-788 | 10.18 us | 0.067 us | 0.063 us | 2.0599 | 0.0458 |     - |  16.91 KB |

**ChildScopeResolve - ~0.5μs / 2% slower**

|  Method |     Mean |    Error |   StdDev |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|-------- |---------:|---------:|---------:|-------:|-------:|------:|----------:|
| v6 | 31.43 us | 0.215 us | 0.202 us | 3.9063 | 1.9531 |     - |  32.09 KB |
| issue-788 | 32.09 us | 0.186 us | 0.165 us | 4.0894 | 2.0142 |     - |  33.88 KB |

tillig: I'm not sure there's anything else I can do; it's a more robust diagnostics system than previous, and maybe that comes with a price. I'll commit it now and if we can figure out new ways to improve we can update then.

alistairjevans: Yeah, I don't think there's much more to do here, by virtue of a more robust system there's a few more method calls that have to take place, and that's that.

We can track any further investigation under #1127.

alistairjevans:
🍻

tillig:
I appreciate the patience on this - it was definitely a long and winding road to get here, but I think it's good overall. Now let me see about getting that DOT graph tracer updated and PR'd based on these changes. :tada:

PR #1171, opened 2020-07-21
Remove conditional public signing on Linux; use full signing on all platforms

tillig:
Conditional public signing is a remnant from when full signing wasn't available on CoreCLR - it's available now.


alistairjevans:
@tillig, I've definitely picked up your habit of using random emojis to approve PRs. It adds a touch of whimsy to my day.

PR #1172, opened 2020-07-22
Add support for Func<TArgs, Task> handlers for Core Autofac Events

### LINKED ISSUE #1069: None of the OnXxx hooks are async-friendly
None of the OnXxx hooks on `IRegistrationBuilder` are async friendly. Here's the problem I am facing right now.

```
OnActivated(args =>
            {
                var context = args.Instance;
                var cts = new CancellationTokenSource();
                var token = cts.Token;
                context.Entities.RemoveRange(context.Entities);
                context.SaveAsync(token).ConfigureAwait(false).GetAwaiter().GetResult();

                context.Entities.Add(new BasicEntity { Id = 1, Name = "abc" });
                context.Entities.Add(new BasicEntity { Id = 2, Name = "def" });
                context.Entities.Add(new BasicEntity { Id = 3, Name = "ghi" });
                context.SaveAsync(token).ConfigureAwait(false).GetAwaiter().GetResult();
            }
```

If the idea sounds good - to add Func<T, Task> based overloads - you can assign this issue to me for implementation.

alistairjevans:
Fixes #1069 

Added overloads for OnPreparing, OnActivating, OnActivated and OnRelease that allow you to provide a `Func<Task>` handler, so you can do async/await inside those methods.

OnPreparing, OnActivating and OnActivated are not 'true' async, because the implementations are basically just a utility wrapper around `handler(args).ConfigureAwait(false).GetAwaiter().GetResult()`.

So they make it more convenient for developers to do some async operations inside these handlers, but there is no `ResolveAsync` method.

OnRelease has a bit more functionality, because if you dispose of the scope/container using `DisposeAsync`, that **will** do a proper await on the task, so it's an actual async flow.

### DIFF
--- a/src/Autofac/Autofac.csproj
+++ b/src/Autofac/Autofac.csproj
@@ -44,10 +44,10 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.0.0">
+    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.0-beta1.final">
       <PrivateAssets>All</PrivateAssets>
     </PackageReference>
-    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="3.0.0">
+    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="3.3.0-beta2.final">
       <PrivateAssets>All</PrivateAssets>
     </PackageReference>
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" Condition="Exists('$(MSBuildThisFileDirectory)../../.git')">
--- a/src/Autofac/Builder/IRegistrationBuilder.cs
+++ b/src/Autofac/Builder/IRegistrationBuilder.cs
@@ -264,7 +264,7 @@ namespace Autofac.Builder
         /// </summary>
         /// <param name="handler">An event handler; the resolve process will not continue until the returned task completes.</param>
         /// <returns>A registration builder allowing further configuration of the component.</returns>
-        IRegistrationBuilder<TLimit, TActivatorData, TRegistrationStyle> OnPreparing(Func<PreparingEventArgs, Task> handler);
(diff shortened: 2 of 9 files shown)


tillig:
The build failure here is really weird. I'm curious if it has to do with the build environment:

```text
Runtime Environment:
 OS Name:     ubuntu
 OS Version:  18.04
 OS Platform: Linux
 RID:         ubuntu.18.04-x64
 Base Path:   /usr/share/dotnet/sdk/3.1.300/
Host (useful for support):
  Version: 5.0.0-preview.3.20214.6
  Commit:  b037784658
```

Which appears it's .NET 5 preview doing the building for a .NET Core 3.x runtime. Normally should be fine, but might have an edge case?

The build scripts skip install of the specific .NET Core SDK on Linux. I don't know if that needs to be fixed to make this work. I can help with that if you don't have a Linux environment to mess with. I was meaning to do it for my Mac build anyway. Let me know and I'll jump on it.

alistairjevans:
That would be helpful, I don't have an equivalent Linux immediately to hand. Thanks.

tillig:
OK, let me see what I can do. I'll probably just jam it right to v6 if it builds so you can pull it into your branch.

alsami:
> The build scripts skip install of the specific .NET Core SDK on Linux

Just FYI the script used to install the dotnet-sdk is not compatible with Unix bases systems, that's why I added the conditional installation.

tillig:
Yup, there's a bash equivalent. I've updated the scripts and the build seems to work and pull the correct stuff for a Mac/Linux host now. Give that a shot; pushed to `v6` directly.

alistairjevans:
I think we need to update the analyser package.

https://github.com/dotnet/roslyn-analyzers/issues/3642

alistairjevans:
I'll just do that in this branch.

tillig:
D'OH! 🍩

alistairjevans:
Ok, that worked; I had to take the beta of the affected analysers.

alistairjevans:
The best use-case I can see for this is when you are activating a SingleInstance 'thing', that requires some warmup or external resource, rather than a transient instance.

In a singleton scenario, this slow bit only runs once, and all subsequent resolves are fast.

I too am firmly against an async resolve pipeline (and async constructors...).

tanveerbadar:
Thank you for this. I never got around to contributing this particular change myself, and probably would have done it as well as you could.

PR #1174, opened 2020-07-23
chore: replace ResolveOperationBase with interface

alsami:
* Copied missing function signatures from `ResolveOperationBase` into `IResolveOperation`
* Copied implementation of `ResolveOperationBase` into `ResolveOperation`
* Replace usage of `ResolveOperationBase` with `IResolveOperation`
* Make `SegmentStack` public because of `push` / `pop` used implicitly inside `CircularDependencyDetectorMiddleware`  because of the change reference from `ResolveOperationBase` to `IResolveOperation`

Part two of #1164


alsami:
@alistairjevans, I know you have made `SegmentStack` internal on purpose.
Sadly I had to expose it, otherwise the usage of the interface is not possible. I could use the implementation here but then we have to unseal the class `ResolveOperation`.

What are your thoughts on that?

alsami:
Raising the code coverage by deleting code is just fun =) 0.01%. Small win but better than nothing :smile:

alistairjevans:
Regarding the `SegmentedStack`, the type being public isn't the issue; I wanted to avoid giving external consumers the ability to manipulate that stack themselves...too much can go wrong internally if someone messes with it by manually calling 'push' or 'pop'.

As of C# 8, you can now mark interface members as 'internal'. So if you expose the segmented stack as an internal member of `IResolveOperation`, `RequestStack`, with an additional public `ActiveRequests` returning an `IEnumerable`, `SegmentedStack` should be able to stay internal.

As for the list of properties, yes, I believe all those are exposed for extensibility.

alistairjevans:
Actually, I'd ignore that internal interface suggestion; it basically stops anyone outside the library mocking it themselves (why on earth they would have internal interface members supported I have no idea).

alsami:
> Actually, I'd ignore that internal interface suggestion;

So, you fine with how it is now?



> (why on earth they would have internal interface members supported I have no idea).

I think that C# 8 in general enabled interface modifier. I guess if they had to enable default implementations, the only thing they could do was allowing everything except of constructors for an interface.

alistairjevans:
I'm still not very happy about having direct access to manipulate the Segmented Stack outside of the Autofac library itself.  Trying to think of a way to allow the circular dependency middleware to access that item without just exposing it or doing a hacky cast inside the middleware.

If you have any ideas, please throw them in here.

tillig:
This may or may not be great, but... what if the _interface_ `IResolveOperation` did not have the segmented stack thing, just the implementation; and then the circular dependency middleware cast the `IResolveOperation` to the concrete type, effectively making the assumption that you don't literally get to plug in your own `IResolveOperation` and still have our middleware detect circular dependencies?

Now, granted, that's a pretty gnarly violation of the Liskov Substitution Principle, but it'd solve the problem. The whole point of the interface here is to try to allow extensions to be written without having to set up chains of dependencies or fight internal constructors. I don't know that anyone would actually implement their own resolve operation.

But... they could? I guess? I don't know how you'd get that actually flowing through the pipeline if you did, though. Sort of like in ASP.NET Core I don't know how you'd get "your own" `HttpContext` running through the system. Basically, the `HttpContextBase` is just there to enable testing, not to allow you to really implement your own version of the thing.

I guess another option [which I have absolutely not vetted or thought through] would be to move circular dependency detection out of middleware and do all of that just in the resolve operation itself. Then it wouldn't be part of the pipeline and wouldn't need to be exposed at all.

alsami:
> I don't know that anyone would actually implement their own resolve operation.

I agree that this would be a pretty hard task to accomplish, given all the middleware based on that interface. Casting down to the explicit type might work for us, never from the outside though, if we were to make `SegmentStack` internal again.

> Sort of like in ASP.NET Core I don't know how you'd get "your own" `HttpContext` running through the system. Basically, the `HttpContextBase` is just there to enable testing, not to allow you to really implement your own version of the thing.

So would you prefer a class similar to `HttpContextBase` implementation that can be easily used for testing over an interface?

alistairjevans:
I suppose my main goal here is to make it so that when a user is writing custom middleware, that when they look at the intellisense for the `Operation` property, they don't immediately see the ability to do things that will break the world.

As always, we find ourselves trying to strike a balance between things being too complex, and things being too breakable... 

I think a big part of this decision comes down to the following:

> When people implement `IResolveOperation` in tests, are they doing so to pass into their own individual middleware,
> or are they doing it to execute the entire pipeline?

I'd say mostly the former; there's little advantage in mocking the operation if you want to test the entire pipeline, you could just call Resolve normally to do that.

That said, if you were to access the `ResolvePipeline` from a `ServiceRegistration`  returned from `IComponentRegistry`, you could just invoke the pipeline manually, bypassing the internal `ResolveOperation`.

So, similar to @tillig's suggestion, but a bit more Liskov-compliant, my proposal would be to:

- Define a new public interface `IDependencyTrackingResolveOperation` that derives from `IResolveOperation`. Move the `RequestStack` and the `EnterNewDependencyDetectionBlock` method to this interface.  Leave the read-only `ActiveRequests` behind though.
- Implement `IDependencyTrackingResolveOperation` on the `ResolveOperation` class.
- The `CircularDependencyDetectorMiddleware` can now do an `as IDependencyTrackingResolveOperation` to check if the operation can track dependencies, then do it's normal behaviour. If that `as` returns null, the middleware does nothing.
- Same goes for the DecoratorMiddleware where it needs to call `EnterNewDependencyDetectionBlock`.

I think this gives the _ability_ to access properties of the dependency management outside the core Autofac library, but without making it super obvious unless you know that you can cast the normal `IResolveOperation` to `IDependencyTrackingResolveOperation`.

Thoughts?

alistairjevans:
Oh, and @tillig, regarding moving the circular dependency middleware into the operation itself, I did try that very early on, but it caused a few problems around completing an operation, and means that you cannot attach anything to 'just before' the circular dependency middleware.

alsami:
> * The `CircularDependencyDetectorMiddleware` can now do an `as IDependencyTrackingResolveOperation` to check if the operation can track dependencies, then do it's normal behaviour. If that `as` returns null, the middleware does nothing.
> * Same goes for the DecoratorMiddleware where it needs to call `EnterNewDependencyDetectionBlock`.

That would mean, if somebody was to write their own implementation of `IResolveOperation`, they wouldn't get decorator support, nor circular-dependency detection, correct?

Sounds applicable to me, gonna give it a try now.

alistairjevans:
> That would mean, if somebody was to write their own implementation of IResolveOperation, they wouldn't get decorator support, nor circular-dependency detection, correct?

Yes, but I think that's fine? Especially since this is mostly a testing support thing.

alistairjevans:
Also, `IDependencyTrackingResolveOperation` is _also_ public, so if they are really a glutton for punishment, they could build the entire thing.

alsami:
> Yes, but I think that's fine? Especially since this is mostly a testing support thing.

Agree.



> Also, `IDependencyTrackingResolveOperation` is _also_ public, so if they are really a glutton for punishment, they could build the entire thing.

I've made it internal now, should it be public?

alistairjevans:
> I've made it internal now, should it be public?

I'm wondering if we could just say it's public now, just in case we wanted to use it in an extension ourselves (for example). The goal of 'hide the SegmentedStack from 99% of users via intellisense' has already been achieved after all.

On the other hand, if we'd rather keep our public API surface smaller, that makes sense to.

alsami:
Sure, can do.

src/Autofac/Core/Resolving/IDependencyTrackingResolveOperation.cs
```diff
@@ -0,0 +1,27 @@
+using System;
+using Autofac.Core.Resolving.Middleware;
+using Autofac.Core.Resolving.Pipeline;
+
+namespace Autofac.Core.Resolving
+{
+    /// <inheritdoc />
```
alistairjevans, line 7: Comments on why this interface exists would be good I think.

src/Autofac/Core/Resolving/ResolveOperation.cs
```diff
@@ -54,5 +72,232 @@ public object Execute(ResolveRequest request)
...
+
+        /// <summary>
+        /// Enter a new dependency chain block where subsequent requests inside the operation are allowed to repeat
+        /// registrations from before the block.
+        /// </summary>
+        /// <returns>A disposable that should be disposed to exit the block.</returns>
+        public IDisposable EnterNewDependencyDetectionBlock() => RequestStack.EnterSegment();
+
+        /// <inheritdoc/>
+        public SegmentedStack<IResolveRequestContext> RequestStack { get; } = new SegmentedStack<IResolveRequestContext>();
```
alistairjevans, line 192: Small nit, could we move this property to before any public methods (i.e. with the other properties)?

src/Autofac/Features/Decorators/DecoratorMiddleware.cs
```diff
@@ -56,6 +57,14 @@ public DecoratorMiddleware(DecoratorService decoratorService, IComponentRegistra
         /// <inheritdoc/>
         public void Execute(IResolveRequestContext context, Action<IResolveRequestContext> next)
         {
+            if (!(context.Operation is IDependencyTrackingResolveOperation dependencyTrackingResolveOperation))
```
alistairjevans, line 60: I'd suggest moving this check down to after the check for 'is decoration disabled', and keep the style consistent with the other two 'early jump out' checks.

alistairjevans:
Also, what's going on with the code coverage checks?

alsami:
> Also, what's going on with the code coverage checks?

Good question. I think because of the new untested branches. Adding some small tests currently to see if that has influence on it.

src/Autofac/Features/Decorators/DecoratorMiddleware.cs
alsami: Addressed with latest commit

alsami:
Latest looks better, only small diff now, will fix that as well. Adding another if statement here and there apparently does change a lot on the coverage.

alistairjevans:
Sad face for the DeepGraphResolve Benchmark; we're now slower than develop.

|  Method |     Mean |    Error |   StdDev |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|-------- |---------:|---------:|---------:|-------:|-------:|------:|----------:|
| develop | 10.91 us | 0.065 us | 0.058 us | 2.2430 | 0.0458 |     - |  18.45 KB |
| v6 |  10.79 us | 0.069 us | 0.065 us | 2.0599 | 0.0458 |     - |  16.91 KB |
| PR | 11.50 us | 0.051 us | 0.045 us | 2.3804 | 0.0610 |     - |  19.49 KB |

Interfaces are slower than classes (marginally), due to the vtable lookup inside the runtime...perhaps this is why `HttpContextBase`` exists rather than `IHttpContext`.

Now I'm not sure whether to just change these interfaces to all be abstract base classes with 100% abstract methods and properties, and no default constructor. That screws up the `IDependencyTrackingResolveOperation` solution though.

☹

alsami:
Can the type check be optimized maybe? I know that when casting down to interfaces the order is important but there is only one so not sure :(

alistairjevans:
I don't think the type check is the issue, I think it's just generally passing around interfaces instead of classes.  Any time the runtime has to invoke a method on an interface, an additional method table dereference occurs, and we call properties of `IResolveRequestContext` and `IResolveOperation` a lot.

In fairness, the `IResolveOperation` changes probably aren't the issue; it's the `IResolveRequestContext` that's the bigger problem.

alsami:
> In fairness, the `IResolveOperation` changes probably aren't the issue; it's the `IResolveRequestContext` that's the bigger problem.

Then I would say I'll create another PR replacing the Interface with `ResolveRequestContext` which only has abstract methods.
Really time for automated Benchmark comparisons hehe ...

alistairjevans:
Agreed, let's bring this one in, and we can look at the abstract class thing in a subsequent PR.

src/Autofac/Core/Resolving/IDependencyTrackingResolveOperation.cs
alsami: Can you check if the summary I've added is enough?

src/Autofac/Core/Resolving/ResolveOperation.cs
alsami: Addressed with latest commit

alsami:
> Agreed, let's bring this one in, and we can look at the abstract class thing in a subsequent PR.

Applied all the requested changes (hopefully). Let me know if it's fine, so I can rebase ahead of merging please.

src/Autofac/Core/Resolving/IDependencyTrackingResolveOperation.cs
alistairjevans: You've got an XML Doc issue in there, but more broadly you could describe this as:

```
The interface indicating that an <see cref="IResolveOperation" /> supports dependency tracking. Required by middleware that must understand the dependency tree.
```

alsami: Okay, addressed

alistairjevans:
Ready to go if you want to do any rebasing.

alsami:
Good to go, cheers!

PR #1181, opened 2020-07-28
fix: make sure benchmarks can be run everywhere

alsami:
Previously it was not possible to something like this:
`dotnet publish -c Release Autofac.Benchmarks.csproj`
`cd publish`
`dotnet Autofac.Benchmarks.dll --filter *ChildScopeResolveBenchmark*` // throws an exception because there is no bin directory.

The changes made make sure that the `bin` directory must not be present when executing them.

It also does not use hard-coded path-separators anymore, even though that somehow didn't seem to be a problem on Unix based systems.

### DIFF
--- a/bench/Autofac.Benchmarks/BenchmarkConfig.cs
+++ b/bench/Autofac.Benchmarks/BenchmarkConfig.cs
@@ -7,7 +7,7 @@ namespace Autofac.Benchmarks
 {
     internal class BenchmarkConfig : ManualConfig
     {
-        private const string BenchmarkArtificatsFolder = "BenchmarkDotNet.Artifacts";
+        private const string BenchmarkArtifactsFolder = "BenchmarkDotNet.Artifacts";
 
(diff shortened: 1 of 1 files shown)


alsami:
@alistairjevans as discussed on gitter, here the changes. I will send another PR after merging `develop` into `v6`.

bench/Autofac.Benchmarks/BenchmarkConfig.cs
```diff
@@ -1,20 +1,23 @@
 using System;
+using System.IO;
 using BenchmarkDotNet.Configs;
 using BenchmarkDotNet.Diagnosers;
 
 namespace Autofac.Benchmarks
 {
     internal class BenchmarkConfig : ManualConfig
     {
+        private const string BenchmarkArtificatsFolder = "BenchmarkDotNet.Artifacts";
```
alistairjevans, line 10: Typo in the name.

PR #1183, opened 2020-08-04
Null Ref Error - Staging some initial changes to run through CI

alistairjevans:
Starting a PR to see what's going on with the intermittent Null Ref error.

Cannot reproduce locally, so going to have to try here.


alistairjevans:
I would encourage people to mute this PR...

PR #1184, opened 2020-08-05
Minor NRT corrections for test files

alistairjevans:
Not sure why the NRT warnings aren't failing the build, but I've fixed the issue.

### DIFF
--- a/Autofac.sln
+++ b/Autofac.sln
@@ -16,9 +16,11 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution 
 		.gitignore = .gitignore
 		appveyor.yml = appveyor.yml
 		build.ps1 = build.ps1
+		codecov.yml = codecov.yml
+		Directory.Build.props = Directory.Build.props
+		Directory.Build.targets = Directory.Build.targets
 		global.json = global.json
 		NuGet.Config = NuGet.Config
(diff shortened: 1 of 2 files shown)


alistairjevans:
Aaargghhhhh!!! The build failed on that CI problem that I've failed to replicate all of yesterday!!! Sob.

tillig:
Before merging... it looks like the code coverage thing changed and is failing the check. I wonder if we need to tweak that; I don't see that we can fix coverage on nullable markup any more than we could affect coverage when adding an `async`/`await` operation where the compiler generates that whole async wait state machine thing. I've been hit by that before, too - can't cover that stupid async/await thing but the lines of code themselves are covered.

alistairjevans:
Is that a setting in CodeCov for a tolerance threshold, or is it a setting when we run the build? @alsami?

alistairjevans:
I'm going to merge this in spite of the codecov problem because it's blocking the merge to develop.

alsami:
> Is that a setting in CodeCov for a tolerance threshold, or is it a setting when we run the build? @alsami?

I will look into it. If it's possible, what threshold would you think is acceptable? 1%? 2%? We could also decide to not let the checks fail, when the coverage lowered.

alistairjevans:
1% might be enough for now to ignore minute fluctuations? 

Not sure, may be a case of trial and error.

PR #1186, opened 2020-08-06
Merge v6 into develop

alistairjevans:
Ok, this is the big one!

I'm raising a PR for this so we have the opportunity for a cursory check, but not really expecting an actual review since all the changes have been reviewed separately.


src/Autofac/Properties/AssemblyInfo.cs
```diff
@@ -8,8 +8,8 @@
...
 [assembly: InternalsVisibleTo("Autofac.Test.Uwp.DeviceRunner, PublicKey=00240000048000009400000006020000002400005253413100040000010001008728425885ef385e049261b18878327dfaaf0d666dea3bd2b0e4f18b33929ad4e5fbc9087e7eda3c1291d2de579206d9b4292456abffbe8be6c7060b36da0c33b883e3878eaf7c89fddf29e6e27d24588e81e86f3a22dd7b1a296b5f06fbfb500bbd7410faa7213ef4e2ce7622aefc03169b0324bcd30ccfe9ac8204e4960be6")]
 [assembly: InternalsVisibleTo("Autofac.Net46.Test, PublicKey=00240000048000009400000006020000002400005253413100040000010001008728425885ef385e049261b18878327dfaaf0d666dea3bd2b0e4f18b33929ad4e5fbc9087e7eda3c1291d2de579206d9b4292456abffbe8be6c7060b36da0c33b883e3878eaf7c89fddf29e6e27d24588e81e86f3a22dd7b1a296b5f06fbfb500bbd7410faa7213ef4e2ce7622aefc03169b0324bcd30ccfe9ac8204e4960be6")]
-[assembly: CLSCompliant(true)]
+[assembly: CLSCompliant(false)]
 [assembly: ComVisible(false)]
 [assembly: NeutralResourcesLanguage("en-US")]
 [assembly: AssemblyCopyright("Copyright © 2015 Autofac Contributors")]
```
tillig, line 14: Should we move copyright, description and other packaging things into the .csproj?

I also see above that we have, like, `Autofac.Net46.Test` and `Autofac.Test.Uwp.DeviceRunner` here - I don't think we actually have tests for those anymore, do we? I think they're old, back when we were fighting with portable class library issues.

bench/Autofac.BenchmarkProfiling/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "profiles": {
+    "Autofac.BenchmarkProfiling": {
+      "commandName": "Project",
+      "commandLineArgs": "KeyedNestedBenchmark 2"
```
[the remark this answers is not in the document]
alistairjevans: The default can be blank I suppose.

src/Autofac/Properties/AssemblyInfo.cs
alistairjevans: I'll clean it up.

alistairjevans: But for the moment I'll leave attributes in here rather than move into the csproj. We can do that later if we want.

PR #1188, opened 2020-08-11
feat: introduce resolve-request abstraction

alsami:
related to #1164


alsami:
@autofac-bot benchmark DeepGraphResolveBenchmark

alsami:
@autofac-bot benchmark DeepGraphResolveBenchmark

alsami:
@autofac-bot benchmark DeepGraphResolveBenchmark

alsami:
@autofac-bot benchmark DeepGraphResolveBenchmark

alsami:
@tillig @alistairjevans basically ready for review.

I ran the benchmarks initially using the bot but the differences were too big comparing develop and PR branch. 

I did that for the first commit (twice), which introduced an interface. For the second commit as well, which kind of used the same principle we have introduced for `ResolveRequestContext`. Then I decided to run them manually again, which seems to bring better results but still, does make a difference.

develop:
| Resolve | 17.91 us | 0.351 us | 0.329 us | 0.4883 |     - |     - |  16.91 KB |


PR branch:
| Resolve | 18.04 us | 0.114 us | 0.107 us | 0.4883 |     - |     - |  16.91 KB |

Any thought's on that?

alistairjevans:
What is the motivation behind making ResolveRequest an interface / abstract class?

I ask because it is a very small type, it can be relatively easily mocked (especially now), and it is very well established as a type you can instantiate and do stuff with. It's going to introduce a lot of extra breaking change.

Might I propose that we have a go updating the unit tests with the changes made so far before we do more refactoring?

It would be good to only change ResolveRequest if we absolutely have to.

alsami:
> What is the motivation behind making ResolveRequest an interface / abstract class?

Basically the requirement to make testing easier and only really have to mock what you need, since it has one constructor with all the parameters required.

> I ask because it is a very small type, it can be relatively easily mocked (especially now), and it is very well established as a type you can instantiate and do stuff with. It's going to introduce a lot of extra breaking change.

I agree, this class has been public for a very long time probably. Most of the parameters can be rather easily mocked as well, but still require mocking.

> Might I propose that we have a go updating the unit tests with the changes made so far before we do more refactoring?

We actually already did that, if you are talking about `ResolveRequestContext`. Or what is it, that you are referring to?

alistairjevans:
I'm thinking about things like [ActivatorPipelineExtensions](https://www.github.com/autofac/Autofac/tree/develop/test/Autofac.Test/ActivatorPipelineExtensions.cs), which has been changed for the new naming, but not properly refactored to take advantage of the new abstract classes.

That class only exists because of how complex it used to be to mock a pipeline invocation.

I feel like we should try reworking things to use the new abstract classes before we do more refactoring.

We may have already made enough changes to satisfy #1164.

alsami:
> I'm thinking about things like [ActivatorPipelineExtensions](https://www.github.com/autofac/Autofac/tree/develop/test/Autofac.Test/ActivatorPipelineExtensions.cs), which has been changed for the new naming, but not properly refactored to take advantage of the new abstract classes.

Okay cool, will look into it.



> We may have already made enough changes to satisfy #1164.

Fine for me, just understand that it was kind of a nice to have to make it even easier in the specific area.

Closing this for now

alistairjevans:
Thanks; it might be we still need this, but it would be good to know for sure.

PR #1189, opened 2020-08-13
Style changes - switch to requiring braces

alistairjevans:
Also did some other minor suggestion fixes.

This was....tedious. Couldn't use an automatic way because of the various cases where throw expressions were better.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -33,6 +33,8 @@ csharp_indent_case_contents_when_block = true
 csharp_indent_switch_labels = true
 csharp_indent_labels = one_less_than_current
 
+csharp_prefer_braces = true
+
 ; Modifier preferences
 csharp_preferred_modifier_order = public,private,protected,internal,static,extern,new,virtual,abstract,sealed,override,readonly,unsafe,volatile,async:suggestion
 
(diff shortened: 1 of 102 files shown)


.editorconfig
```diff
@@ -33,6 +33,8 @@ csharp_indent_case_contents_when_block = true
 csharp_indent_switch_labels = true
 csharp_indent_labels = one_less_than_current
 
+csharp_prefer_braces = true
```
tillig, line 36: Line 89 already has `csharp_prefer_braces = true:refactoring`

.editorconfig
```diff
@@ -55,14 +57,23 @@ dotnet_naming_symbols.constant_fields.applicable_kinds   = field
...
 
+; Static readonly fields should be PascalCase
+dotnet_naming_rule.static_readonly_fields_should_be_pascal_case.severity = suggestion
+dotnet_naming_rule.static_readonly_fields_should_be_pascal_case.symbols  = static_readonly_fields
+dotnet_naming_rule.static_readonly_fields_should_be_pascal_case.style    = pascal_case_style
+dotnet_naming_symbols.static_readonly_fields.applicable_kinds   = field
+dotnet_naming_symbols.static_readonly_fields.required_modifiers = static, readonly
+dotnet_naming_symbols.static_readonly_fields.applicable_accessibilities = private, internal, private_protected
+
+
```
tillig, line 76: Just one blank space. Unsure how to make a suggestion that removes a line. It's not... rendering right when I try.

build/Analyzers.ruleset
```diff
@@ -2,57 +2,31 @@
 <RuleSet Name="Autofac Analyzer Rules" Description="Analyzer rules for Autofac assemblies." ToolsVersion="16.0">
   <IncludeAll Action="Warning" />
   <Rules AnalyzerId="Microsoft.Usage" RuleNamespace="Microsoft.Usage">
-    <!-- Implement standard exception constructors - not all of the exception constructors (e.g., parameterless) are desired in our system. -->
```
tillig, line 5: This seems to happen a lot - the comments explaining what the rules are and why we've omitted them get dropped. I find it helpful to know what the rules are; unclear if others do. Might need to stop editing this one in VS.

src/Autofac/Core/Registration/ServiceRegistrationInfo.cs
```diff
@@ -43,6 +43,7 @@ internal class ServiceRegistrationInfo : IResolvePipelineBuilder
         private volatile bool _isInitialized;
 
         [SuppressMessage("Microsoft.Performance", "CA1823:AvoidUnusedPrivateFields", Justification = "The _service field is useful in debugging and diagnostics.")]
+        [SuppressMessage("CodeQuality", "IDE0052:Remove unread private members", Justification = "The _service field is useful in debugging and diagnostics.")]
```
tillig, line 46: Perhaps something should actually be _done_ with the field. For example, a `ToString` override or a [`DebuggerDisplay` implementation that uses it](https://docs.microsoft.com/en-us/visualstudio/debugger/using-the-debuggerdisplay-attribute?view=vs-2019).

src/Autofac/Features/Metadata/MetadataViewProvider.cs
```diff
@@ -26,6 +26,7 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
```
tillig, line 29: I think this is unused.

test/Autofac.Test/Features/LightweightAdapters/LightweightAdapterRegistrationExtensionsTests.cs
```diff
@@ -149,7 +147,7 @@ public DecoratorParameterization()
             [Fact]
             public void ParametersGoToTheDecoratedInstance()
             {
-                var expected = new Implementer1();
+                _ = new Implementer1();
```
tillig, line 150: Uh... what is... uh... going on here?

build/Analyzers.ruleset
alistairjevans: I somehow manage to do this every single time, I need to put a post-it on my monitor not to open the rulesets file in VS...

test/Autofac.Test/Features/LightweightAdapters/LightweightAdapterRegistrationExtensionsTests.cs
alistairjevans: The refactoring suggested I could use the discard operator, because 'expected' isn't used, but what I should actually have done is just remove the line.

src/Autofac/Core/Registration/ServiceRegistrationInfo.cs
alistairjevans: Makes sense; used it in a ToString implementation.

PR #1191, opened 2020-08-17
Support for delegates providing Open Generics

### LINKED ISSUE #1190: Autofac equivalent of NInject's ToMethod() with open generics
## Problem Statement

I'm trying to migrate to Autofac from NInject, and I'm running into a seemingly simple problem.  It does not appear Autofac has a way of applying a factory method to an open generic binding.

## Desired Solution

NInject's solution seems about as concise as one could hope for:

```cs
Kernel.Bind(typeof(IOptions<>)).ToMethod(c => CreateOptions(c.Request.Service));
```

In other words, when a closed `IOptions<>` is required, call `CreateOptions()` with the requested type and let me create the implementation.

This solution is also nice because one can treat this binding just like any other and go on to fluently set interceptors, lifetime, etc.

## Alternatives You've Considered

The best I've come up with in Autofac is either:

1. Create a dummy implementation of `IOptions<>`, let Autofac construct it with `RegisterGeneric()`, then use `ReplaceInstance()` in the **OnActivating** event.
2. Create an `IRegistrationSource` that filters by type and then calls my factory method.

Method (1) seems kludgey - surely I don't have to create an unused object just to get its type?

Method (2) is okay, but it's a *mass* of code (see below), and it gets even bigger if I want to set things like scope, etc.

In both cases the amount of code to simply delegate to a factory method seems wild, especially compared to Autofac's slick factory system for non-open generic types.

(For completeness, below is my `IRegistrationSource`.  Note the two abstract members that must be overridden to get the NInject functionality.)

```cs
public abstract class DelegatingRegistrationSource : IRegistrationSource
{
    protected abstract bool CanDelegate(Type limitType);
    protected abstract DelegateActivator CreateDelegate(Type limitType);

    public IEnumerable<IComponentRegistration> RegistrationsFor(Service service,
        Func<Service, IEnumerable<IComponentRegistration>> registrationAccessor)
    {
        if (!(service is IServiceWithType swt) || !CanDelegate(swt.ServiceType))
            return Enumerable.Empty<IComponentRegistration>();

        var registration = new ComponentRegistration(
            Guid.NewGuid(),
            CreateDelegate(swt.ServiceType),
            new CurrentScopeLifetime(),
            InstanceSharing.None,
            InstanceOwnership.OwnedByLifetimeScope,
            new[] { service },
            new Dictionary<string, object>());

        return new IComponentRegistration[] { registration };
    }

    public bool IsAdapterForIndividualComponents => false;
}

```

## Additional Context

This request stems from my [StackOverflow question](https://stackoverflow.com/q/63419785/222481) and the request there from Alistair Evans.

Thanks for your time!

alistairjevans:
As per #1190, provide a mechanism (via a new registration source) where a delegate can provide a closed generic for an open generic service.

I've added as many tests as I can think of; I'll take a look at the codecov results to see what I may have missed amongst the new changes.

Closes  #1190


src/Autofac/Features/OpenGenerics/OpenGenericDelegateActivatorData.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// </summary>
+    public class OpenGenericDelegateActivatorData
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="OpenGenericDelegateActivatorData"/> class.
+        /// </summary>
+        /// <param name="factory">The factory method that will create a closed generic instance.</param>
+        public OpenGenericDelegateActivatorData(Func<IComponentContext, Type[], IEnumerable<Parameter>, object> factory)
+        {
+            Factory = factory;
```
tillig, line 21:
```suggestion
            Factory = factory ?? throw new ArgumentNullException(nameof(factory));
```

src/Autofac/Features/OpenGenerics/OpenGenericRegistrationExtensions.cs
```diff
@@ -59,6 +62,29 @@ internal static class OpenGenericRegistrationExtensions
             return rb;
         }
 
+        /// <summary>
+        /// Register an un-parameterised generic type, e.g. Repository&lt;&gt;.
+        /// Concrete types will be made as they are requested, e.g. with Resolve&lt;Repository&lt;int&gt;&gt;().
+        /// </summary>
+        /// <param name="builder">Container builder.</param>
+        /// <param name="factory">The open generic implementation type.</param>
```
tillig, line 70:
```suggestion
        /// <param name="factory">Delegate responsible for generating an instance of a closed generic based on the open generic type being registered.</param>
```

src/Autofac/Features/OpenGenerics/OpenGenericRegistrationExtensions.cs
```diff
@@ -86,6 +112,28 @@ internal static class OpenGenericRegistrationExtensions
             return rb;
         }
 
+        /// <summary>
+        /// Creates a registration builder for an un-parameterised generic type, e.g. Repository&lt;&gt;, without registering it.
+        /// Concrete types will be made as they are requested, e.g. with Resolve&lt;Repository&lt;int&gt;&gt;().
+        /// </summary>
+        /// <param name="factory">The factory method.</param>
```
tillig, line 119:
```suggestion
        /// <param name="factory">Delegate responsible for generating an instance of a closed generic based on the open generic type being registered.</param>
```

src/Autofac/Features/OpenGenerics/OpenGenericServiceBinder.cs
```diff
@@ -66,6 +65,33 @@ internal static class OpenGenericServiceBinder
             return false;
         }
 
+        /// <summary>
+        /// Given a closed generic service (that is being requested), creates a closed generic implementation type
+        /// and associated services from the open generic implementation and services.
+        /// </summary>
+        /// <param name="closedService">The closed generic service to bind.</param>
+        /// <param name="configuredOpenGenericServices">The set of configured open generic services.</param>
+        /// <param name="openGenericFactory">The open generic factory delegate.</param>
```
tillig, line 74:
```suggestion
        /// <param name="openGenericFactory">Delegate responsible for generating an instance of a closed generic based on the open generic type being registered.</param>
```

src/Autofac/Features/OpenGenerics/OpenGenericServiceBinder.cs
```diff
@@ -121,6 +147,49 @@ internal static class OpenGenericServiceBinder
             return false;
         }
 
+        /// <summary>
+        /// Given a closed generic service (that is being requested), creates a regular delegate callback
+        /// and associated services from the open generic delegate and services.
+        /// </summary>
+        /// <param name="serviceWithType">The closed generic service to bind.</param>
+        /// <param name="configuredOpenGenericServices">The set of configured open generic services.</param>
+        /// <param name="openGenericFactory">The open generic factory delegate.</param>
```
tillig, line 156:
```suggestion
        /// <param name="openGenericFactory">Delegate responsible for generating an instance of a closed generic based on the open generic type being registered.</param>
```

src/Autofac/RegistrationExtensions.Generics.cs
```diff
@@ -48,5 +51,29 @@ public static partial class RegistrationExtensions
         {
             return OpenGenericRegistrationExtensions.RegisterGeneric(builder, implementer);
         }
+
+        /// <summary>
+        /// Register a delegate that can provide instances of an open generic registration.
+        /// </summary>
+        /// <param name="builder">The container builder.</param>
+        /// <param name="factory">A delegate that receives an array of <see cref="Type"/>, containing the generic type arguments of the requested service.</param>
```
tillig, line 59:
```suggestion
        /// <param name="factory">Delegate responsible for generating an instance of a closed generic based on the open generic type being registered. It will be called with the current <see cref="IComponentContext"/> and the array of generic type arguments expected to be in the closed generic.</param>
```

src/Autofac/RegistrationExtensions.Generics.cs
```diff
@@ -48,5 +51,29 @@ public static partial class RegistrationExtensions
...
+            RegisterGeneric(this ContainerBuilder builder, Func<IComponentContext, Type[], object> factory)
+        {
+            return builder.RegisterGeneric((ctxt, types, parameters) => factory(ctxt, types));
+        }
+
+        /// <summary>
+        /// Register a delegate that can provide instances of an open generic registration.
+        /// </summary>
+        /// <param name="builder">The container builder.</param>
+        /// <param name="factory">A delegate that receives an array of <see cref="Type"/>, containing the generic type arguments of the requested service.</param>
```
tillig, line 71:
```suggestion
        /// <param name="factory">Delegate responsible for generating an instance of a closed generic based on the open generic type being registered. It will be called with the current <see cref="IComponentContext"/>, the array of generic type arguments expected to be in the closed generic, and any parameters passed to the resolve operation.</param>
```

test/Autofac.Test/Features/OpenGenerics/OpenGenericDelegateTests.cs
```diff
@@ -0,0 +1,141 @@
+﻿using System;
```
tillig, line 1: These tests should probably go in the specification project since this will be in the public API and we'll want to make sure we don't break the usage later.

test/Autofac.Test/Features/OpenGenerics/OpenGenericDelegateTests.cs
```diff
@@ -0,0 +1,141 @@
...
+        public void CastExceptionIfDelegateReturnsBadObject()
+        {
+            var builder = new ContainerBuilder();
+
+            builder.RegisterGeneric((ctxt, types) => "bad")
+                   .As(typeof(IInterfaceA<>));
+
+            var container = builder.Build();
+
+            Assert.Throws<InvalidCastException>(() => container.Resolve<IInterfaceA<string>>());
```
tillig, line 138: I'm wondering if this should be a `DependencyResolutionException` with the cast exception inside. That way we know what was _trying_ to be resolved, and then the inner exception explains more about why it got jacked up.

tillig: Also, additional test case - what happens if the delegate throws? Like, maybe it throws `DivideByZeroException`. What bubbles up?

alistairjevans: I did debate whether to add specific behaviour to check the generated type, but I'm now wondering if instead I should throw a try/catch around [the actual cast](https://github.com/autofac/Autofac/blob/develop/src/Autofac/ResolutionExtensions.cs#L314) and throw the nested error.

tillig: There are a few of those casts, like [`ResolveOptionalService`](https://github.com/autofac/Autofac/blob/develop/src/Autofac/ResolutionExtensions.cs#L617) has a few. I think it'd be fine to handle it at that level, it'd just have to be handled in all the spots.

alistairjevans: OK, the cast wrapping is in.  I've also added that additional test case, which is treated like any activator exception, and wrapped as normal.

test/Autofac.Test/Features/OpenGenerics/OpenGenericDelegateTests.cs
alistairjevans: Moved.

alistairjevans:
Well, I enjoyed the hypothetical joke anyway.

PR #1192, opened 2020-08-19
Expose a Registration's Resolve Pipeline in a Fluent Manner

alistairjevans:
I'm working on the docs for the pipelines stuff, and realised that the current way of adding to a registration's resolve pipeline (using a `ResolvePipeline` property that is not displayed in intellisense) is pretty rubbish, and completely goes against the fluent nature of registering something.

With these changes the new usage looks like this:

```csharp
var registration = builder.RegisterType<MyImplementation>().As<IMyService>().Pipeline(p =>
{
    p.Use(PipelinePhase.RegistrationPipelineStart, (ctxt, next) =>
    {
        Console.WriteLine("Yay");
    });
});
```

I'm happy to take suggestions for better method names than `Pipeline`.

### DIFF
--- a/src/Autofac/Builder/IRegistrationBuilder.cs
+++ b/src/Autofac/Builder/IRegistrationBuilder.cs
@@ -334,6 +334,6 @@ namespace Autofac.Builder
         /// </summary>
         /// <param name="configurationAction">An action that can configure the registration's pipeline.</param>
         /// <returns>A registration builder allowing further configuration of the component.</returns>
-        IRegistrationBuilder<TLimit, TActivatorData, TRegistrationStyle> Pipeline(Action<IResolvePipelineBuilder> configurationAction);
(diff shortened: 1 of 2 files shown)


tillig:
We have `WithParameter` - maybe `WithPipeline`?

Otherwise this looks great, very simple.

alistairjevans:
I thought about With, but my only reservation would that there already is a pipeline, you're just adding to it...

Perhaps `AddToPipeline`?

tillig:
Other ideas:

- `ConfigurePipeline` - sort of like using `IOptions<T>` where you can configure options
- `PatchPipeline` - I've been working in REST APIs too much recently

I started thinking about what the ASP.NET Core verbs are, like, `UsePipeline` or something like that. Not sure if that's better than `WithPipeline` since it has the same sort of "replacement" connotation.

alistairjevans:
I like `ConfigurePipeline`. It implies a different type of configuration behaviour compared to the other methods (that don't have Configure at the start), which I like.

PR #1198, opened 2020-09-09
Remove some assemblyinfo attributes and replace with project properties

alistairjevans:
As discussed in the MVC integration. 

The AssemblyDescription has no project property, so will just use the Description property normally allocated for the NuGet package.


tillig (approved):
🕵️‍♀️

PR #1202, opened 2020-09-27
Add OperationSucceeded flag to OperationTraceCompletedArgs.

### LINKED ISSUE #1201: Diagnostics: Add 'Succeeded' flag to OperationTraceCompletedArgs
While putting together a blog post on the v6 release, it occurred to me that we do not have a way for Diagnostic Event handlers to know whether the Trace they are getting is for a successful resolve operation, or a failing one.

I can imagine users might want to save failing resolves to whatever output, to reduce the amount of trace they have to wade through when diagnosing errors in a bigger system.

I think we should add a boolean `OperationSucceeded` flag to `OperationTraceCompletedArgs`, which just indicates whether the operation failed or not. We already have this information available in the Diagnostic Tracer, so should be a trivial addition.

I'll raise a PR for the change, but thought I'd raise the issue first for discussion.

alistairjevans:
Just add the flag to the event object.

Fixes #1201.

### DIFF
--- a/src/Autofac/Diagnostics/DefaultDiagnosticTracer.cs
+++ b/src/Autofac/Diagnostics/DefaultDiagnosticTracer.cs
@@ -198,7 +198,7 @@ namespace Autofac.Diagnostics
                     builder.AppendLine(TracerMessages.ExitBrace);
                     builder.AppendException(TracerMessages.OperationFailed, data.OperationException);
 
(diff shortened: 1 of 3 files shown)


alsami (approved):

tillig:
We'll need a corresponding update in the DotGraph package.

tillig (approved):
💫

alistairjevans:
I'll raise an issue in the DotGraph repo, we can fix it when we update it to the released v6 version.

PR #1205, opened 2020-09-28
Capture prop injection instance when the middleware runs

### LINKED ISSUE #1204: Decorated Instances do not get properties injected if they are marked with AllowCircularDependencies
Related to #1200, but this is definitely just a v6 issue that wasn't in v5.

If you have a decorated instance, with property injection, and circular dependencies allowed on said property injection, the properties won't be injected.

The instance used for property injection in that event is the decorator, rather than the instance.

This (new) test fails:

```csharp
private interface IMyService
{
    void AssertProp();
}

private sealed class DecoratedService : IMyService
{
    public string Prop { get; set; }

    public void AssertProp()
    {
        if (Prop is null)
        {
            throw new NullReferenceException();
        }
    }
}

private sealed class ServiceDecorator : IMyService
{
    private readonly IMyService _decorating;

    public ServiceDecorator(IMyService decorating)
    {
        _decorating = decorating;
    }

    public void AssertProp()
    {
        _decorating.AssertProp();
    }
}

[Fact]
public void DecoratedInstanceWithPropertyInjectionAllowingCircularReferencesStillInjects()
{
    var val = "Value";

    builder.RegisterInstance(val);
    builder.RegisterType<DecoratedService>().As<IMyService>().PropertiesAutowired(PropertyWiringOptions.AllowCircularDependencies);
    builder.RegisterDecorator<ServiceDecorator, IMyService>();

    var instance = container.Resolve<IMyService>();

    instance.AssertProp();
}
```

Luckily, unlike #1200, this one should be an easy fix. I just need to ensure I capture the new instance and pass that into the event handler.

alistairjevans:
This fixes #1204; we capture the activated instance in the middleware now, and while we do populate it later, we're populating the right instance.

### DIFF
--- a/src/Autofac/Builder/RegistrationBuilder{TLimit,TActivatorData,TRegistrationStyle}.cs
+++ b/src/Autofac/Builder/RegistrationBuilder{TLimit,TActivatorData,TRegistrationStyle}.cs
@@ -584,11 +584,13 @@ namespace Autofac.Builder
 
                 if (allowCircularDependencies)
                 {
+                    var capturedInstance = ctxt.Instance;
+
                     // If we are allowing circular deps, then we need to run when all requests have completed (similar to Activated).
                     ctxt.RequestCompleting += (o, args) =>
                     {
                         var evCtxt = args.RequestContext;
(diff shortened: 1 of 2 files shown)


tillig (approved):
I'm glad this appears to be such a simple fix!

PR #1220, opened 2020-10-24
Use ephemeral service info blocks for any lookups before the registry is built.

### LINKED ISSUE #1217: Decorator + IfNotRegistered on the decorated type breaks in 6.0
## Describe the Bug
If I have a service registration that is registered with `IfNotRegistered`, and then try to decorate it with `RegisterDecorator`, the decorator is not resolved from the container when resolving the interface. This is new in autofac 6.0.

## Steps to Reproduce

```c#
public class ReproTest
{
  [Fact]
  public void Repro()
  {
    builder.RegisterType<X>().As<IX>().IfNotRegistered(typeof(IX));
    builder.RegisterDecorator<DX, IX>();
    using var container = builder.Build();
    var instance = container.Resolve<IX>();
    Assert.Equal(instance.GetType(), typeof(DX));
  }
  public interface IX { }
  public class X : IX { }
  public class DX : IX
  {
    public DX(IX wrapped) { }
  }
}
```

## Expected Behavior

I would expect the decorator to be applied to the resolved service, as it was in previous versions.

## Dependency Versions

Autofac: 6.0.0

alistairjevans:
This change constructs a temporary set of `ServiceRegistrationInfo` for calls to `GetInitializedServiceInfo` that take place during the build callbacks that populate the `DefaultRegisteredServicesTracker`.

Marginal extra performance cost to `IfNotRegistered` calls at container build time to allocate a new set, and the overhead of evaluating sources multiple times, but once the tracker is marked as `Complete`, the only extra cost is a single boolean flag check the first time we initialize a `ServiceRegistrationInfo`. No concurrency issues expected, because the Container Build process is always a single-threaded operation anyway.

Fixes #1217.  Also added a test for conditional registrations based on dynamic sources, and on service middleware.

Any suggestions of additional test cases would be most welcome.

### DIFF
--- a/test/Autofac.Specification.Test/Registration/RegistrationOnlyIfTests.cs
+++ b/test/Autofac.Specification.Test/Registration/RegistrationOnlyIfTests.cs
@@ -129,6 +129,78 @@ namespace Autofac.Specification.Test.Registration
         }
 
         [Fact]
+        public void IfNotRegistered_CanBeDecoratedByModuleWhenModuleRegistered1st()
+        {
+            var builder = new ContainerBuilder();
+
+            var module = new MyModule(build =>
+            {
+                build.RegisterDecorator<Decorator, IService>();
+            });
+
+            builder.RegisterModule(module);
+            builder.RegisterType<ServiceA>().As<IService>().IfNotRegistered(typeof(IService));
+
+            var container = builder.Build();
+            var result = container.Resolve<IService>();
+            Assert.IsType<Decorator>(result);
+        }
+
+        [Fact]
+        public void IfNotRegistered_CanBeDecoratedByModuleWhenModuleRegistered2nd()
+        {
+            var builder = new ContainerBuilder();
+
+            var module = new MyModule(build =>
+            {
+                build.RegisterDecorator<Decorator, IService>();
+            });
+
+            builder.RegisterType<ServiceA>().As<IService>().IfNotRegistered(typeof(IService));
+            builder.RegisterModule(module);
+
+            var container = builder.Build();
+            var result = container.Resolve<IService>();
+            Assert.IsType<Decorator>(result);
+        }
+
+        [Fact]
+        public void IfNotRegistered_ConditionalInModuleIsDecoratedWhenModuleRegistered1st()
+        {
+            var builder = new ContainerBuilder();
+
+            var module = new MyModule(build =>
+            {
+                build.RegisterType<ServiceA>().As<IService>().IfNotRegistered(typeof(IService));
+            });
+
+            builder.RegisterModule(module);
+            builder.RegisterDecorator<Decorator, IService>();
+
+            var container = builder.Build();
+            var result = container.Resolve<IService>();
+            Assert.IsType<Decorator>(result);
+        }
+
+        [Fact]
+        public void IfNotRegistered_ConditionalInModuleIsDecoratedWhenModuleRegistered2nd()
+        {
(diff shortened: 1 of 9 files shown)


tillig:
If the purpose is to resolve #1217 and related things, brainstorming test cases:

- [Un-skip the test for #1217](https://github.com/autofac/Autofac/blob/bda20f902cf1d99f50f5f9c309705f58b29b5f0c/test/Autofac.Specification.Test/Registration/RegistrationOnlyIfTests.cs#L53)
- Test combinations in different orders of things that run build callbacks:
  - Registering a module does a callback to run `module.Configure` - does `IfNotRegistered` _outside_ the module work with a decorator registered _inside_ a module (and vice versa)?
  - `RegisterServiceMiddlewareSource` uses a callback for stuff like decorating open generics. Can we (still?) register an open generic outside a module and a decorator inside a module and have it work (and vice versa)?
  - Decorate a component that's `IStartable`. (Maybe not the actual `IStartable` interface, but some other interface on the component.)

This seems like some sort of combinatoric data driven test could be done but I'm not sure how that'd work. Probably too many indirections to lambdas and things to make it intelligible long term.

alistairjevans:
Thanks @tillig; I'll probably just create the additional tests directly. As you say, the setup for them is going to be a bit too varied.

alistairjevans:
Added more tests based on your comments.

We already have a `StartableTypesCanBeDecorated` test in `DecoratorTests`, so I didn't add another one.

test/Autofac.Specification.Test/Registration/RegistrationOnlyIfTests.cs
```diff
@@ -62,6 +83,123 @@ public void IfNotRegistered_CanBeDecorated()
...
+            var result = container.Resolve<SimpleGeneric<int>>();
+            Assert.IsType<SimpleGeneric<int>>(result);
+        }
+
+        [Fact]
+        public void IfNotRegistered_CanConditionGenericServiceWithMultipleInterfaces()
+        {
+            var builder = new ContainerBuilder();
+            builder.RegisterGeneric(typeof(MultiServiceGeneric<>)).As(typeof(IService<>)).As(typeof(IService2<>));
+            builder.RegisterGeneric(typeof(MultiServiceGeneric<>)).As(typeof(IService<>)).IfNotRegistered(typeof(IService2<int>));
```
tillig, line 102: I think this would be more clear if two different open generics were being registered. Regardless of whether `IfNotRegistered` works, `IService<>` will be a `MultiServiceGeneric<>`.

tillig (changes requested):
Minor test thing, otherwise I think this is a good solution to the problem. I can't think of a way to deal with it other than this and I've been sort of churning it around in my mind for a few days.

tillig:
I sometimes wonder if the fluent callback mechanism for registration is worth the trouble. `IServiceCollection` seems so much easier to deal with in the general case.

alistairjevans:
I can see some of the benefit of the `IServiceCollection` approach, but it's worth noting that each individual method for adding services to the collection in MS DI has little or nothing in the way of additional configuration options for that registration.  Not sure how well that model would continue to hold up as features are added.

tillig (approved):
🔏

tillig:
I don't think we should change to collection-based registration necessarily, just sometimes the callbacks and the callbacks that call callbacks and the order of those callbacks seems... pretty complex for what actually ends up happening.

PR #1221, opened 2020-10-26
初始化

xrcdev:
去掉 global.json


tillig:
We want to keep our global.json, but thanks!

Issue #842, opened 2017-04-01
Crash on IOS with linker enabled


tzachshabtay:
On IOS, when the mono linker is enabled (and it’s enabled by default and recommended by Xamarin) it strips out all parts of the framework which are not in use. The linker does not support reflection, however, therefore it does not recognize the call to 'GetRuntimeMethod(nameof(Convert.ChangeType)' (which is called in ConstructorParameterBinding -> ConvertPrimitiveType) as using the 'Convert.ChangeType' method and strips it away unless it's explicitly used somewhere else, crashing Autofac if that method is called (null reference exception as ‘GetRuntimeMethod’ returns null).

Note PR https://github.com/autofac/Autofac/pull/841 that I opened as possible solution for that.

To reproduce, try resolving a constructor with optional float parameters on an IOS device when the linker is enabled. Note that by default the linker is disabled on the IOS simulator but enabled on a device, so you won't reproduce it on the simulator unless you explicitly enable the linker on the simulator.

As a real case for reproduction, you can use this branch (note that I will add a workaround in a future commit so get the commit from this day): https://github.com/tzachshabtay/MonoAGS/tree/IOS, and run DemoQuest.IOS.

<details>
<summary>This is the exception I got by doing that:</summary>

```
Autofac.Core.DependencyResolutionException: An error occurred during the activation of a particular registration. See the inner exception for details. Registration: Activator = AGSCamera (ReflectionActivator), Services = [AGS.API.ICamera], Lifetime = Autofac.Core.Lifetime.CurrentScopeLifetime, Sharing = None, Ownership = OwnedByLifetimeScope ---> Value cannot be null.
Parameter name: method (See inner exception for details.) ---> System.ArgumentNullException: Value cannot be null.
Parameter name: method
  at System.Linq.Expressions.Expression.Call (System.Reflection.MethodInfo method, System.Linq.Expressions.Expression arg0, System.Linq.Expressions.Expression arg1) [0x0004c] in /Library/Frameworks/Xamarin.iOS.framework/Versions/10.6.0.10/src/mono/mcs/class/dlr/Runtime/Microsoft.Scripting.Core/Ast/MethodCallExpression.cs:548 
  at Autofac.Core.Activators.Reflection.ConstructorParameterBinding.ConvertPrimitiveType (System.Linq.Expressions.Expression valueExpression, System.Type conversionType) [0x0003b] in <fdcd048585a94c00af1b381645caf165>:0 
[...]

Issue #782, opened 2016-08-11
System.SerializableAttribute appears in Autofac.dll


forX01:
I got conflict with autofac 4.0.0.0 for .net4.5.1 .
why there's a System.SerializableAttribute inside the librairie?

tillig:
Can you be more specific about the error message you're seeing? For example, if it's a build time error, what's the full error? If it's a runtime error, what's the exception with stack trace?

The attribute we have [is a polyfill](http://www.michael-whelan.net/porting-dotnet-framework-library-to-dotnet-core/) that allows us to mark exceptions with the `[Serializable]` attribute whilst still targeting .NET Core.

If you are seeing that attribute in the Autofac assembly you've referenced it's because your NuGet process executed incorrectly and has added a reference to the `lib/netstandard1.1/Autofac.dll` or `lib/uap10.0/Autofac.dll`. You will see the polyfill in those. If your project is targeting a full version of the .NET framework (e.g., 4.5.1, 4.6, etc.) then NuGet should add the reference to `lib/net451/Autofac.dll` which does _not_ contain that attribute.

Bad references can sometimes happen if you already have a NuGet package added to your project and you change your project's target framework after that. Sometimes you have to remove and re-add your packages to ensure the assembly references and bindings are set up correctly. It can also happen if you don't have the latest tooling for Visual Studio and NuGet. Especially within the last year with the .NET Core work and new .NET target framework moniker changes, using old tooling can inadvertently cause issues.

I tried setting up a standard .NET console app targeting .NET 4.5.1 and a dotnet CLI app (project.json) targeting .NET 4.5.1. In both cases I tried using the `[Serializable]` attribute and was unable to reproduce any issue.

Check your project references and verify things are referenced correctly. Also make sure you're up to date on your VS and NuGet patches. If you are positive that everything's in order, please post a reproduction of the issue somewhere that we can see and figure out.

forX01:
You maybe right, I will look tomorrow why there's a system.* into autofac librairie.
I try an empty project and there's no system.s* 

my problem was:
when I try myself to add serializable, it said that I have a conflict between autofac.system.serializableattribute and the original system.serializableattribute.
your dll contain a version of system.s*

forX01:
noget create a reference to netstandard1.1 instead of net451. just need to change manually and problem solved.

tank you

jahmai:
[...]

PR #1227, opened 2020-11-03
NET 5.0 Support

alistairjevans:
Ahead of the .NET 5.0 release, I've done an experimental upgrade, and proved some of the key new features.

When reviewing, it may be best to go commit-by-commit, due to the nullable warning fixes done in the first commit polluting the other changes.

I've:

- Resolved the new nullability warnings that arise from the runtime having the correct NRT markers.
- Added Tests for records support and init-only property support.
- Excluded the BinaryFormatter exception serialisation test on .NET 5.
- Added experimental support for App Trimming (https://devblogs.microsoft.com/dotnet/app-trimming-in-net-5/); if you use `RegisterType`, Autofac now instructs the linker to preserve public constructors of the specified type.

I'd like to tinker with adjustments to the AggregateService library that use Record types rather than custom interfaces, because I believe it will make our code simpler, and user's code more concise.


src/Autofac/Core/Activators/Reflection/ConstructorBinder.cs
```diff
@@ -154,7 +154,9 @@ public BoundConstructor Bind(IEnumerable<Parameter> availableParameters, ICompon
         private static MethodCallExpression ConvertPrimitiveType(Expression valueExpression, Type conversionType)
         {
             var changeTypeMethod = typeof(Convert).GetRuntimeMethod(nameof(Convert.ChangeType), new[] { typeof(object), typeof(Type) });
-            return Expression.Call(changeTypeMethod, valueExpression, Expression.Constant(conversionType));
+
+            // changeTypeMethod will always be non-null; Convert.ChangeType definitely exists.
```
tillig, line 158: Interestingly enough, at least in early linker trimming [`Convert.ChangeType` was _not_ definitely always there.](https://github.com/autofac/Autofac/issues/842)

src/Autofac/Features/Collections/CollectionRegistrationSource.cs
```diff
@@ -160,7 +161,9 @@ public override string ToString()
             var parameter = Expression.Parameter(typeof(int));
             var genericType = typeof(List<>).MakeGenericType(elementType);
             var constructor = genericType.GetMatchingConstructor(new[] { typeof(int) });
-            var newList = Expression.New(constructor, parameter);
+
+            // We know that List<> has the constructor we need.
+            var newList = Expression.New(constructor!, parameter);
```
tillig, line 166: Do we actually know that the `List<T>(int)` constructor is kept once linker trimming is enabled?

src/Autofac/RegistrationExtensions.AssemblyScanning.cs
```diff
@@ -19,12 +19,15 @@ namespace Autofac
     [SuppressMessage("Microsoft.Maintainability", "CA1506:AvoidExcessiveClassCoupling")]
     public static partial class RegistrationExtensions
     {
+        private const string AssemblyScanningWarning = "Assembly Scanning is unlikely to be compatible with member-level trimming; the linker will not be able to determine which types to preserve.";
```
tillig, line 22:
```suggestion
        private const string AssemblyScanningWarning = "Assembly scanning is unlikely to be compatible with member-level trimming; the linker will not be able to determine which types to preserve.";
```

tillig: I like that we can provide a warning. That's interesting.

src/Autofac/RegistrationExtensions.Generics.cs
```diff
@@ -25,7 +25,7 @@ public static partial class RegistrationExtensions
         /// <param name="implementer">The open generic implementation type.</param>
         /// <returns>Registration builder allowing the registration to be configured.</returns>
         public static IRegistrationBuilder<object, ReflectionActivatorData, DynamicRegistrationStyle>
-            RegisterGeneric(this ContainerBuilder builder, Type implementer)
+            RegisterGeneric(this ContainerBuilder builder, [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)] Type implementer)
```
tillig, line 28: I won't mark all of these, but it'd be interesting to consider how we end up handling custom `IConstructorFinder` implementations that potentially allow for non-public constructor usage. I wonder if we may have to keep _all_ constructors even if the use of non-public constructors is like the 99% case.

src/Autofac/RegistrationExtensions.cs
```diff
@@ -396,6 +396,7 @@ public static void RegisterComponent(this ContainerBuilder builder, IComponentRe
         /// <param name="registration">Registration to set policy on.</param>
         /// <param name="propertySelector">Policy to be used when searching for properties to inject.</param>
         /// <returns>A registration builder allowing further configuration of the component.</returns>
+        [RequiresUnreferencedCode("Autowired Properties with a custom selector may not be compatible with member-level trimming.")]
```
tillig, line 399:
```suggestion
        [RequiresUnreferencedCode("Autowired properties with a custom selector may not be compatible with member-level trimming.")]
```

src/Autofac/RegistrationExtensions.Generics.cs
tillig: Or maybe this is a case where we have to mark something around the use of the `IConstructorFinder` with that `[RequiresUnreferencedCode]` attribute to issue a warning. Hmmm.

src/Autofac/TypeExtensionsResources.resx
```diff
@@ -117,6 +117,9 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="MemberNotFound" xml:space="preserve">
+    <value />
```
tillig, line 121: Whoops! This is a bit too literal, I think, for the `MemberNotFound` message.

src/Autofac/Util/LinkerAttributes.cs
```diff
@@ -0,0 +1,143 @@
+﻿// Copyright (c) Autofac Project. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#pragma warning disable SA1649 // File name should match first type name
+#pragma warning disable SA1402 // File may only contain a single type
+#if !NET5_0
```
tillig, line 6: We'll have to be careful about this as we add/remove TFMs to the project. We did something similar with `[Serializable]` [back in the early netstandard days](https://github.com/autofac/Autofac/issues/782) and when there were NuGet bugs such that the build was referencing the wrong TFM folks got "duplicate attribute" errors. It was pretty hard to track down. If a .NET 5 project accidentally ends up referencing (directly or indirectly) the `netstandard2.0` version of Autofac, for example, then someone might see that error.

tillig (changes requested):
Looking good. A couple of potential gotchas that we've hit before that you may want to consider.

I can see how the nullability stuff could be a separate PR but I'm not _that_ concerned about it.

src/Autofac/Util/LinkerAttributes.cs
alistairjevans: For this one I sort of followed the same pattern as the [NullableAttributes](https://github.com/autofac/Autofac/blob/develop/src/Autofac/Util/NullableAttributes.cs) added for NRT compatibility shims.

I suppose there's an argument that I should use `#if NETSTANDARD2_0 || NETSTANDARD2_1` rather than `#if NET5_0` though.

alistairjevans:
As a general point regarding app trimming; my view on it is that any paths where we know we can preserve the things Autofac requires through attribute decoration, or just regular compiler behaviour, we mark with those attributes, and say you can use trimming.

Anything outside of that, we shouldn't even attempt to support app trimming, and should slap warnings on it and say "might work with app trimming, use at your own risk". Things like:

- Custom Constructor finders
- Assembly Scanning
- Custom Property Selectors
- Anything that results in a new limit type going into the registry where an explicit `Type` or generic type arg is not provided.

 The .NET guidance on this explicitly states that unit tests can't catch this stuff, and you just have to try out your application. I'd suggest we take the same approach, mark the trimming support as experimental (as it currently is in .NET5), but allow people to try it out. That way, people with Blazor apps can use Autofac Modules (for example), and Autofac will inform the trimmer that some things need to be kept.

In the future, we may need to go down the path of source generators to support things like `RegisterAssemblyTypes`, and build an Autofac Module behind the scenes that does that, but it's perhaps for another day...

tillig:
I think that stance on trimming is reasonable. "Best effort, sorry it's not perfect." Mental note that something to that effect should end up in docs so people aren't like, "How come this didn't work?"

src/Autofac/Util/LinkerAttributes.cs
tillig: I don't think it'd fix the problem, which is that a transient dependency chain not _quite_ working right would end up referencing the wrong framework's version of the assembly and the shims would show up in a .NET 5 project. But I think they're a lot further down the road with this than they were when we ran into it before, so let's leave it and see what happens. Just something to be aware of in case you see an issue about it.

test/Autofac.Specification.Test/Autofac.Specification.Test.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFrameworks>net5.0</TargetFrameworks>
```
tillig, line 4: Is there potentially value in multi-targeting the test projects so they run under the different frameworks? We did this for a bit when it was .NET 4.5 and netstandard because we had conditional tests. Now we have the shims that are conditional... but, then, the .NET 4.5 stuff was about partial trust and code access security crap, so maybe it doesn't make sense to repeat the same tests twice.

tillig (commented):
Looking good. Are you still messing with it or should I click the buttons and make it go?

test/Autofac.Specification.Test/Autofac.Specification.Test.csproj
alistairjevans: It also means our build system would need to change (which is why I switched back to a single target); currently we only install one version of the .NET SDK; we'd need to install two.

Thinking about it though, I did explicitly `#if` out a block of DependencyInjectionException tests to run on .NET5 (to do with `BinaryFormatter`, but now that test just won't run at all.

Do we need to amend our build system to install netcoreapp3.1 & net5.0 versions?

tillig: Yeah, I think we probably have to get that sorted. Once we start doing `#if` stuff to support different frameworks it seems like we should test both code paths out. It's one of the big reasons I hate `#if` conditional compilation; it makes everything harder to test and maintain. I was super glad to get rid of the code access security stuff because of that.

alistairjevans: OK; I've added a new option into the global.json (seemed like the best place for it) that specifies an older version of the SDK to install in addition to the latest. Tests are now running on both frameworks.

alsami: Just as a hint; When you do multi-target, then there are two different code-coverage files generated. Both have to be uploaded then AFAIK.

global.json
```diff
@@ -1,6 +1,10 @@
 {
   "sdk": {
-    "version": "3.1.302",
+    "version": "5.0.100-rc.2.20479.15",
     "rollForward": "latestFeature"
+  },
+
+  "previousSdk": {
+    "version": "3.1.302"
```
tillig, line 8: Should this allow more than one?

```json
{
  "additionalSdks": [
    "3.0.0",
    "3.1.0"
  ]
}
```

build.ps1
```diff
@@ -12,7 +12,12 @@ try {
 
     $artifactsPath = "$PSScriptRoot/artifacts"
     $packagesPath = "$artifactsPath/packages"
-    $sdkVersion = (Get-Content "$PSScriptRoot/global.json" | ConvertFrom-Json).sdk.version
+
+    $globalJson = (Get-Content "$PSScriptRoot/global.json" | ConvertFrom-Json);
```
tillig, line 16: Gotcha here if you end up making `previousSdk` into an array - you'll want `ConvertFrom-Json -NoEnumerate`. Might want that anyway. Ran into this literally yesterday - if the array in JSON has only one member, by default `ConvertFrom-Json` will convert the array to a string. Found this because `foreach` and `.Length` started iterating over the _letters_ in a value rather than just finding one value.

tillig (changes requested):
Potential gotcha in the build. Getting there.

build/Autofac.Build.psm1
```diff
@@ -36,8 +36,8 @@ function Install-DotNetCli {
     )
 
     if ($null -ne (Get-Command "dotnet" -ErrorAction SilentlyContinue)) {
-        $installedVersion = dotnet --version
-        if ($installedVersion -eq $Version) {
+        $installedVersions = dotnet --list-sdks
+        if ([bool]($installedVersions -match $Version)) {
```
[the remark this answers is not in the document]
alistairjevans: I knew it was too good to be true...

global.json
alistairjevans: Probably should be, yes; changing.

build.ps1
alistairjevans: That's just generally a handy tip, thanks.

test/Autofac.Specification.Test/Autofac.Specification.Test.csproj
alistairjevans: Thanks @alsami; fixing the coverage was something of a pain because of the multiple files as you said. If you had a few moments to check my changes are ok, that would be appreciated.

Basically had to end up putting the coverage artifacts in separate folders, because the `MergeWith` property can't be framework-specific.

alistairjevans:
I think we're all good now.  Matching on an array of additional SDKs is fixed, as is the code coverage with multiple TFMs.

tillig (approved):
🦠

src/Autofac/Core/Activators/Reflection/ConstructorBinder.cs
tillig: Again, just to draw attention to this - `Convert.ChangeType` wasn't there in the past; and the `List<T>(int)` constructor may suffer the same fate. I don't think there's any action to take, but just something of which to be aware.

tillig:
Interesting gotcha on the build and local SDK installs:

I had one (the .NET 5 RC) already installed. I did _not_ have exactly 3.1.302 installed. The build detected that I had .NET 5 RC installed so did not get a local copy; it saw that I didn't have 3.1.302 installed so it locally installed that... then during the build because the CLI had changed to the clone-local install, it couldn't find the .NET 5 RC anymore.

Basically, you can't mix and match "already installed" with "local install." It borks.

I can pretty easily work around it, but there are cases where this won't "just work." 😦

Issue #1226, opened 2020-11-02
Type must not be ByRef (Parameter 'type') exception when constructor contains `in` parameters


shatl:
## Describe the Bug

`System.ArgumentException` with message `Type must not be ByRef (Parameter 'type')` is thrown when building container if
 type constructor contains `in` parameter.

Same code did not throw any exception.

## Steps to Reproduce

```c#
    public readonly struct Input
    {
        public Input(string foo)
        {
            Foo = foo;
        }
        public string Foo { get; }
    }

    public class Service
    {
        public Service(in Input input)
        { }
    }

    public class UnitTest1
    {
        [Fact]
        public void CanBuildContainer()
        {
            var builder = new ContainerBuilder();
            builder.RegisterType<Service>();
            builder.Build();
        }
    }
```

## Expected Behavior

Looks like this is limitation of lambda. In that case, detailed error message with reference to type being processed (`Service` in this case) should be thrown.

## Exception with Stack Trace

```text

System.ArgumentException
Type must not be ByRef (Parameter 'type')
   at System.Dynamic.Utils.TypeUtils.ValidateType(Type type, String paramName, Boolean allowByRef, Boolean allowPointer)
   at System.Linq.Expressions.Expression.Convert(Expression expression, Type type, MethodInfo method)
   at Autofac.Core.Activators.Reflection.ConstructorBinder.GetConstructorInvoker(ConstructorInfo constructorInfo)
   at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
   at Autofac.Core.Activators.Reflection.ConstructorBinder..ctor(ConstructorInfo constructorInfo)
   at Autofac.Core.Activators.Reflection.ReflectionActivator.ConfigurePipeline(IComponentRegistryServices componentRegistryServices, IResolvePipelineBuilder pipelineBuilder)
   at Autofac.Core.Registration.ComponentRegistration.BuildResolvePipeline(IComponentRegistryServices registryServices, IResolvePipelineBuilder pipelineBuilder)
   at Autofac.Core.Registration.ComponentRegistration.BuildResolvePipeline(IComponentRegistryServices registryServices)
   at Autofac.Core.Registration.ComponentRegistryBuilder.Build()
   at Autofac.ContainerBuilder.Build(ContainerBuildOptions options)
   at AutofacInParameter.UnitTest1.CanBuildContainer()

```

## Dependency Versions

Autofac: 6.0.0
[...]

PR #1228, opened 2020-11-04
fix: make sure that constructor-binder detects illegal parameter when parameter is by ref

alsami:
If I understood correctly, this should fix the issue described in #1226


tillig (approved):
💯

PR #1231, opened 2020-11-24
chore: update build-scripts and module

alsami:


tillig (approved):
I think we missed an indent on the last line of `Add-Path`. I'll fix it after merge.

Issue #215, opened 2014-01-22
Support registering open generic types when scanning assemblies


alexmg:
_From [lanwin...@gmail.com](https://code.google.com/u/114531279769120480412/) on April 29, 2010 22:11:22_

Currently I register all my ViewModels this way:

```
builder.RegisterAssemblyTypes(typeof(App).Assembly)
  .InstancePerDependency()
  .AsImplementedInterfaces()
  .AsSelf()
  .AssignableTo<ViewModelBase>();
```

But this only works for non generic versions. I have some generic
`ViewModel<T>` which are only resolved `AsSelf`( `Resolve<ViewModel<string>>()`
). But with the above can not resolve this. The ViewModels have no
OpenGeneric base type.

_Original issue: http://code.google.com/p/autofac/issues/detail?id=215_

MehdiSaeedifar:
any progress?

tillig:
Nope.

MehdiSaeedifar:
So what is equivalent to the following code if i want to use assembly scanning?

`builder.RegisterGeneric(typeof(NHibernateRepository<>))
       .As(typeof(IRepository<>))
       .InstancePerLifetimeScope();`

alexeyzimarev:
@MehdiSaeedifar:

```
builder.RegisterAssemblyTypes(assemblies)
    .Where(t => t.GetTypeInfo()
        .ImplementedInterfaces.Any(
            i => i.IsGenericType && i.GetGenericTypeDefinition() == typeof (NHibernateRepository<>)))
    .AsImplementedInterfaces()
    .InstancePerLifetimeScope():
```

MehdiSaeedifar:
@alexeyzimarev  thank you.

nphmuller:
I think the 4.0 milestone can be removed ;)

nphmuller:
@alexeyzimarev 
Your solution didn't work for me when the assembly type also contained generics. (MyType<,> instead of MyType)
It probably doesn't work because RegisterGenerics adds an registration source and RegisterAssemblyTypes probably only adds real registrations. (ComponentRegistry.Sources vs .Registrations).

I'm trying to write an extension method which adds an registration source in RegisterAssemblyTypes but so far it's pretty hacky...

dasiths:
@alexeyzimarev @nphmuller `RegisterAssemblyTypes` won't pipe generic types to the where clause.

This is the work around

```C#
            var middlewareTypes = assembly.GetTypes().Where(t =>
[...]

PR #1232, opened 2020-11-29
Fix issue 215

RaymondHuy:
Issue https://github.com/autofac/Autofac/issues/215 has been lasted for 6 years so I would like to try my self fixing it. :grin:
My approach is to use a new extension method for registering open generic types when scanning assemblies following @tillig 's suggestion. This PR hasn't been finished yet, there are still some functions I haven't implemented yet like `AsImplementedInterfaces(), Named(), Keyed(), Metadata()`. Adding those functions at this time will make this PR too large to review so I will add them later when we accept this approach and also easily reviewing PR.

### DIFF
--- a/src/Autofac/Features/Scanning/OpenGenericScanningActivatorData.cs
+++ b/src/Autofac/Features/Scanning/OpenGenericScanningActivatorData.cs
@@ -1,0 +1,40 @@
+﻿// Copyright (c) Autofac Project. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Autofac.Builder;
+using Autofac.Core.Registration;
+
+namespace Autofac.Features.Scanning
+{
+    /// <summary>
+    /// Activation data for open generic types located by scanning assemblies.
+    /// </summary>
+    public class OpenGenericScanningActivatorData : ReflectionActivatorData
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="OpenGenericScanningActivatorData"/> class.
+        /// </summary>
+        public OpenGenericScanningActivatorData()
+            : base(typeof(object))
+        {
+        }
+
+        /// <summary>
+        /// Gets the filters applied to the types from the scanned assembly.
+        /// </summary>
+        public ICollection<Func<Type, bool>> Filters { get; } = new List<Func<Type, bool>>();
+
+        /// <summary>
+        /// Gets the additional actions to be performed on the concrete type registrations.
+        /// </summary>
+        public ICollection<Action<Type, IRegistrationBuilder<object, ReflectionActivatorData, DynamicRegistrationStyle>>> ConfigurationActions { get; }
+            = new List<Action<Type, IRegistrationBuilder<object, ReflectionActivatorData, DynamicRegistrationStyle>>>();
+
+        /// <summary>
+        /// Gets the actions to be called once the scanning operation is complete.
+        /// </summary>
+        public ICollection<Action<IComponentRegistryBuilder>> PostScanningCallbacks { get; } = new List<Action<IComponentRegistryBuilder>>();
+    }
+}
--- a/src/Autofac/Features/Scanning/OpenGenericScanningActivatorData.cs
+++ b/src/Autofac/Features/Scanning/OpenGenericScanningActivatorData.cs
@@ -4,37 +4,20 @@
 using System;
 using System.Collections.Generic;
 using Autofac.Builder;
-using Autofac.Core.Registration;
 
 namespace Autofac.Features.Scanning
 {
     /// <summary>
     /// Activation data for open generic types located by scanning assemblies.
     /// </summary>
-    public class OpenGenericScanningActivatorData : ReflectionActivatorData
+    public class OpenGenericScanningActivatorData : BaseScanningActivatorData<ReflectionActivatorData, DynamicRegistrationStyle>
     {
         /// <summary>
         /// Initializes a new instance of the <see cref="OpenGenericScanningActivatorData"/> class.
         /// </summary>
         public OpenGenericScanningActivatorData()
-            : base(typeof(object))
+            : base(new List<Action<Type, IRegistrationBuilder<object, ReflectionActivatorData, DynamicRegistrationStyle>>>())
         {
         }
-
-        /// <summary>
-        /// Gets the filters applied to the types from the scanned assembly.
-        /// </summary>
-        public ICollection<Func<Type, bool>> Filters { get; } = new List<Func<Type, bool>>();
-
-        /// <summary>
-        /// Gets the additional actions to be performed on the concrete type registrations.
-        /// </summary>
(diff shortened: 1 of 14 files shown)


test/Autofac.Test/Features/Scanning/OpenGenericScanningRegistrationTests.cs
```diff
@@ -0,0 +1,102 @@
...
+
+            c.AssertRegistered<RedoOpenGenericCommand<int>>();
+        }
+
+        [Fact]
+        public void WhenFiltersAppliedNonMatchingOpenGenericTypesExcluded()
+        {
+            var cb = new ContainerBuilder();
+            cb.RegisterAssemblyOpenGenericTypes(typeof(AComponent).GetTypeInfo().Assembly)
+                .Where(t => t.Name.StartsWith("Re"))
```
tillig, line 53: I'm not sure why there are two `Where` clauses here that overlap. Probably just need one.

tillig:
I think this is as good a solution as any. Nothing internally has changed that I'm aware of that would make the approach here any different. I haven't had a chance to dig into this in detail yet.

I do see a bit of redundant code, like `OpenGenericScanningRegistrationExtensions` has some near copy-paste private methods where it's either the type of a parameter or maybe one to two lines that are different. I wonder if there's any opportunity to reduce some of that duplication, especially in areas like the `Where` clause in the `ScanTypes` method where it only differs by one `!` and was part of the cause of an issue before. If we have another issue, for example, it'd be nice to not have to fix that in two places.

@alexmg has been around for a while and also might have some input.

test/Autofac.Test/Features/Scanning/OpenGenericScanningRegistrationTests.cs
RaymondHuy: The reason for there are two `Where` clauses is that I follow the same test case in this file: https://github.com/autofac/Autofac/blob/develop/test/Autofac.Test/Features/Scanning/ScanningRegistrationTests.cs#L76

tillig: That's just... weird. I don't know why that'd make any difference. I think in this one we can remove the first `Where` clause.

RaymondHuy: Thanks @tillig , just updated my PR to resolve it.

src/Autofac/Util/TypeExtensions.cs
```diff
@@ -182,6 +182,45 @@ public static bool IsGenericTypeDefinedBy(this Type @this, Type openGeneric)
                     && key.Item1.GetGenericTypeDefinition() == key.Item2);
         }
 
+        /// <summary>
+        /// Checks whether this type is an open generic type of a given type.
+        /// </summary>
+        /// <param name="this">The type we are checking.</param>
+        /// <param name="type">The open generic type to validate against.</param>
+        /// <returns>True if <paramref name="this"/> is a closed type of <paramref name="type"/>. False otherwise.</returns>
+        public static bool IsOpenGenericTypeOf(this Type @this, Type type)
```
tillig, line 191: I think I got confused by this method because `this` is a closed generic and `type` is the open generic, so the usage would be like `typeof(IGeneric<string>).IsOpenGenericTypeOf(typeof(IGeneric<>)`... but that's backwards for the name. It should be `IsClosedGenericTypeOf` if it's that direction, right? `X.IsClosedGenericTypeOf(openGeneric)`

It would be really good to get some unit tests on this. I don't know if we have too many tests for these sorts of utilities, but for stuff like `IsAssignableFrom` vs `IsAssignableTo` it's always hard to remember the direction it operates, so seeing unit tests helps explain the intent.

It might also help in refactoring. For example, I see on line 198 that we're already checking if `@this == type` but then on line 203 there's the same thing. It's nice to be able to safely refactor and trust the tests will catch issues.

tillig (changes requested):
I think this is a reasonable general solution. If we can refactor out some of the duplicate code I think we can pull it in and solve this thing.

src/Autofac/Util/TypeExtensions.cs
RaymondHuy: Thanks @tillig  I've just refactored and added tests for this method, could you check it ?

alistairjevans: @RaymondHuy; just passing through, but I wonder if there should be some negative tests for `IsOpenGenericTypeOf` as well as positive ones?

src/Autofac/Features/Scanning/OpenGenericScanningActivatorData.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Collections.Generic;
+using Autofac.Builder;
+using Autofac.Core.Registration;
+
+namespace Autofac.Features.Scanning
+{
+    /// <summary>
+    /// Activation data for open generic types located by scanning assemblies.
+    /// </summary>
+    public class OpenGenericScanningActivatorData : ReflectionActivatorData
```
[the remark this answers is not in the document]
alistairjevans: Yes, it seems sensible to consolidate into a single generic base class for ScanningActivatorData.  I can't imagine there would be many people depending on `ScanningActivatorData` directly; if they are, the members and names of that type won't change, just the inheritance hierarchy.

src/Autofac/Features/Scanning/OpenGenericScanningRegistrationExtensions.cs
```diff
@@ -0,0 +1,192 @@
...
+
+            return rb;
+        }
+
+        private static void ScanAssemblies(IEnumerable<Assembly> assemblies, IComponentRegistryBuilder cr, IRegistrationBuilder<object, OpenGenericScanningActivatorData, DynamicRegistrationStyle> rb)
+        {
+            ScanTypes(assemblies.SelectMany(a => a.GetLoadableTypes()), cr, rb);
+        }
+
+        private static void ScanTypes(IEnumerable<Type> types, IComponentRegistryBuilder cr, IRegistrationBuilder<object, OpenGenericScanningActivatorData, DynamicRegistrationStyle> rb)
```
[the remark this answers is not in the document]
RaymondHuy: Thanks @tillig @alistairjevans I've just refactored this function.

src/Autofac/Util/TypeExtensions.cs
RaymondHuy: Thanks @alistairjevans  I've just added negative tests for it but not sure is it enough to cover ? 😕

alistairjevans: Looks like enough, thanks!

tillig (approved):
Looks great! It's nice to have this resolved after so much time.

tillig:
I'll hold off on a release until we have the rest of the `AsImplementedInterfaces` and others ready to go, as noted in the original PR post at the top.

RaymondHuy:
Ok @tillig Maybe it takes about 2-3 weeks for the PR of remaining features because I'm going to have a vacation for this Christmas. 😄

PR #1234, opened 2020-11-30
Disposer, ability to track objects using WeakReference.

### LINKED ISSUE #1233: Disposer prevents objects from being garbage collected
## Describe the Bug

AutoFac supports automatic disposal of objects when the ILifetimeScope is disposed. In order to do so, disposable objects are added to the lifetime scope's Disposer object, which keeps reference to these objects in a `Stack<object>`.  When the lifetime scope is disposed, the Disposer pops each object from its stack and disposes it. 

The problem is, this mechanism helps dispose the objects when the lifetime is disposed, but also _prevents_ them from being garbage collected before the lifetime scope is disposed.  A client might resolve a disposable object and dispose it immediately when they're done with it.  However, because Disposer still has a reference to the object, it cannot be garbage collected until the lifetime scope is disposed.

Ideally, the lifetime scope should auto-dispose items like it's doing today, but at the same time, it should not prevent objects from being garbage collected if nothing else is holding a reference to them.  

## Steps to Reproduce

```c#
public class ReproTest
{
	private class MyDisposable : IDisposable
	{
		public Action OnFinalize;

		~MyDisposable() => OnFinalize();

		public void Dispose() { }
	}

	private static void DoWork(ILifetimeScope container, Action onCompleted)
	{
		var obj = container.Resolve<MyDisposable>();
		obj.OnFinalize = onCompleted;
	}

	[Fact]
	public void Repro()
	{
		builder.RegisterType<MyDisposable>();
		bool finalized = false;

		DoWork(container, () => finalized = true);
		GC.Collect();
		GC.WaitForPendingFinalizers();
		Assert.False(finalized);  // object is still alive because the container holds a reference to it

		container.Dispose();
		GC.Collect();
		GC.WaitForPendingFinalizers();
		Assert.True(finalized);  // object is now finalized because disposing the container removed the ref
	}
}
```

eersonmez:
fixed #1233

### DIFF
--- a/src/Autofac/Autofac.csproj
+++ b/src/Autofac/Autofac.csproj
@@ -31,6 +31,7 @@
     <Authors>Autofac Contributors</Authors>
     <Company>Autofac</Company>
     <Product>Autofac</Product>
+    <Version>6.0.0</Version>
   </PropertyGroup>
 
   <!-- Disable nullability warnings in netstandard2.0 -->
--- a/src/Autofac/Autofac.csproj
+++ b/src/Autofac/Autofac.csproj
@@ -31,7 +31,6 @@
     <Authors>Autofac Contributors</Authors>
     <Company>Autofac</Company>
     <Product>Autofac</Product>
-    <Version>6.0.0</Version>
   </PropertyGroup>
 
   <!-- Disable nullability warnings in netstandard2.0 -->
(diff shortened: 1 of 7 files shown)


tillig:
Per discussion in #1233, we're not moving to `WeakReference` here.

Issue #1211, opened 2020-10-08
Ability to modify the registration pipeline from a module or other external mechanism


xumix:
## Problem Statement
It is currently unclear how to add some Aspects to components if the component type is not known at the registration time.
For example some aspects like Logging and Caching are perfect for this. 
Is there any way to conditionally enable interception after the registration?

So I want to add caching for my methods in all components if some conditions are met, like Attribute is used on method or type.

Current api for this:
```
builder.RegisterType<SomeAdapter>().As<ISomeAdapter>().EnableClassInterceptors().InterceptedBy(typeof(CacheInterceptor)).SingleInstance();

```

## Desired Solution
Add ability to modify registration in some pipeline/add interceptors after component registration.
There is no way to enable interception for some particular components in some common core module
The desired api would look like this:

```
public class AspectSupportModule : Module
    {
       // Not sure if this is the right place, maybe it is already possible to do this in another method?
        protected override void AttachToComponentRegistration(
            IComponentRegistryBuilder componentRegistry,
            IComponentRegistration registration)
        {
             if (someCondition)
{
    registration.EnableClassInterceptors();
}
        }
```

## Alternatives You've Considered
As of now, I added my custom module to intercept object activation but this solution does not look very robust

```
public class AspectSupportModule : Module
    {
        private static readonly Dictionary<Type, Type> Aspects = new Dictionary<Type, Type>
        {
            {
                typeof(LoggedAttribute), typeof(LoggingInterceptor)
            },
            {
                typeof(CachedAttribute), typeof(CacheInterceptor)
            }
        };

        private readonly ProxyGenerator generator;

        public AspectSupportModule()
        {
            generator = new ProxyGenerator();
        }

        /// <inheritdoc />
        protected override void Load(ContainerBuilder builder)
        {
            base.Load(builder);
            foreach (var aspect in Aspects)
            {
                builder.RegisterType(aspect.Value);
            }
        }

        protected override void AttachToComponentRegistration(
            IComponentRegistryBuilder componentRegistry,
            IComponentRegistration registration)
        {
            var type = registration.Activator.LimitType;

            var aspects = GetInterceptors(type).ToArray();
            if (aspects.Length == 0)
            {
                return;
            }

            var ctors = type.GetConstructors();
            var infos = ctors.First().GetParameters();

            // Attach to the registration's pipeline build.
            registration.PipelineBuilding += (sender, pipeline) =>
            {
                pipeline.Use(PipelinePhase.Activation,
[...]

PR #1236, opened 2020-12-05
Add ConfigurePipeline to IComponentRegistration

alistairjevans:
As discussed in https://github.com/autofac/Autofac/issues/1211#issuecomment-739160937, I'm adding ConfigurePipeline to IComponentRegistration, as a convenience wrapper for the PipelineBuilding event.

This should make the modification of a registration's pipeline simpler to understand.

Also identified an unlikely but possible memory leak due to holding on to an event handler longer than necessary, which I have fixed here as well.

### DIFF
--- a/src/Autofac/Core/Registration/ComponentRegistration.cs
+++ b/src/Autofac/Core/Registration/ComponentRegistration.cs
@@ -20,6 +20,8 @@ namespace Autofac.Core.Registration
     {
         private readonly IComponentRegistration? _target;
         private readonly IResolvePipelineBuilder _lateBuildPipeline;
+        private EventHandler<IResolvePipelineBuilder>? _pipelineBuildEvent;
+
         private IResolvePipeline? _builtComponentPipeline;
 
         /// <summary>
@@ -200,25 +202,22 @@ namespace Autofac.Core.Registration
         public RegistrationOptions Options { get; }
 
         /// <inheritdoc />
-        public event EventHandler<IResolvePipelineBuilder>? PipelineBuilding;
-
-        /// <inheritdoc />
-        public void ConfigurePipeline(Action<IResolvePipelineBuilder> configurationAction)
+        public event EventHandler<IResolvePipelineBuilder>? PipelineBuilding
         {
-            if (configurationAction is null)
+            add
             {
-                throw new ArgumentNullException(nameof(configurationAction));
+                if (_builtComponentPipeline is object)
+                {
+                    throw new InvalidOperationException(ComponentRegistrationResources.PipelineAlreadyBuilt);
+                }
+
+                _pipelineBuildEvent += value;
             }
 
-            if (_builtComponentPipeline is object)
+            remove
             {
-                throw new InvalidOperationException(ComponentRegistrationResources.PipelineAlreadyBuilt);
+                _pipelineBuildEvent -= value;
             }
-
-            PipelineBuilding += (sender, pipeline) =>
-            {
-                configurationAction(pipeline);
-            };
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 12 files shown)


src/Autofac/Core/Registration/ComponentRegistration.cs
```diff
@@ -202,6 +202,25 @@ public class ComponentRegistration : Disposable, IComponentRegistration
         /// <inheritdoc />
         public event EventHandler<IResolvePipelineBuilder>? PipelineBuilding;
 
+        /// <inheritdoc />
+        public void ConfigurePipeline(Action<IResolvePipelineBuilder> configurationAction)
```
tillig, line 206: Would it be valuable to make this an extension method instead of something right on the interface? The only thing it actually needs that isn't already public is a way to determine if the pipeline has already been built; and given the very few times this is likely to be called, maybe that's just a `catch(InvalidOperationException)` instead of an `if (x is object)` check.

Alternatively, perhaps that informs the design of the interface in that there may need to be a better public way to check if the pipeline is built, so other folks _can_ write their own extension methods like this.

tillig (changes requested):
Design question, otherwise it looks good. Definitely opens the door for some easier pipeline configuration.

alsami:
One question not really related to this PR but how did the coverage drop so much?

alistairjevans:
Did it change when we merged the new build changes for multiple SDKs? 

I know I had to make changes in the original .NET 5 PR to make codecov absorb both platform's coverage.

alsami:
Something else must have happened. The asp net core integration uses pretty much the same logic now. I'll look into it the next days. I highly doubt that 35 percent is the current number hehe

src/Autofac/Core/Registration/ComponentRegistration.cs
alistairjevans: Defining an extension method does make sense; also prevents the "technically" breaking change on `IComponentRegistration`.

The thing on my mind about exposing the pipeline state is that attempting to configure the pipeline for a registration after the pipeline has been built is by it's nature an exceptional action; it would only happen if the developer has made a mistake. 

Either the developer:
- Attaches to the right point, inside a module, before the container build has finished, so they're good, well-behaved and they get presents for Christmas, or...
- Naughty developers try to reach into the component registry while their code is running to try and bypass the immutable container, at which point we wag our finger at them and put coal in their stocking.

Basically, someone trying to write an extension that can react differently to a built or not-built pipeline is probably approaching things the wrong way.

alistairjevans: Okay, so trying to make it an extension method instead, but the problem is that if we rely on an exception being raised from the `ResolvePipeline` get, it means `ConfigurePipeline` is only valid to be called **if** ResolvePipeline throws an exception, so we'd be potentially getting a lot of exceptions raised.

I think this may have to stay on `ComponentRegistration`...while it is technically a breaking change, I can't imagine many people have created their own `IComponentRegistration` implementation.

tillig: > it means `ConfigurePipeline` is only valid to be called if `ResolvePipeline` throws an exception

True, but that's also sort of the semantics around wrapping a `TryDoSomething` around a `DoSomething` method sometimes - the point might be to swallow the exception. Since `ConfigurePipeline` is only during registration time, is the throwing of the exceptions internally that big of a deal?

I'm not glued to the extension method thing, just thinking most of the time we've exposed enough on our interfaces for folks to write this sort of method themselves using extension methods; and all this does is add a convenience wrapper around an event that's already exposed. If someone _doesn't_ use `ConfigurePipeline` and instead goes directly to the `PipelineBuilding` event, there are other problems like "I added an event too late and it wasn't considered in the pipeline build."

Almost makes me wonder if the event handler for `PipelineBuilding` should be where the check for pipeline state is so if someone adds to the event too late they won't wonder why. There's no other way to really see if the pipeline has already been built.

Unless we add a convenience `PipelineState` property (`Built` or `Pending`?), which you'd need for attaching to the `PipelineBuilding` event anyway, at which point `ConfigurePipeline` could use that and it'd address the whole issue.

But like I said, I don't feel terribly strongly; if you do, say so and I'll just close this and call it a day.

alistairjevans: Okay; the fact that `ConfigurePipeline` throws, but `PipelineBuilding += {something}` does not feels wrong. So, modified to throw inside the `add` handler of `PipelineBuilding`, with an extension method to wrap it.

tillig (approved):
I dig it! 🦄

PR #1239, opened 2020-12-11
Fix #1238 - Only consider instance constructors

### LINKED ISSUE #1238: Avoid accidentally getting static and public constructors
Autofac use `DeclaredOnlyPublicFlag` to get the type constructors, But it may get a static & public constructor. However, `Expression.New` does not support static constructors.

https://github.com/autofac/Autofac/blob/e662b6bace37a569eec1e42335336b3fe015855c/src/Autofac/TypeExtensions.cs#L19-L20
https://github.com/autofac/Autofac/blob/e662b6bace37a569eec1e42335336b3fe015855c/src/Autofac/Core/Activators/Reflection/ConstructorBinder.cs#L148

#### Why does this happen?
If I use `Castle.DynamicProxy` to create a proxy class in `Blazor webassembly app` there will be a public and static constructor in an instance.

I think the problem is not Autofac but Castle. But maybe we can move `BindingFlags.Static` from `DeclaredOnlyPublicFlags` to `DeclaredOnlyFlags`:

```cs
private const BindingFlags DeclaredOnlyPublicFlags = BindingFlags.Public | BindingFlags.Instance | BindingFlags.DeclaredOnly;
private const BindingFlags DeclaredOnlyFlags = DeclaredOnlyPublicFlags | BindingFlags.Static | BindingFlags.NonPublic;
```

```cs
@page "/index"
@using Castle.DynamicProxy
@using System.Reflection

<h1>Hello Blazor</h1>

@code {
    public class MyClass
    {

    }

    protected override async Task OnInitializedAsync()
    {
        var proxyGenerator = new ProxyGenerator();

        foreach (var constructorInfo in proxyGenerator.CreateClassProxyWithTarget(new MyClass())
            .GetType()
            .GetConstructors(BindingFlags.Public | BindingFlags.Static | BindingFlags.Instance | BindingFlags.DeclaredOnly))
        {
            if (constructorInfo.IsPublic && constructorInfo.IsStatic)
            {
                // Will output!
                Console.WriteLine("constructor is public and static");
            }
        }

        await Task.CompletedTask;
    }
}

```

tillig:
#1238 shows that when we query types for constructors, usually when building reflection activators, we're inadvertently getting static constructors along with instance constructors. Since we can't create an instance of a type based on a static constructor, this PR removes the `BindingFlags.Static` when searching for constructors.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,5 @@
 {
+  "dotnet-test-explorer.testProjectPath": "test/**/*Test.csproj",
   "omnisharp.enableEditorConfigSupport": true,
   "omnisharp.enableRoslynAnalyzers": true
 }
--- a/test/Autofac.Test/TypeExtensionsTests.cs
+++ b/test/Autofac.Test/TypeExtensionsTests.cs
@@ -62,7 +62,7 @@ namespace Autofac.Test
         [Fact]
         public void IsInNamespaceOf_DifferentNamespace()
         {
-            Assert.False(typeof(TypeExtensionsTests).IsInNamespaceOf<ContainerBuilder>());
+            Assert.False(typeof(ContainerBuilder).IsInNamespaceOf<TypeExtensionsTests>());
         }
 
         [Theory]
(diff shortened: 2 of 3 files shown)


.vscode/settings.json
```diff
@@ -1,4 +1,5 @@
 {
+  "dotnet-test-explorer.testProjectPath": "test/**/*Test.csproj",
```
tillig, line 2: This helps avoid showing the benchmark tests in the VS Code test explorer so you can just "run all tests" without having to worry about eating the next 45 minutes with benchmarks.

test/Autofac.Test/TypeExtensionsTests.cs
```diff
@@ -41,5 +42,176 @@ public void AnOpenGenericTypeIsNotAClosedTypeOfAnything()
         {
             Assert.False(typeof(CommandBase<>).IsClosedTypeOf(typeof(CommandBase<>)));
         }
+
```
tillig, line 45: We didn't really have any test coverage specifically around these extensions so I added it to ensure the update to the constructor search flags didn't break anything else.

alistairjevans (approved):
Looks great! 🎢

PR #1240, opened 2020-12-12
fix: make sure code-coverage is collected correctly

alsami:


alistairjevans:
Side note, it was [this commit](https://github.com/autofac/Autofac/commit/a99be1ddd7435a945cc9af5c9d195435efc06eb8) that caused the codecov to stop collecting. Looks like turning on ContinuousIntegrationBuild broke the code coverage somehow.

alsami:
Yeah just noticed, we get "NaN%" now as the result. Still excluding the scanned-assemblies project is the right thing to do, since it does not contain code that should be covered directly.

alistairjevans:
Ah, `NaN`, my old friend.

alsami:
Okay looks like this is a [bug from coverlet](https://github.com/coverlet-coverage/coverlet/issues/900) itself. I've added their nightly build feed in order to install the latest versions (they have both been consolidated to 3.0.0-x). 

That seems to work well with the [attribute in place](https://ci.appveyor.com/project/Autofac/autofac/builds/36799062).

alistairjevans (approved):
Looks great @alsami, thanks for cracking that one. 

We need to make a note somewhere to upgrade to the stable coverlet when it gets released.

alsami:
> We need to make a note somewhere to upgrade to the stable coverlet when it gets released.

I'll create a ticket for that and assign it to me for now.

PR #1246, opened 2021-01-23
Open generic scanning extensions

RaymondHuy:
Add remaining methods as discussed in https://github.com/autofac/Autofac/pull/1232.
Some notes:
- `AsClosedTypesOf` is renamed to `AsOpenTypesOf`
- I think we shouldn't support `AssignableTo` because https://stackoverflow.com/questions/5461295/using-isassignablefrom-with-open-generic-types#:~:text=The%20difference%20is%20that%20open,%22assignable%22%20to%20the%20other.

### DIFF
--- a/test/Autofac.Test/Features/Scanning/OpenGenericScanningRegistrationTests.cs
+++ b/test/Autofac.Test/Features/Scanning/OpenGenericScanningRegistrationTests.cs
@@ -10,6 +10,7 @@ using Autofac.Core;
 using Autofac.Core.Lifetime;
 using Autofac.Core.Registration;
 using Autofac.Features.Metadata;
+using Autofac.Features.OpenGenerics;
 using Autofac.Features.Scanning;
 using Autofac.Test.Scenarios.ScannedAssembly;
 using Autofac.Test.Scenarios.ScannedAssembly.MetadataAttributeScanningScenario;
@@ -117,6 +118,30 @@ namespace Autofac.Test.Features.Scanning
 
             Assert.Throws<ArgumentNullException>(() => cb.RegisterAssemblyOpenGenericTypes(typeof(ICommand<>).GetTypeInfo().Assembly)
                 .AssignableTo(null, t => t));
+        }
+
+        [Theory]
+        [InlineData(typeof(ICloseCommand))]
+        [InlineData(typeof(CloseCommand))]
+        public void AssignableToClosedTypeProvidedNoneOpenGenericSourceRegistered(Type closedType)
+        {
+            var cb = new ContainerBuilder();
+            cb.RegisterAssemblyOpenGenericTypes(typeof(ICommand<>).GetTypeInfo().Assembly)
+                .AssignableTo(closedType);
+            var c = cb.Build();
+
+            Assert.False(c.RegisteredAnyOpenGenericTypeFromScanningAssembly());
+        }
+
+        [Fact]
+        public void ServiceIsNotAssignableToIsNotRegistered()
+        {
+            var cb = new ContainerBuilder();
+            cb.RegisterAssemblyOpenGenericTypes(typeof(ICommand<>).GetTypeInfo().Assembly)
+                .AssignableTo(typeof(RedoOpenGenericCommand<>));
+            var c = cb.Build();
+
+            Assert.Throws<ComponentNotRegisteredException>(() => c.Resolve<DeleteOpenGenericCommand<int>>());
         }
 
         [Fact]
--- a/test/Autofac.Test/Features/Scanning/OpenGenericScanningRegistrationTests.cs
+++ b/test/Autofac.Test/Features/Scanning/OpenGenericScanningRegistrationTests.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Reflection;
 using Autofac.Core;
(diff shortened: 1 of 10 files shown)


alistairjevans:
What is the motivation behind the name AsOpenTypesOf? Are you able to elaborate on it's purpose? What is an open type of an already open generic?

Regarding 'AssignableTo', are you simply saying that it's not possible because one generic type is not assignable to another? I think in terms of the strict implementation of the runtime's `AssignableFrom`, that's true, but I believe there are other places in the codebase where we say 'if I were to create a closed type of `X`, is that assignable to `Y`, closed with the same type'?. I'm not saying we absolutely need it for open generic scanning, but I'm curious if it's still possible.  It may be that because we can't know about all the type constraints at registration time, we can't implement it, and if so, that's OK by me.

alistairjevans:
Unrelated; @alsami, did we stabilise the code coverage metrics already? There's a big drop in this PR for what isn't really a huge quantity of code.

alsami:
> Unrelated; @alsami, did we stabilise the code coverage metrics already? There's a big drop in this PR for what isn't really a huge quantity of code.

No, still have to look into it. Will do so soon. Really don't know why it does it again, maybe we will need to downgrade to the pre-release again. Will figure it out.

RaymondHuy:
Hi @alistairjevans `AsOpenTypesOf` is equivalent with `AsClosedTypeOf` term (when scanning assembly) but used for open-generic scanning because we don't register closed type in this context. For example:
```
// RedoOpenGenericCommand<T> is an open type of ICommand<T>
public class RedoOpenGenericCommand<T> : ICommand<T> {}

cb.RegisterAssemblyOpenGenericTypes(typeof(ICommand<>).GetTypeInfo().Assembly)
    .AsOpenTypesOf(typeof(ICommand<>));
var c = cb.Build();

Assert.NotNull(c.Resolve<RedoOpenGenericCommand<int>>());
```

alistairjevans:
@RaymondHuy, I think it's just the name that of the method that may need a rethink. `AsClosedTypesOf` requires you to provide an open type as a parameter, but `AsOpenTypesOf` doesn't require you to provide a closed type? Seems a bit...unexpected. 
What happens if I was to pass a closed type into that method, e.g.:

```csharp
// RedoOpenGenericCommand<T> is an open type of ICommand<T>
public class RedoOpenGenericCommand<T> : ICommand<T> {}

cb.RegisterAssemblyOpenGenericTypes(typeof(ICommand<>).GetTypeInfo().Assembly)
    .AsOpenTypesOf(typeof(ICommand<int>));
var c = cb.Build();
```

I just feel like the user expectation of the name `AsOpenTypesOf` would expect to pass a closed type.

Haven't you kind of implicitly created an open generic `AssignableTo` method here?

RaymondHuy:
@alistairjevans  yah, maybe somehow I have self implemented the `AssignableTo` method but I don't know. Therefore how about the `AsClosedTypeOf` method, do you think it shouldn't exist when scanning Open Generic types ?

alistairjevans:
Yeah, if you call `RegisterAssemblyOpenGenericTypes`, you'd implicitly expect to not be able to register closed types. Closed Types != Open Generic, so it wouldn't make much sense. Can we constrain `AsClosedTypesOf` appropriately?

RaymondHuy:
@alistairjevans Yes, currently we can't use the method `AsClosedTypesOf` if it is inside the Open Generic Scanning context.
image

alistairjevans:
Good stuff; did you want to rename your `AsOpenTypesOf`, and see if it meets the semantics of `AssignableTo`?  If it does, that feels like a better name.

RaymondHuy:
Thanks @alistairjevans  I've updated my PR.

alistairjevans:
Thanks @RaymondHuy; can I suggest we add some tests for passing invalid types into `AssignableTo`? Mostly thinking of closed types, or types that aren't in the right inheritance chain or something.

RaymondHuy:
Hi @alistairjevans  I've add testcases to my PR for passing closed types and also resolving services which are not assignable to the target open generic type. Can you check if it is satisfied ? feel free to suggest more testcases :smiley:

test/Autofac.Test/Features/Scanning/OpenGenericScanningRegistrationTests.cs
```diff
@@ -119,6 +120,30 @@ public void AssignableToNullTypeProvidedThrowsException()
...
         }
 
+        [Theory]
+        [InlineData(typeof(ICloseCommand))]
+        [InlineData(typeof(CloseCommand))]
+        public void AssignableToClosedTypeProvidedNoneOpenGenericSourceRegistered(Type closedType)
+        {
+            var cb = new ContainerBuilder();
+            cb.RegisterAssemblyOpenGenericTypes(typeof(ICommand<>).GetTypeInfo().Assembly)
+                .AssignableTo(closedType);
```
alistairjevans, line 130: It feels like passing a closed type into `AssignableTo` for Open Generics is enough of an input mistake that we should throw an `ArgumentException` if someone tries it?

alistairjevans (changes requested):

test/Autofac.Test/Features/Scanning/OpenGenericScanningRegistrationTests.cs
RaymondHuy: @alistairjevans  when using `AssignableTo` in registering assembly closed type, it doesn't throw exception when the argument is open generic type. Therefore should we keep this behavior in this case or we will modify it too ?

alistairjevans: Ah, good point. We should retain the same behaviour for consistency, and I don't think it's worth a breaking change in the existing `AssignableTo`.

alistairjevans (approved):
Good to go! Thanks @RaymondHuy for all the work!

RaymondHuy:
Thanks @alistairjevans, btw I wonder there are any plans for adding SourceGenerator to Autofac ? I can spend times on the implementation if we can finalize the design or something like that.

alistairjevans:
I've given it a little thought before when implementing pipelines, but didn't get too far with it (https://github.com/autofac/Autofac/issues/1096#issuecomment-620524455).

The problem we have is the extremely dynamic nature of Autofac's registration system, that allows components to late-resolve objects using `Func<>` or `Lazy<>`, combined with all the assembly scanning functionality.  It gets really complicated to statically analyse, if not impossible. If we were to support source generators, there would most likely have to be extremely tight constraints on what you can use. 

I'm still open to using it though if we can find a good way; if you want to raise an issue to discuss design that would be just fine.

Issue #1252, opened 2021-02-16
OnActivated leaks memory


RogerKratz:
It seems OnActivated (and possibly other callbacks?) leaks memory. Below a little test that uses more and more memory. If OnActivated is removed, problem is gone

```c#
public class FooTest
{
  private int i;
  
  [Test]
  [Repeat(1000)]
  public void Bar()
  {
	  var builder = new ContainerBuilder();
	  Action<ContainerBuilder> registrationAction;
	  builder
		  .RegisterType<FakeService>()
		  .OnActivated(c =>
		  {
			  registrationAction = b =>
			  {
				  b.RegisterInstance(c.Instance);
			  };
		  });
	  var container = builder.Build();
	  container.Resolve<FakeService>();
	  container.Dispose();
	  
	  
	  i++;
	  if (i % 100 == 0)
		  Console.WriteLine((GC.GetTotalMemory(true) / 1024 / 1024) + "mb");
  }
  
  public class FakeService
  {
	  private IEnumerable<int> _data;
	  public FakeService()
	  {
		  _data = Enumerable.Range(1, 100).ToArray();
	  }
  }
}
```

tillig:
It'd probably be good to throw this in a memory profiler to see what's actually holding references to what. For example, I'm unclear what the `registrationAction` assignment is doing there. I'd also say you should do a direct `GC.Collect()` rather than having `GetTotalMemory` do it - `GetTotalMemory` [times out](https://docs.microsoft.com/en-us/dotnet/api/system.gc.gettotalmemory?view=net-5.0) rather than literally waiting for everything to be collected so it may not be entirely accurate.

So:

1. Get a profiler on that thing to find out for sure what's holding references.
2. Try not assigning something inside the closure. (Lose `registrationAction`.)
3. Consider using something like [BenchmarkDotNet](https://github.com/dotnet/BenchmarkDotNet) instead of manually accessing the memory information. BenchmarkDotNet will do all sorts of things like warming up, handling the garbage collection, etc., and will give you something possibly a bit more accurate. [Here's an example benchmark showing how to use `MemoryDiagnoser` to get the stats.](https://benchmarkdotnet.org/articles/samples/IntroNativeMemory.html)

On a side note... @alistairjevans this seems familiar, like we saw a similar issue where, again, some sort of assignment or work done inside the event handler was causing issues. Am I imagining that? I searched the issues list and couldn't find it.

RogerKratz:
Some quick replies...
[...]

PR #1253, opened 2021-02-17
A benchmark showing issue #1252

RogerKratz:
running this benchmark will allocate more and more memory without releasing it.


tillig:
I'll respond to this in #1252. Thanks!

PR #1271, opened 2021-04-24
Revisit doc-comments on a couple of key interfaces

### LINKED ISSUE #1269: Revisit `ILifetimeScope` interface docs
## Problem Statement

Based on [this StackOverflow question](https://stackoverflow.com/questions/67186529/using-registerinstance-in-beginlifetimescope-configuration-causes-memory-leak) it looks like we haven't really looked at the `ILifetimeScope` documentation (particularly [this section](https://github.com/autofac/Autofac/blob/39696a967e8826f7f1ebc8c1ff4523c9dd75abe0/src/Autofac/ILifetimeScope.cs#L73)) since doing the internals overhauls. I don't think this is accurate anymore but it persists.

Unclear if there are other unbroken/original interfaces that might have misleading docs.

## Desired Solution

Update docs as needed - it'd be nice if it's detailed and accurate, but doc maintenance is hard... so at least removing the stale stuff so it's not misleading folks would be good.

alistairjevans:
I've gone back over ILifetimeScope and a couple of other 'main' exposed types and updated them for accuracy/clarity.

Fixes #1269

### DIFF
--- a/src/Autofac/ILifetimeScope.cs
+++ b/src/Autofac/ILifetimeScope.cs
@@ -10,10 +10,15 @@ using Autofac.Core.Resolving;
 namespace Autofac
 {
     /// <summary>
+    /// <para>
     /// An <see cref="ILifetimeScope"/> tracks the instantiation of component instances.
     /// It defines a boundary in which instances are shared and configured.
-    /// Disposing an <see cref="ILifetimeScope"/> will dispose the components that were
-    /// resolved through it.
+    /// </para>
+    /// <para>
+    /// Disposing an <see cref="ILifetimeScope"/> will dispose any components that were
+    /// newly created when resolving from it, provided those components were not registered as
+    /// <see cref="IRegistrationBuilder{TLimit, TActivatorData, TRegistrationStyle}.ExternallyOwned"/>.
+    /// </para>
     /// </summary>
     /// <example>
     /// <code>
@@ -32,14 +37,21 @@ namespace Autofac
     /// </code>
     /// </example>
     /// <remarks>
+    /// <para>
     /// All long-running applications should resolve components via an
     /// <see cref="ILifetimeScope"/>. Choosing the duration of the lifetime is application-
-    /// specific. The standard Autofac WCF and ASP.NET/MVC integrations are already configured
-    /// to create and release <see cref="ILifetimeScope"/>s as appropriate. For example, the
-    /// ASP.NET integration will create and release an <see cref="ILifetimeScope"/> per HTTP
-    /// request.
+    /// specific.
+    /// </para>
+    /// <para>
+    /// The standard Autofac integrations are already configured
+    /// to create and dispose <see cref="ILifetimeScope"/>s as appropriate. For example, the
+    /// integration with ASP.NET Core (via the .NET DI library) will create and dispose
+    /// one <see cref="ILifetimeScope"/> per HTTP request.
+    /// </para>
+    /// <para>
     /// Most <see cref="ILifetimeScope"/> functionality is provided by extension methods
     /// on the inherited <see cref="IComponentContext"/> interface.
+    /// </para>
     /// </remarks>
     /// <seealso cref="IContainer"/>
     /// <seealso cref="IComponentContext"/>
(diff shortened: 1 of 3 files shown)


src/Autofac/ILifetimeScope.cs
```diff
@@ -70,9 +82,34 @@ public interface ILifetimeScope : IComponentContext, IDisposable, IAsyncDisposab
...
+        /// Resolving a service from this scope that was only defined in a parent <see cref="ILifetimeScope"/> (or root <see cref="IContainer"/>)
+        /// will result in a component 'owned' by the scope it was registered in (rather than this new scope).
+        /// E.g. if you registered a SingleInstance() component in the root container, but resolved it in the new sub-scope, the instance you get will
+        /// be attached to the root container, not your scope, so the same instance will be shared by all other <see cref="ILifetimeScope"/>s.
+        /// </para>
+        ///
+        /// <para>
+        /// SingleInstance() components registered in this scope will be held as long as this scope exists, but will be discarded
+        /// when the scope is disposed.
+        /// </para>
```
tillig, line 112:
```suggestion
        /// <para>
        /// Any components registered in the sub-scope will only be resolvable from this new scope,
        /// or additional child scopes subsequently resolved from it.
        /// </para>
        /// <para>
        /// Components registered in the sub-scope that provide services already registered
        /// in a parent <see cref="ILifetimeScope"/> (or root <see cref="IContainer"/>) will
        /// override that same service registration when resolving instances.
        /// </para>
        /// <para>
        /// If you resolve an <see cref="IEnumerable{TService}"/> in the new scope, the returned set of components will
        /// include both components registered in the new sub-scope, plus all registered components all the way back
        /// up to the root <see cref="IContainer"/>.
        /// </para>
        /// <para>
        /// Resolving a service from this scope that was only defined in a parent <see cref="ILifetimeScope"/> (or root <see cref="IContainer"/>)
        /// will result in a component 'owned' by the scope it was registered in (rather than this new scope).
        /// For example, if you registered a <c>SingleInstance()</c> component in the root container, but resolved it in the new sub-scope, the instance you get will
        /// be attached to the root container, not your scope, so the same instance will be shared by all other <see cref="ILifetimeScope"/> instances.
        /// </para>
        /// <para>
        /// <c>SingleInstance()</c> components registered in this scope will be held as long as this scope exists, but will be discarded
        /// when the scope is disposed.
        /// </para>
```

tillig: - Removed empty space between paragraphs
- Added `<see cref/>` around class references
- Denoted `SingleInstance()` as code using `<c></c>`

src/Autofac/ILifetimeScope.cs
```diff
@@ -85,9 +122,34 @@ public interface ILifetimeScope : IComponentContext, IDisposable, IAsyncDisposab
...
```
tillig, line 152:
```suggestion
        /// <para>
        /// Any components registered in the sub-scope will only be resolvable from this new scope,
        /// or additional child scopes subsequently resolved from it.
        /// </para>
        /// <para>
        /// Components registered in the sub-scope that provide services already registered
        /// in a parent <see cref="ILifetimeScope"/> (or root <see cref="IContainer"/>) will
        /// override that same service registration when resolving instances.
        /// </para>
        /// <para>
        /// If you resolve an <see cref="IEnumerable{TService}"/> in the new scope, the returned set of components will
        /// include both components registered in the new sub-scope, plus all registered components all the way back
        /// up to the root <see cref="IContainer"/>.
        /// </para>
        /// <para>
        /// Resolving a service from this scope that was only defined in a parent <see cref="ILifetimeScope"/> (or root <see cref="IContainer"/>)
        /// will result in a component 'owned' by the scope it was registered in (rather than this new scope).
        /// For example, if you registered a <c>SingleInstance()</c> component in the root container, but resolved it in the new sub-scope, the instance you get will
        /// be attached to the root container, not your scope, so the same instance will be shared by all other <see cref="ILifetimeScope"/> instances.
        /// </para>
        /// <para>
        /// <c>SingleInstance()</c> components registered in this scope will be held as long as this scope exists, but will be discarded
        /// when the scope is disposed.
        /// </para>
```

src/Autofac/IStartable.cs
```diff
@@ -12,10 +13,18 @@ namespace Autofac
     /// an <c>OnActivated()</c> event handler instead.
     /// </summary>
     /// <remarks>
+    /// <para>
     /// For equivalent "Stop" functionality, implement <see cref="IDisposable"/>. Autofac
     /// will always dispose a component before any of its dependencies (except in the presence
     /// of circular dependencies, in which case the components in the cycle are disposed in
     /// reverse-construction order.)
+    /// </para>
+    ///
```
tillig, line 22:
```suggestion
```

It's more like HTML where you don't actually put spaces between paragraphs.

tillig (changes requested):
Nice updates. A couple of minor suggested fixes, otherwise 🚀

src/Autofac/IStartable.cs
alistairjevans: I mean, to be honest, I still put spaces in my HTML...

tillig (dismissed):
🌮

src/Autofac/ILifetimeScope.cs
```diff
@@ -70,9 +82,30 @@ public interface ILifetimeScope : IComponentContext, IDisposable, IAsyncDisposab
...
+        /// Any components registered in the sub-scope will only be resolvable from this new scope,
+        /// or additional child scopes subsequently resolved from it.
+        /// </para>
+        /// <para>
+        /// Components registered in the sub-scope that provide services already registered
+        /// in a parent <see cref="ILifetimeScope"/> (or root <see cref="IContainer"/>) will
+        /// override that same service registration when resolving instances.
+        /// </para>
+        /// <para>
+        /// If you resolve an <see cref="IEnumerable{TService}"/> in the new scope, the returned set of components will
```
tillig, line 95:
```suggestion
        /// If you resolve an <see cref="System.Collections,Generic.IEnumerable{TService}"/> in the new scope, the returned set of components will
```
Compiler is complaining about not finishing IEnumerable 😩

alistairjevans: Resolved this problem manually.

tillig (approved):
🍏

Issue #1235, opened 2020-12-04
Enable OnlyIf handling for decorators and modules


tillig:
## Problem Statement

There is not currently an easy way to conditionally register decorators or modules using the `OnlyIf()` style syntax. You can register types conditionally...

```c#
builder.RegisterType<X>()
       .OnlyIf(b => b.IsRegistered(new TypedService(typeof(Y))));
```

...but you can't do that for decorators or modules. `RegisterDecorator<T, U>()` is a `void` method. While it offers the ability to conditionally _apply_ the decorator, if you don't want the perf hit of executing the condition function on every resolve there's no easy way to conditionally _register_.

`ContainerBuilder.RegisterModule<T>()` is similar - it has a special `IModuleRegistrar` that gets returned, but that only allows chained module registrations and not conditions.

This issue is [based on the StackOverflow question here.](https://stackoverflow.com/questions/65141943/autofac-6-register-decorators-conditional-on-the-availability-of-another-servi)

## Desired Solution

It would be nice to have `OnlyIf` support in the cases where we don't currently have it. This may require changing `RegisterDecorator<T, U>()` to return a new `IDecoratorRegistrar` or something similar such that it can add a chained callback for `OnlyIf()`. For modules, it's a new extension on that `IModuleRegistrar`.

## Alternatives You've Considered

- Make `DecoratorMiddleware` public: While this would potentially solve the issue in the StackOverflow question (conditional registration of a decorator), it would still mean the entire contents of `RegisterDecorator` would be copy/pasted into another codebase and have to be managed there. If we change the set of things in Autofac that are required to make a decorator work, the external user's code would break unexpectedly.
- ... that's about it. That wouldn't solve it for modules, either.

## Additional Context

Changing `RegisterDecorator<T, U>` to return a type instead of void is a breaking API change (I think... pretty sure) because the method signature changes. Transitive dependencies (my code => some other code that's compiled and registers decorators => Autofac) might break because it couldn't find the `void` version of `RegisterDecorator`. I'd have to test that, but... pretty sure. If so, that'd be a 7.x release and we might want to consider a new method/overload to handle that if we don't want to release the breaking change so soon after 6.0.

ModernRonin:
I think maybe this one here falls into the same category?
https://stackoverflow.com/questions/67022317/autofac-6-register-a-decorator-as-also-implementing-another-interface

tillig:
[...]

PR #1272, opened 2021-04-24
Add support for OnlyIf and IfNotRegistered on module registration

alistairjevans:
This PR is a partial fix for #1235, by adding OnlyIf support for Module Registrations.

This change is _technically_ breaking, through the addition of a new `RegistrarData` property on `IModuleRegistrar`, however if someone has implemented a custom `IModuleRegistrar`, I will be quite surprised.

This change allows you to do this:

```csharp

builder.RegisterModule(new MyModule())
       .OnlyIf(regBuilder => someCondition);
```

or this:

```csharp

builder.RegisterModule(new MyModule())
       .IfNotRegistered(typeof(string));
```

You can stack `OnlyIf` calls in the same way you can do it on normal registrations:

```csharp

builder.RegisterModule(new MyModule())
       .OnlyIf(regBuilder => someCondition)
       .OnlyIf(regBuilder => somethingElse);
```

The way I've laid out the evaluation of the conditions and the module registrations is as follows (after some thought):

- Evaluate any predicates (in reverse definition order, same as normal registrations), stopping if one fails.
- If all predicates pass, then run all module registrations.

This means that any `OnlyIf` statements used against a single `IModuleRegistrar` instance will be applied to the state of the registry **before** any modules got registered.

This behaviour seemed like the most predicatable/logical, rather than (for example), evaluating the predicate per-module.

### DIFF
--- a/src/Autofac/Core/Registration/ModuleRegistrar.cs
+++ b/src/Autofac/Core/Registration/ModuleRegistrar.cs
@@ -12,10 +12,7 @@ namespace Autofac.Core.Registration
     /// </summary>
     internal class ModuleRegistrar : IModuleRegistrar
     {
-        /// <summary>
-        /// The <see cref="ContainerBuilder"/> into which registrations will be made.
-        /// </summary>
-        private readonly ContainerBuilder _builder;
+        private Action<IComponentRegistryBuilder>? _nextModuleCallback;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="ModuleRegistrar"/> class.
@@ -28,8 +25,18 @@ namespace Autofac.Core.Registration
         /// </exception>
         public ModuleRegistrar(ContainerBuilder builder)
         {
-            _builder = builder ?? throw new ArgumentNullException(nameof(builder));
+            if (builder is null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            var callback = builder.RegisterCallback(reg => _nextModuleCallback?.Invoke(reg));
+
+            RegistrarData = new ModuleRegistrarData(callback);
         }
+
+        /// <inheritdoc />
+        public ModuleRegistrarData RegistrarData { get; }
 
         /// <summary>
         /// Add a module to the container.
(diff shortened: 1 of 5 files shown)


src/Autofac/Core/Registration/ModuleRegistrar.cs
```diff
@@ -12,10 +12,7 @@ namespace Autofac.Core.Registration
     /// </summary>
     internal class ModuleRegistrar : IModuleRegistrar
     {
-        /// <summary>
-        /// The <see cref="ContainerBuilder"/> into which registrations will be made.
-        /// </summary>
-        private readonly ContainerBuilder _builder;
+        private Action<IComponentRegistryBuilder>? _nextModuleCallback;
```
tillig, line 15: It took me a bit to figure out that this is effectively the call to `module.Configure` for all modules registered via this registrar. The `RegistrationData` allows wrapping of that chain with predicates, but `RegisterModule` needs the ability to add modules _inside that wrapper_.

I wonder if a more descriptive name like `_moduleConfigureChain` or something might help? I'm not glued to that as the name, but it did take me a few reads through here to go, "OOOHHH, it's predicates wrapped around this callback, and this callback gets sort of lazy-initialized via closure." Like two levels of callbacks, not just one "pipeline." I think `_nextModule` threw me off because I was expecting one pipeline, like ASP.NET Core `_next` and my poor zombie brain didn't connect that it's sorta two levels going on here.

I also had a big block of `//` comments written but was like... maybe it's just me and a more descriptive name would be enough.

tillig (commented):
I think this is pretty slick, though I made a note where the actual logic was somewhat confusing for a moment.

I think the "breaking" `IModuleRegistrar` change is fine. Honestly, if you've implemented `IModuleRegistrar` on your own, God have mercy on your soul.

src/Autofac/Core/Registration/ModuleRegistrar.cs
alistairjevans: The word 'next' does imply pipeline, you're right (in Autofac as well, never mind ASP.NET Core). Will re-word and add some comments.

tillig (approved):
🖲️

PR #1276, opened 2021-05-13
Fix #1275: Filter out 'value' named parameters during property injection

### LINKED ISSUE #1275: Delegate factory with parameter name 'value' throws
## Describe the Bug

Delegate factory throws on instance activation when it has parameter with name `value`

## Steps to Reproduce

```c#
public interface ISomeService { }
public class SomeService : ISomeService { }

public class MyComponent
{
    public ISomeService Service { get; set; }
    public delegate MyComponent Factory(string value);

    public MyComponent(string value)
    {
    }
}

public class ReproTest
{
    [Fact]
    public void Test()
    {
        // Arrange
        builder.RegisterType<SomeService>().As<ISomeService>();
        builder.RegisterType<MyComponent>().PropertiesAutowired();

        // Act
        var factory = container.Resolve<MyComponent.Factory>();

        // Assert
        Assert.NotNull(factory("Autofac is great!"));
    }
}

```

## Expected Behavior

Delegate factory should produce properly activated component with `Service` property resolved.

## Exception with Stack Trace

Autofac.Core.DependencyResolutionException : An exception was thrown while activating UnitTests.MyComponent.
---- System.InvalidCastException : Unable to cast object of type 'System.String' to type 'UnitTests.ISomeService'.

## Dependency Versions

Autofac: 6.2.0

## Additional Info

What I've tried so far:
- changing delegate factory parameter name to something else fixed the issue
- changing parameter type to other type (e.g. string, object) without changing the name still throws

So, to work around the issue, delegate factory should not have parameter with name `value`.

tillig:
Resolves #1275. When doing property injection, the set of parameters provided to the resolve operation are all considered for properties. A special edge case occurs when a named parameter is provided where the name is `value` - this overlaps with the reserved word `value` in a property setter method and it's seen to be a "match" for the property.

This change removes this very specific edge case from the considered list of parameters when doing property injection.

### DIFF
--- a/src/Autofac/Core/Activators/Reflection/AutowiringPropertyInjector.cs
+++ b/src/Autofac/Core/Activators/Reflection/AutowiringPropertyInjector.cs
@@ -75,7 +75,12 @@ namespace Autofac.Core.Activators.Reflection
                 // SetMethod will be non-null if GetInjectableProperties included it.
                 var setParameter = property.SetMethod!.GetParameters()[0];
                 var valueProvider = (Func<object?>?)null;
-                var parameter = resolveParameters.FirstOrDefault(p => p.CanSupplyValue(setParameter, context, out valueProvider));
+
+                // Issue #1275: If a delegate factory has a named parameter 'value' it will try to
+                // inject into any property being autowired.
+                var parameter = resolveParameters.FirstOrDefault(p =>
+                    p.CanSupplyValue(setParameter, context, out valueProvider) &&
+                    !(p is NamedParameter n && n.Name.Equals("value", StringComparison.Ordinal)));
                 if (parameter != null)
                 {
                     var setter = PropertySetters.GetOrAdd(property, MakeFastPropertySetter);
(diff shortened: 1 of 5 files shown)


alistairjevans (approved):
All looks good here; definitely an amusing issue. A little surprised it hasn't cropped up before!

OwnageIsMagic:
I doubt that `NamedParameter`s should be considered for property injection at all. C# doesn't allow to change setter parameter name and but VB allows naming it and for auto-property it's `AutoPropertyValue`.
https://sharplab.io/#v2:DYLgbgRgNALiCWwA+BhYBDAzpgBAUQA90BbAB2AFMBYAKB3pwAUBXCYeAYyYCcB7UitxgBPHADlmxCIIAUAShwBBXAEkAdjAoBzQbVoMmrdlwDq3eJoDya4KMZ8BQ0RKmCATPKWqN23XQMMAMoUMDgyACK8Yrww6gBWFBwwALIUXjjqmjrccvoB9HhqACY4wTB5BcU8/IIiejSFJWhYmEA==

PR #1285, opened 2021-08-21
Ensure that IAsyncDisposable instances that have not been activated are correctly disposed when the scope is disposed.

alistairjevans:
This change was slightly larger than expected; due to the somewhat viral nature of async code, I had to add `DisposeAsync` methods in a number of places.  

This necessitated the addition of `IAsyncDisposable` as a base interface on some of our public types. However, the types modified are `IComponentRegistration` and `IComponentRegistry`, which fall into the category of "things that people probably won't implement themselves."

I avoided a breaking change on `IInstanceActivator` (perhaps more likely to be implemented) by checking if it's an IAsyncDisposable before continuing.

### DIFF
--- a/src/Autofac/Core/Activators/ProvidedInstance/ProvidedInstanceActivator.cs
+++ b/src/Autofac/Core/Activators/ProvidedInstance/ProvidedInstanceActivator.cs
@@ -5,6 +5,7 @@ using System;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
+using System.Threading;
 using System.Threading.Tasks;
 using Autofac.Core.Resolving;
 using Autofac.Core.Resolving.Pipeline;
@@ -85,7 +86,11 @@ namespace Autofac.Core.Activators.ProvidedInstance
                     Trace.TraceWarning(DisposerResources.TypeOnlyImplementsIAsyncDisposable, LimitType.FullName);
 
                     // Type only implements IAsyncDisposable. We will need to do sync-over-async.
-                    asyncDisposable.DisposeAsync().AsTask().ConfigureAwait(false).GetAwaiter().GetResult();
+                    // We want to ensure we lose all context here, because if we don't we can deadlock.
+                    // So we push this disposal onto the threadpool.
+                    Task.Run(async () => await asyncDisposable.DisposeAsync().ConfigureAwait(false))
+                        .ConfigureAwait(false)
+                        .GetAwaiter().GetResult();
                 }
             }
 
--- a/src/Autofac/Core/Activators/ProvidedInstance/ProvidedInstanceActivator.cs
+++ b/src/Autofac/Core/Activators/ProvidedInstance/ProvidedInstanceActivator.cs
@@ -85,7 +85,7 @@ namespace Autofac.Core.Activators.ProvidedInstance
                     Trace.TraceWarning(DisposerResources.TypeOnlyImplementsIAsyncDisposable, LimitType.FullName);
 
                     // Type only implements IAsyncDisposable. We will need to do sync-over-async.
-                    asyncDisposable.DisposeAsync().AsTask().GetAwaiter().GetResult();
+                    asyncDisposable.DisposeAsync().AsTask().ConfigureAwait(false).GetAwaiter().GetResult();
                 }
             }
 
--- a/src/Autofac/Core/Activators/ProvidedInstance/ProvidedInstanceActivator.cs
+++ b/src/Autofac/Core/Activators/ProvidedInstance/ProvidedInstanceActivator.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.Threading.Tasks;
@@ -79,14 +80,12 @@ namespace Autofac.Core.Activators.ProvidedInstance
                 {
                     disposable.Dispose();
                 }
-                else if (_instance is IAsyncDisposable)
+                else if (_instance is IAsyncDisposable asyncDisposable)
                 {
-                    // Type only implements IAsyncDisposable, which is not valid if there
-                    // is a synchronous dispose being done.
-                    throw new InvalidOperationException(string.Format(
-                        DisposerResources.Culture,
-                        DisposerResources.TypeOnlyImplementsIAsyncDisposable,
-                        _instance.GetType().FullName));
+                    Trace.TraceWarning(DisposerResources.TypeOnlyImplementsIAsyncDisposable, LimitType.FullName);
+
+                    // Type only implements IAsyncDisposable. We will need to do sync-over-async.
+                    asyncDisposable.DisposeAsync().AsTask().GetAwaiter().GetResult();
                 }
             }
 
--- a/src/Autofac/Core/Activators/ProvidedInstance/ProvidedInstanceActivator.cs
+++ b/src/Autofac/Core/Activators/ProvidedInstance/ProvidedInstanceActivator.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
+using System.Threading.Tasks;
 using Autofac.Core.Resolving;
 using Autofac.Core.Resolving.Pipeline;
 
@@ -64,21 +65,58 @@ namespace Autofac.Core.Activators.ProvidedInstance
         /// </summary>
         public bool DisposeInstance { get; set; }
 
-        /// <summary>
-        /// Releases unmanaged and - optionally - managed resources.
-        /// </summary>
-        /// <param name="disposing"><c>true</c> to release both managed and unmanaged resources; <c>false</c> to release only unmanaged resources.</param>
+        /// <inheritdoc />
         protected override void Dispose(bool disposing)
         {
             // Only dispose of the instance here if it wasn't activated. If it was activated,
             // then either the owning lifetime scope will dispose of it automatically
             // (see InstanceLookup.Activate) or an OnRelease handler will take care of it.
-            if (disposing && DisposeInstance && _instance is IDisposable disposable && !_activated)
+            if (disposing && DisposeInstance && !_activated)
             {
-                disposable.Dispose();
+                // If we are in synchronous dispose, and an object implements IDisposable,
+                // then use it.
+                if (_instance is IDisposable disposable)
+                {
+                    disposable.Dispose();
+                }
+                else if (_instance is IAsyncDisposable)
+                {
+                    // Type only implements IAsyncDisposable, which is not valid if there
+                    // is a synchronous dispose being done.
+                    throw new InvalidOperationException(string.Format(
+                        DisposerResources.Culture,
+                        DisposerResources.TypeOnlyImplementsIAsyncDisposable,
+                        _instance.GetType().FullName));
+                }
             }
 
             base.Dispose(disposing);
+        }
+
+        /// <inheritdoc />
+        protected override async ValueTask DisposeAsync(bool disposing)
+        {
+            if (disposing && DisposeInstance && !_activated)
+            {
+                // If the item implements IAsyncDisposable we will call its DisposeAsync Method.
+                if (_instance is IAsyncDisposable asyncDisposable)
+                {
+                    var vt = asyncDisposable.DisposeAsync();
+
+                    // Don't await if it's already completed (this is a slight gain in performance of using ValueTask).
+                    if (!vt.IsCompletedSuccessfully)
+                    {
+                        await vt.ConfigureAwait(false);
+                    }
+                }
+                else if (_instance is IDisposable disposable)
+                {
+                    // Call the standard Dispose.
+                    disposable.Dispose();
+                }
+            }
+
+            // Do not call the base, otherwise the standard Dispose will fire.
         }
 
         [SuppressMessage("CA2222", "CA2222", Justification = "False positive. GetType doesn't hide an inherited member.")]
--- a/src/Autofac/Core/Registration/ComponentRegistryBuilder.cs
+++ b/src/Autofac/Core/Registration/ComponentRegistryBuilder.cs
@@ -64,6 +64,7 @@ namespace Autofac.Core.Registration
             _registeredServicesTracker.Registered -= OnRegistered;
             _registeredServicesTracker.RegistrationSourceAdded -= OnRegistrationSourceAdded;
 
+            // Do not call the base, otherwise the standard Dispose will fire.
             return _registeredServicesTracker.DisposeAsync();
         }
 
--- a/src/Autofac/Core/Registration/ComponentRegistryBuilder.cs
+++ b/src/Autofac/Core/Registration/ComponentRegistryBuilder.cs
@@ -4,7 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
-
+using System.Threading.Tasks;
 using Autofac.Builder;
 using Autofac.Core.Resolving.Pipeline;
 using Autofac.Util;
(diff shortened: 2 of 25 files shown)


src/Autofac/Core/Activators/ProvidedInstance/ProvidedInstanceActivator.cs
```diff
@@ -64,23 +65,60 @@ private object GetInstance()
...
+                // then use it.
+                if (_instance is IDisposable disposable)
+                {
+                    disposable.Dispose();
+                }
+                else if (_instance is IAsyncDisposable)
+                {
+                    // Type only implements IAsyncDisposable, which is not valid if there
+                    // is a synchronous dispose being done.
+                    throw new InvalidOperationException(string.Format(
```
tillig, line 86: I'm somewhat torn here. I'm curious if we should throw or just do like a `DisposeAsync().GetAwaiter().Wait()`. Or maybe nothing?

 I'm not sure about throwing during `Dispose`, like how [WCF would do if a network connection dropped](https://docs.microsoft.com/en-us/dotnet/framework/wcf/samples/use-close-abort-release-wcf-client-resources) and we ended up having to [work around that with a special `OnRelease` handler](https://github.com/autofac/Autofac.Wcf/blob/8f9a319957fc6e837d1ec4dd0fcc91eacecb9a14/src/Autofac.Integration.Wcf/RegistrationExtensions.cs#L25). I know I'm using [`ServiceBusClient`, which is only `IAsyncDisposable`](https://github.com/Azure/azure-sdk-for-net/blob/9ee228648bcd83987b67d7207b140a3e8f8af2b6/sdk/servicebus/Azure.Messaging.ServiceBus/src/Client/ServiceBusClient.cs#L20), in some of our code and an exception here would be both unexpected and possibly not easy to fix, especially if the container is getting disposed by something outside my control during app shutdown.

src/Autofac/Core/Activators/ProvidedInstance/ProvidedInstanceActivator.cs
```diff
@@ -64,23 +65,60 @@ private object GetInstance()
...
 
+        /// <inheritdoc />
+        protected override async ValueTask DisposeAsync(bool disposing)
+        {
+            if (disposing && DisposeInstance && !_activated)
+            {
+                // If the item implements IAsyncDisposable we will call its DisposeAsync Method.
+                if (_instance is IAsyncDisposable asyncDisposable)
+                {
+                    var vt = asyncDisposable.DisposeAsync();
```
tillig, line 104: This little block of "if/else" plus "get value task, check completed, await result if not" appears a few times. Would it be worth having a static method for this with `[MethodImpl(MethodImplOptions.AggressiveInlining)]` on it to force it to be inlined? I think it should compile down to the same thing but maybe reduce some duplication? Or maybe it's not worth the lack of reading clarity. Hmmm. I don't feel strongly about it, figured I'd raise it as a possibility.

src/Autofac/Core/Registration/ComponentRegistryBuilder.cs
```diff
@@ -61,6 +58,15 @@ protected override void Dispose(bool disposing)
...
         }
 
+        /// <inheritdoc />
+        protected override ValueTask DisposeAsync(bool disposing)
+        {
+            _registeredServicesTracker.Registered -= OnRegistered;
+            _registeredServicesTracker.RegistrationSourceAdded -= OnRegistrationSourceAdded;
+
+            return _registeredServicesTracker.DisposeAsync();
+        }
```
tillig, line 68: This comment is on the other spots where base is intentionally not called.

```suggestion

            // Do not call the base, otherwise the standard Dispose will fire.
        }
```

test/Autofac.Specification.Test/Lifetime/AsyncDisposeProvidedInstanceTests.cs
```diff
@@ -0,0 +1,209 @@
...
+            var container = builder.Build();
+            var scope = container.BeginLifetimeScope(b => b.RegisterInstance(disposable));
+
+            await scope.DisposeAsync();
+
+            Assert.True(disposable.IsAsyncDisposed);
+        }
+
+        [Fact]
+        public async Task UnresolvedProvidedInstances_ActualAsyncDisposable_CorrectlyDisposed()
```
tillig, line 150: The name on this test reads weird. It sounds like I should be looking for the instance to be disposed, but in this case it's `ExternallyOwned` and not disposed. Looks like it's a copy/paste of `ResolvedProvidedInstances_NotOwnedByNestedLifetimeScope_NeverDisposed` up on line 93?

test/Autofac.Specification.Test/Util/AsyncOnlyDisposeTracker.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        public event EventHandler<EventArgs> Disposing;
+
+        public bool IsAsyncDisposed { get; set; }
+
+        public async ValueTask DisposeAsync()
+        {
+            if (_completeAsync)
+            {
+                await Task.Delay(1);
```
tillig, line 26: Would `await Task.CompletedTask` accomplish the same thing or is the intent here to also introduce a minor timing delay to force some nature of asynchronicity into it?

tillig (changes requested):
No joke on the viral nature of async stuff! Looks pretty good, a couple of minor comments/questions.

src/Autofac/Core/Activators/ProvidedInstance/ProvidedInstanceActivator.cs
alistairjevans: So, this "throw if you're service isn't going to be disposable" was the approach we took for other service types when we added IAsyncDisposable support.

I suppose the problematic scenario is the one where you aren't implementing the component, like `ServiceBusClient`, so can't remedy the problem.

We can do the `GetAwaiter().Wait()`, it just means that user's might not be getting the nice async dispose they think they're getting.  I too am torn.

src/Autofac/Core/Activators/ProvidedInstance/ProvidedInstanceActivator.cs
alistairjevans: I get kind of nervous whenever I write code that passes a `ValueTask` as a parameter. The guidance from Microsoft is pretty clear that ValueTasks should generally just be [immediately awaited, or have the IsCompleted check](https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/#valid-consumption-patterns-for-valuetasks), and that's it.  Not saying it's impossible trying to write a method that makes this check common, but it seems to go against best practice.

It seems more verbose, but I think the controlled management of the ValueTask, plus the readability, outweighs the duplication.

test/Autofac.Specification.Test/Util/AsyncOnlyDisposeTracker.cs
alistairjevans: The intent here is just to force the method to be properly async, rather than completing synchronously, therefore ensuring that calling code that checks the `!vt.IsCompletedSuccessfully` flag gets invoked, and does so on a different thread.

test/Autofac.Specification.Test/Lifetime/AsyncDisposeProvidedInstanceTests.cs
alistairjevans: Yeah, that's just me getting the test wrong.

src/Autofac/Core/Activators/ProvidedInstance/ProvidedInstanceActivator.cs
tillig: Yeah, I didn't think the exception was going to be an issue until I hit this `ServiceBusClient` thing where it's causing all sorts of weird sync/async disposal problems because they didn't implement both. I have to imagine there are other cases like this out there, I just happen to find a concrete pain in my ass right here so I'm latching onto it.

I guess it comes down to: would it be better to dispose synchronously or not dispose at all? And if it doesn't dispose (and throws), it looks like [it stops everything else from being disposed, too.](https://github.com/autofac/Autofac/blob/79bfeeffbf753a86c7907cc58abc204602ff4014/src/Autofac/Core/Disposer.cs#L52) Which I'll admit sounded right at the time but `IAsyncDisposable` was kind of new and I didn't really _grok it_ like I'm starting to now.

It's almost like it should be a warning rather than an exception. Not that that's a thing, just like, "Hey, you did something wrong and we'll recover for you by doing it synchronously, but you're not getting the nice async you think you are."

Barring that... my _gut_ is telling me we should do what we can to dispose without throwing.

Another option might be to require folks to opt in with something like

```c#
builder.RegisterType<ServiceBusClient>().SynchronousDisposal();
```

where `SynchronousDisposal` could be an `OnRelease` thing sort of like the WCF `SafeRelease` extension where it wraps `DisposeAsync().GetAwaiter().Wait()` logic.

What's the "pit of success" here? What will make people more readily successful with the least amount of client work? Successful by default?

alistairjevans: Minor addition, we would need to `AsTask()` on the result of `DisposeAsync`, you cannot `GetAwaiter()` on a `ValueTask`. Only worth mentioning as another allocation that would happen during dispose.

alistairjevans: Like you say, raising a warning would be nice. I think the pit of success is "probably" `GetAwaiter().Wait()`, but perhaps we could write something somewhere, like to `Debug.Trace`?

src/Autofac/Core/Activators/ProvidedInstance/ProvidedInstanceActivator.cs
tillig: I'll buy that.

src/Autofac/Core/Activators/ProvidedInstance/ProvidedInstanceActivator.cs
tillig: I'd go for a `Debug.WriteLine` or something similar. Maybe `Trace` instead of `Debug` - doesn't `Debug` get compiled out in release mode? I vaguely recall that being a thing. I could see possibly raising an event with the diagnostic listener thing for issues like this, too, maybe... but that'd mean propagating it around down to this level and yeahno.

alistairjevans: Ok, this is done; used `Trace.TraceWarning`. Attached tracers understand it's a warning level, which is good.

tillig (approved):
📱

tillig:
It looks like the build is hanging on the unit tests in `netcoreapp3.1` for `Autofac.Test`. I'm not sure what's up, it gets past the `Autofac.Specification.Test` without issue but hangs up for `Autofac.Test`.

tillig:
I can't replicate the hang locally. I wonder if updating the `global.json` to the latest SDK versions would have any effect. I used the `build.ps1` script so it installed the specified versions, but I'm guessing the `latestFeature` thing allowed it to slip forward to use my later installed version instead of forcing the older version.

alistairjevans:
I'll stop the build and change the global.json.

alistairjevans:
That same path separator problem needed applying to the main Autofac repo, so including it in this PR.

tillig (dismissed):
🖱️

alistairjevans:
Still hanging.

These deadlock issues remind me of one of the reasons why I didn't do this sync-over-async in the first place.

https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html

I tried `ConfigureAwait(false)`, but to no avail. 

Worth noting that a deadlock is far harder to understand/debug than an exception....

alistairjevans:
Interestingly only hangs on net5; perhaps some threading thing has been changed.

tillig:
Hmmmm. Agreed that exception is easier to troubleshoot. I also have to admit I don't really understand why disposal needs to be async at all but OOOOOKAY. Waiting on async constructors now because that'll be the best idea EVAR. 😩 

I think my biggest issue with the exception was that we didn't allow stuff to be disposed that _could be_ disposed before throwing.

On the other hand, if I wanted to bridge the gap and put some sort of extension that allows sync disposal of an only async component, I'd still be stuck with this issue and it'd be entirely outside my control - thing X (`ServiceBusClient`) does async disposal but thing Y (`BackgroundService`) only does sync disposal. 💥

alistairjevans:
OK; fixed the deadlock via the somewhat nuclear option of bumping the Disposal of IAsyncDisposable-only objects onto the threadpool and waiting for completion, to ensure we get a completely fresh async context.

Basically, if I just call `ConfigureAwait(false).GetAwaiter().GetResult()` from the current thread, every method called by the disposing type will also need `ConfigureAwait(false)` to avoid deadlocks, which is impossible to enforce.

tillig (approved):
🤯

tillig:
I can't think of anything better. Luckily it should be very rare. Nice work figuring it out!

src/Autofac/Core/Activators/ProvidedInstance/ProvidedInstanceActivator.cs
```diff
@@ -64,23 +67,62 @@ private object GetInstance()
...
+
+                    // Don't await if it's already completed (this is a slight gain in performance of using ValueTask).
```
OwnageIsMagic, line 110: there is no perf gain as generated state machine is doing exactly the same. It's worth doing that manually only in non-async functions (this avoids creating state machine).

tillig: @OwnageIsMagic While we appreciate folks getting more eyes on the code, I'm not sure what the goal is with some of the comments here. If you have a pull request you'd like to submit to help improve the product, awesome, we'd love to check it out; however, many of these things are months or years completed, so we're not likely to go back and take any action on them at this time.

OwnageIsMagic: I'm just backporting stuff to netfx and looking through some commits.
Before commenting I check develop branch and the code in question is still there, so it quite actionable.
For now I'm just informing you that this comment here is plain wrong, so you (and other maintainers) can consider not doing this in the future (sharing knowledge).
Regardless my [other comment](https://github.com/autofac/Autofac/pull/1276#issuecomment-1243804201), I think it's have more sense it context of old issue, rather then creating new issue for it.

OwnageIsMagic: @tillig as I said it's not _something from a commit months or years ago_.
It right there, in the HEAD of develop (no pun intended) https://github.com/autofac/Autofac/blob/develop/src/Autofac/Core/Activators/ProvidedInstance/ProvidedInstanceActivator.cs#L106
I guess that thing you are trying to say, is that this style of contribution doesn't fit into your workflow and I should open a new issue.

tillig: It may be in `develop` but scroll up and you'll see this PR with this commit is from August 2021, so, yes, it is from over a year ago. But, also, correct, comments on code without any context or issue filed are not as helpful in our flow as it may be in other repositories. There isn't a big team here like on ASP.NET Core or the .NET Runtime and no one here's getting paid. Given that, if folks want action, we have to be a little more specific about how the intake process works. It's great to get more eyes on the code, but if something needs to change or if there's a bug, something incorrect, whatever, it really needs to be an issue so we can prioritize it. Comments like this are potentially interesting discussion points, but likely won't result in anything but discussion, and, as noted, we don't have a ton of bandwidth given the small size of the team, so it may not even result in discussion.

test/Autofac.Specification.Test/Util/AsyncOnlyDisposeTracker.cs
OwnageIsMagic: @alistairjevans you can use `Task.Yield()`. it doesn't involve timer queue.

PR #1289, opened 2021-10-08
Check iscompilergenerated last in canberegistered for perf reasons

RogerKratz:
Reimplements PR https://github.com/autofac/Autofac/pull/994, please see that PR for more info what/why this fix is for.
We recently upgraded autofac and noticed that the fix was gone. This makes "test" `AssemblyScanningPerformanceTests.MeasurePerformance` perf better again.


tillig:
[6.3.0 is cut!](https://github.com/autofac/Autofac/releases/tag/v6.3.0)

alistairjevans:
If that scanning method is a hot path, it may also be worth looking at reducing the LINQ usage in future, it's possible there's additional allocations happening inside the Where loop that we could avoid to save some extra CPU.

PR #1299, opened 2021-12-09
Move WithMetadata with mapping for OpenGenericAssemblyScanning

romerod:
Add support for adding metadata accepting a type to metadata mapper.

### DIFF
--- a/src/Autofac/Features/Scanning/OpenGenericScanningRegistrationExtensions.cs
+++ b/src/Autofac/Features/Scanning/OpenGenericScanningRegistrationExtensions.cs
@@ -252,23 +252,6 @@ namespace Autofac.Features.Scanning
         }
 
         /// <summary>
-        /// Specify how an open generic type from a scanned assembly provides metadata.
-        /// </summary>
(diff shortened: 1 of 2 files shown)


tillig:
I think this looks good. I'm not sure why it was internal to start - we have a test for it to verify it works, but it was working due to access to internals. Before I approve - @alistairjevans @RaymondHuy do you guys know if there was a specific reason this was internal or was it just an oversight?

alistairjevans:
Looks like it was just an oversight; the tests for it clearly believe it to be publicly accessible. @romerod, can you also move the `WithMetadataFrom` method below it, which I suspect fell foul of the same problem?

RaymondHuy:
@tillig  yes, it's my mistake, I'm sorry about it.

romerod:
OK, i'll move the other method on monday.

Thanks for the review.

Holen Sie sich Outlook für Android<https://aka.ms/AAb9ysg>
________________________________

tillig (approved):
LGTM! Thanks!

PR #1300, opened 2021-12-10
Target .Net 6 Framework and upgrade additionalSdks to latest version

anwarjaved:


anwarjaved:
It Seems .net SDK not there on appveyor which failing build

tillig:
This is going to be possibly more involved than you imagined. I'll start a review for it, but if it sounds like something you're not interested in taking on, it's cool if you want to abandon the PR.

bench/Autofac.BenchmarkProfiling/Autofac.BenchmarkProfiling.csproj
```diff
@@ -1,13 +1,13 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFrameworks>netcoreapp3.1;net5.0;net6.0</TargetFrameworks>
```
tillig, line 5:
```suggestion
    <TargetFramework>net6.0</TargetFramework>
```
We don't actually run the benchmarks in every framework - just one, enough to see if something has changed dramatically. Going with just net6 is fine.

bench/Autofac.Benchmarks/Autofac.Benchmarks.csproj
```diff
@@ -28,13 +28,13 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.6.1" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.0.0" />
```
tillig, line 31: I can't make a comment at the top of this file, but if the `Autofac.BenchmarkProfiling.csproj` changes to net6, this project does, too.

global.json
```diff
@@ -1,10 +1,10 @@
 {
   "sdk": {
-    "version": "5.0.400",
+    "version": "6.0.0",
```
tillig, line 3: You need to reference an actual, specific version in `global.json`. [Steal the `global.json` from Autofac.Extensions.DepdencyInjection, where we already target .NET 6.](https://github.com/autofac/Autofac.Extensions.DependencyInjection/blob/develop/global.json)

src/Autofac/Autofac.csproj
```diff
@@ -4,7 +4,7 @@
     <Description>Autofac is an IoC container for Microsoft .NET. It manages the dependencies between classes so that applications stay easy to change as they grow in size and complexity.</Description>
     <!-- VersionPrefix patched by AppVeyor -->
     <VersionPrefix>0.0.1</VersionPrefix>
-    <TargetFrameworks>netstandard2.1;netstandard2.0;net5.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.1;netstandard2.0;net5.0;net6.0</TargetFrameworks>
     <LangVersion>latest</LangVersion>
     <Nullable>enable</Nullable>
     <NoWarn>$(NoWarn);CS1591;IDE0008</NoWarn>
```
tillig, line 10: Down on line 31 you'll see `<Features>` - I can't make a comment that far down in the PR interface. The way analyzers work in .NET 6 is different than in other frameworks. We have to turn some things off and turn some things on.

- `<Features>` needs to be removed. That's obsolete now.
- `<AnalysisMode>AllEnabledByDefault</AnalysisMode>` needs to be added where the `<Features>` line is.

This also needs to be done in the test projects that have analyzers. I don't know if they all have `<Features>` but that `<AnalysisMode>` line needs to be there.

This may cause some new warnings to show up. All warnings need to be resolved as part of the .NET upgrade - we don't allow builds with warnings. This may be more than you want to take on.

src/Autofac/Autofac.csproj
```diff
@@ -48,19 +48,21 @@
...
-      <PrivateAssets>All</PrivateAssets>
+    <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="6.0.0">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" Condition="Exists('$(MSBuildThisFileDirectory)../../.git')">
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" Condition="Exists('$(MSBuildThisFileDirectory)../../.git')">
       <PrivateAssets>All</PrivateAssets>
     </PackageReference>
     <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.205">
```
tillig, line 62:
```suggestion
    <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.376">
```

The .NET 6 support in StyleCop is constantly changing. This also needs to be updated in the test projects that reference StyleCop.

src/Autofac/Autofac.csproj
```diff
@@ -48,19 +48,21 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.1">
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="6.0.0" />
```
tillig, line 51: This `Microsoft.Bcl.AsyncInterfaces` package was originally inside a conditional include so it only applied for `netstandard2.0`. It needs to stay conditional. I don't think it needs to even be updated if it's in the conditional, and that'd be fine. We don't want to force folks to take package updates.

test/Autofac.Test.Scenarios.ScannedAssembly/Autofac.Test.Scenarios.ScannedAssembly.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.1</TargetFramework>
```
tillig, line 4: This assembly is only used as a target for assembly scanning. It should only target the lowest framework required - revert to `netstandard2.1`.

tillig (changes requested):
I think I caught all the stuff that needs to change. Unfortunately, it's not necessarily as straightforward as you might think. You can use [Autofac.Extensions.DependencyInjection](https://github.com/autofac/Autofac.Extensions.DependencyInjection) as a help/guide if you get stuck or wonder what we did elsewhere - that's the only project so far that we've updated to .NET 6.

Note there's not really a reason to update core Autofac that I'm aware of. We updated A.E.DI because there were new Microsoft.Extensions.DependencyInjection features that only exist in .NET 6 that we needed to support. There's nothing specific in .NET 6 that I'm aware of that core Autofac would need. That said, having some of the nicer new language features is helpful, so I'm for it.

test/Autofac.Test.Scenarios.ScannedAssembly/Autofac.Test.Scenarios.ScannedAssembly.csproj
anwarjaved: done

src/Autofac/Autofac.csproj
anwarjaved: Updated StyleCop Package

src/Autofac/Autofac.csproj
anwarjaved: Done

global.json
anwarjaved: Done

bench/Autofac.Benchmarks/Autofac.Benchmarks.csproj
anwarjaved: Yes Done

bench/Autofac.BenchmarkProfiling/Autofac.BenchmarkProfiling.csproj
anwarjaved: Done

anwarjaved:
> I think I caught all the stuff that needs to change. Unfortunately, it's not necessarily as straightforward as you might think. You can use [Autofac.Extensions.DependencyInjection](https://github.com/autofac/Autofac.Extensions.DependencyInjection) as a help/guide if you get stuck or wonder what we did elsewhere - that's the only project so far that we've updated to .NET 6.
> 
> Note there's not really a reason to update core Autofac that I'm aware of. We updated A.E.DI because there were new Microsoft.Extensions.DependencyInjection features that only exist in .NET 6 that we needed to support. There's nothing specific in .NET 6 that I'm aware of that core Autofac would need. That said, having some of the nicer new language features is helpful, so I'm for it.

I am going through Autofac.Extensions.DependencyInjection to understand what else I need to do

anwarjaved:
I think its done now, Please review again @tillig

tillig:
Looks like the build is failing with some of the new analyzer warnings (warnings-as-errors is on in the final build). At first glance the fixes for some of these, while totally correct and valid, may cause some breaking API changes. It may be that we need to put this on a net6 branch or something while we ramp up for a v7 major release. In particular I'm looking at some of the ones where it wants to change to `EventHandler<T>` - which appears to be one of the new analyzer messages. It also looks like some of our [back-compat shims](https://github.com/autofac/Autofac/blob/develop/src/Autofac/Util/LinkerAttributes.cs) need to get some `NET6_0` conditionals.

@alistairjevans do you have a preference on going straight to `develop` with this or maybe a `v7` branch where we can queue up some breaking changes? Thinking v7. There's one `[Obsolete]` item I need to mark for another issue as part of v7 as well.

tillig:
Hmmmm. Also not sure if doing the `EventHandler<T>` thing is worth the breaking changes. Maybe do that last. I'm thinking about suppressing that one because the breaking change on event handler types isn't really worth the "clean code" of the `EventHandler<T>` thing.

alistairjevans:
If we think the .NET6 version will have breaking changes, then yes, a separate branch makes sense. Is the next version going to have to be major to introduce the new obsoleted feature? 

From my experience moving to .NET6, there's not a huge motivation to introduce breaking change, it's just nice to get the more performant IL and runtime.

If we have any `NET5_0` conditions, we should change them to `NET5_0_OR_GREATER` instead, to save us having to do this again.

tillig:
I just didn't want to add `[Obsolete]` warnings to folks in a 0.1.0 update. I've had that happen and it sucks.

I think if we can get away without breaking changes, going to develop with this is fine. I can postpone the `[Obsolete]` thing for later.

Let's use
`[SuppressMessage("CA1003", "CA1003", Justification = "Not changing to EventHandler<T> to avoid breaking changes.")]`
to suppress the individual event handler warnings. I'd like to keep the check ON in case we add new events - I'd like to start heading in that direction. But the attribute (from `System.Diagnostics.CodeAnalysis`) will turn off the individual warnings for those messages and avoid the breaking change. We can decide when or if to change them later.

tillig:
There's also a warning about `SegmentedStack` where it doesn't like the name. We'll need to suppress that one, too, since it's public and not interesting to break.

`[SuppressMessage("CA1711", "CA1711", Justification = "Not changing Stack suffix to avoid breaking changes.")]`

tillig:
Thanks for the kickstart here; I've opened a new PR #1306 to take this and finish the remaining work. I'll close this one and we can iterate there as needed.

PR #1306, opened 2022-02-23
.NET 6 Support

tillig:
Includes:

- Adding `net6.0` TFM to all projects.
- Updating test framework, analyzer, and other dev dependencies.
- Resolving new analyzer warnings, primarily:
  - Use the `new` operator to simplify object creation.
  - Pattern matching - `x is not y` instead of `!(x is y)`, `x is not null` instead of `x is object`.
  - Marking things `static` that don't access instance data (really only in test fixtures).

A future PR can do some cleanup like setting up `ImplicitUsings` and file-scoped namespaces.


tillig:
This PR supersedes #1300.

src/Autofac/Builder/BuildCallbackManager.cs
```diff
@@ -11,7 +11,7 @@ namespace Autofac.Builder
     /// </summary>
     internal static class BuildCallbackManager
     {
-        private static readonly TypedService CallbackServiceType = new TypedService(typeof(BuildCallbackService));
+        private static readonly TypedService CallbackServiceType = new(typeof(BuildCallbackService));
```
alistairjevans, line 14: I think this is the sort of `new()` usage that confuses me. I've generally only used `new()` for default constructors, because my brain melts when I try to read it with parameters, but I'll get used to it if you think we should standardise on the pattern.

src/Autofac/Core/Resolving/Pipeline/PipelineBuilderEnumerator.cs
```diff
@@ -39,16 +39,16 @@ public bool MoveNext()
                 return false;
             }
 
-            if (_current is object)
+            if (_current is not null)
             {
                 _current = _current.Next;
 
-                _ended = !(_current is object);
+                _ended = !(_current is not null);
```
alistairjevans, line 46: Should this not be `_ended = _current is null`?

test/Autofac.Specification.Test/Util/AsyncOnlyDisposeTracker.cs
```diff
@@ -28,6 +28,7 @@ public async ValueTask DisposeAsync()
 
             IsAsyncDisposed = true;
             Disposing?.Invoke(this, EventArgs.Empty);
+            GC.SuppressFinalize(this);
```
alistairjevans, line 31: Feels weird to add performance optimisation calls to a test-only implementation of a dispose tracker, no? Are we sacrificing test readability just to satisfy the analyser?

I have in the past added a `TestSuppressions.cs` file to test projects where I want to add assembly-level suppressions without putting them in the editorconfig?

test/Autofac.Test/Core/Activators/ProvidedInstance/ProvidedInstanceActivatorTests.cs
```diff
@@ -20,9 +20,9 @@ public void NullIsNotAValidInstance()
         [Fact]
         public void WhenInitializedWithInstance_ThatInstanceIsReturnedFromActivateInstance()
         {
-            object instance = new object();
+            object instance = new();
 
-            ProvidedInstanceActivator target = new ProvidedInstanceActivator(instance);
+            ProvidedInstanceActivator target = new(instance);
```
alistairjevans, line 25: For method fields, I'd usually lean towards using `var` and leaving the type on the other side rather than put the type at the front this way...possibly personal preference.

test/Autofac.Test/SourceRegistrationExtensionsTests.cs
```diff
@@ -26,7 +26,7 @@ public void RegisterSource()
                 Assert.True(source.RegistrationsForCalled);
             }
 
-            new ObjectRegistrationSource();
+            _ = new ObjectRegistrationSource();
```
alistairjevans, line 29: Does this line actually do anything? I might be missing it.

test/Autofac.Test/TypeExtensionsTests.cs
```diff
@@ -170,15 +170,18 @@ private DeclaredConstructorType(bool parameter)
 
         private class DeclaredMethodType
         {
+            [SuppressMessage("CA1822", "CA1822", Justification = "Method needs to be instance for testing.")]
```
alistairjevans, line 173: These test static warnings feel like a candidate for an assembly-level suppression.

alistairjevans (changes requested):
Only 1 real thing that should change, that's the !(not null) check which can be simplified. The rest are mostly just thoughts, options, etc.

I'll try not to take the British English spelling removal too personally... 😂

test/Autofac.Specification.Test/Util/AsyncOnlyDisposeTracker.cs
tillig: We could have two different analyzer.ruleset files, one for production code, one for tests. I've done that for other projects.

I didn't think this really affected readability so much and... I mean, it _is_ disposing, so _technically_ it's correct, right?

I don't feel terribly strongly about it. It might be worth having two different ruleset files just to add the flexibility.

test/Autofac.Test/SourceRegistrationExtensionsTests.cs
tillig: Yeah, this is weird. I'll remove that.

src/Autofac/Builder/BuildCallbackManager.cs
tillig: I fixed it so `new()` like this is only in class field sorts of things and not in methods. Agreed, this looks weird. Kinda what I get for "fix all" with the analyzers.

src/Autofac/Core/Resolving/Pipeline/PipelineBuilderEnumerator.cs
tillig: Uh. Y... yes? 😆  Fixed.

test/Autofac.Specification.Test/Util/AsyncOnlyDisposeTracker.cs
tillig: I added `Test.ruleset` with some of the various suppressions that were around the test projects. It should clean up a bit of the `#pragma` and other suppression noise in there. Also removed the `GC` stuff here and added that rule to the ruleset so it won't be enforced in tests.

test/Autofac.Test/TypeExtensionsTests.cs
tillig: Fixed with the Test.ruleset.

alistairjevans (approved):
I have decided your spelling changes are not an attack on my Britishness, so we're tickety boo governor!

maulik-modi:
@alistairjevans or @tillig , When do you plan to publish Nuget package?

tillig:
@maulik-modi Is there something here that your code requires? The existing package should work perfectly well with .NET 6.

maulik-modi:
.net 6 being major version, may be  breaking changes requiring new version of autofac.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

tillig:
There's are no breaking changes. Instead of guessing, try it out and let us know if there's something you find not working.

PR #1307, opened 2022-02-23
File-scoped namespaces and implicit usings

tillig:
Now that we are on the .NET 6 bandwagon, a little housekeeping:

- Enabled implicit usings on all projects, including adding a couple (e.g., `Moq` and `Xunit` in unit test projects) to clean up a lot of the fluff at the top of every file.
- Updated all files to file-scoped namespaces, which removes a bunch of whitespace at the start of every line and some open/close curly brackets... or braces... or [whatever you call them where you are](https://twitter.com/davidfowl/status/1467613566509350912).


alistairjevans:
Only annoying thing about the file scoped namespaces (I discovered after doing it on another project), is that git can't figure out that all you've done is change the indentation, so it effectively erases all git blame for every file. A minor annoyance perhaps, but worth remembering.

tillig:
This is true. I'm not sure how to get around that, though, and it seems like even if we put it off, if _anyone_ switches _any_ file at some point in the future, the same thing will happen. Comes down to "we don't do it at all, ever" or "rip off the bandage" I guess.

I'm for ripping the sucker off and going for it, but I won't [get involved in a land war in Asia](https://www.rottentomatoes.com/m/princess_bride/quotes/#:~:text=The%20most%20famous%20of%20which%20is%20%22never%20get%20involved%20in,DEATH%20is%20on%20the%20line) over it.

alistairjevans:
I also agree; but remember, never challenge a Sicilian when death is on the line!

alistairjevans (approved):
I have absolutely 100% been through all these files to check them.

tillig:
✅ Bandage ripped off

PR #1308, opened 2022-02-24
Add readme to nuget package and update README for compatibility with nuget.org

### LINKED ISSUE #1295: Include the README in Autofac NuGet packages
## Problem Statement

As of .NET 5 / NuGet 5.10.0 preview 2, you can [include your README in the NuGet package](https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packagereadmefile) to get something richer showing up on the NuGet.org site. Without that, you just get the package description and it looks very sad and lonely in there.

## Desired Solution

- Update the README to make sure it works both on the GitHub site and on NuGet.org. I'm not sure I can quantify everything, but, for example...
  - Links to things probably need to be absolute instead of relative to the GitHub repo.
  - Only use the [image sources and markdown features](https://docs.microsoft.com/en-us/nuget/nuget-org/package-readme-on-nuget-org) they support. Double check that things like badges and the 'Open in VS Code' thing are all on the approved list and are using full URLs.
- Update the .csproj as shown [in the instructions](https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packagereadmefile) to include the README.
- Possibly test this using a manual upload first, to see what it's going to look like. They talk about how to do that on the NuGet.org site. We don't want to iterate over this by actually releasing a bunch of packages.

I don't think this is worth pushing a whole new version for. I think we should just get it ready and then when the next release goes out, this will just be there.

If there are other packages going out sooner, it might be worth doing those first so we get some experience.

alistairjevans:
I've updated the README for compatibility with nuget.org, and added it to the package spec. This is how it looks in the nuget upload preview:

image

Annoyingly, making the README work for nuget.org has possibly made our Github README worse, as follows:

- Images cannot be center aligned.
- I have had to add fixed width/height variants of our logos to the autofac.github.com repository because I can't specify the size in MD.
- The myget and "open in vscode" badges do not work, so I have removed.

I'm contemplating just having a separate README file used only for nuget.org, given how infrequently we change it, so we don't have to take that stuff out. If that seems sensible to others, I will update the PR accordingly.

Closes #1295

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,11 @@
-<p align="center">
-  <img src="https://autofac.org/img/autofac_web-banner_character.svg" width="250">
-</p>
-<p align="center">
-  <img src="https://autofac.org/img/autofac_logo-type.svg" height="100">
-</p>
+![autofac character](https://raw.githubusercontent.com/autofac/autofac.github.com/8737b1213a85ad8157ab7958aeb560c7af5eb494/img/autofac_web-banner_character_fixed_width.svg)
+![autofac logo](https://raw.githubusercontent.com/autofac/autofac.github.com/8737b1213a85ad8157ab7958aeb560c7af5eb494/img/autofac_logo-type_fixed_height.svg)
 
 Autofac is an [IoC container](http://martinfowler.com/articles/injection.html) for Microsoft .NET. It manages the dependencies between classes so that **applications stay easy to change as they grow** in size and complexity. This is achieved by treating regular .NET classes as *[components](https://autofac.readthedocs.io/en/latest/glossary.html)*.
 
-[![Build status](https://ci.appveyor.com/api/projects/status/s0vgb4m8tv9ar7we?svg=true)](https://ci.appveyor.com/project/Autofac/autofac) [![codecov](https://codecov.io/gh/Autofac/Autofac/branch/develop/graph/badge.svg)](https://codecov.io/gh/Autofac/Autofac) ![MyGet publish status](https://www.myget.org/BuildSource/Badge/autofac?identifier=e0f25040-634c-4b7d-aebe-0f62b9c465a8) [![NuGet](https://img.shields.io/nuget/v/Autofac.svg)](https://nuget.org/packages/Autofac)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,15 +1,11 @@
-<p align="center">
-  <img src="https://autofac.org/img/autofac_web-banner_character.svg" width="250">
-</p>
-<p align="center">
-  <img src="https://autofac.org/img/autofac_logo-type.svg" height="100">
-</p>
+![autofac character](https://raw.githubusercontent.com/autofac/autofac.github.com/8737b1213a85ad8157ab7958aeb560c7af5eb494/img/autofac_web-banner_character_fixed_width.svg)
```
tillig, line 1:
```suggestion
![Autofac character](https://raw.githubusercontent.com/autofac/autofac.github.com/8737b1213a85ad8157ab7958aeb560c7af5eb494/img/autofac_web-banner_character_fixed_width.svg)
```

I know, picky, but Autofac is a proper noun so the Grammar Police showed up 👮

README.md
```diff
@@ -1,15 +1,11 @@
...
+![autofac logo](https://raw.githubusercontent.com/autofac/autofac.github.com/8737b1213a85ad8157ab7958aeb560c7af5eb494/img/autofac_logo-type_fixed_height.svg)
```
tillig, line 2:
```suggestion
![Autofac logo](https://raw.githubusercontent.com/autofac/autofac.github.com/8737b1213a85ad8157ab7958aeb560c7af5eb494/img/autofac_logo-type_fixed_height.svg)
```

tillig (changes requested):

tillig:
It's weird that those other badges don't work. Huh. I mean, I'd have been totally OK if that Gitter badge accidentally got lost 🙈, but it seems weird that MyGet doesn't work.

I think if we have two different READMEs we'll forget that one is there. I'm OK with these minor changes.

I'm guessing it's just the images that won't render - what if we made them just... text links instead of "badges?" I mean, there's technically already a link to the MyGet stuff, so that's taken care of. Maybe under the "Contributing / Pull Requests" there could be something about "or [open it right now in VS Code]".

alistairjevans:
OK, grammar corrected, VS Code reference added to the end.

alistairjevans:
OK, grammar corrected, VS Code reference added to the end.

tillig (approved):
🌭

PR #1309, opened 2022-02-24
Fix #995: Publish symbols as .snupkg

### LINKED ISSUE #995: Switch Autofac build to use .snupkg instead of embedding the .pdb directly
Autofac version affected: 4.9.x

Description:
The .net 4.5 folder of the nuget package appears to erroneously have the portable PDB, not the full pdb.  The result is a build error on legacy TFS servers.

Please consider that Json.net / Newtonsoft had a similar issue when they went to portable PDBs, and they reverted to the full PDB for the .Net 4.5 dll.

It would be great if the full debug PDB could be used again for the .Net 4.5+ targeting dll, as that would relieve this build error for 4.9.x and higher.

(Obviously the portable one is fine for the .Net Standard 1.1/2.0 versions).

tillig:
Updates the build to produce a separate NuGet symbols package (.snupkg) so the .pdb is no longer embedded right in the main package. Reduces the size of the main package by over half (~2MB down to ~800KB).

I also organized the .csproj properties into logical sections because it was becoming quite the dumping ground. It was hard to see if things were already set or not.


tillig:
This might cause a conflict with #1308 because of the addition of the PackageReadme element in the .csproj. Probably should merge that first so I can take the update in here.

tillig:
OK, got that README thing merged in.

alistairjevans (approved):
Looks good; snupkg is generated by `dotnet pack`. 

Will we need to change anything about our release process in MyGet when we push to NuGet? Does MyGet know to carry the snupkg files along with it?

tillig:
In MyGet, what you'll see is there are _two_ download links - one for the package, one for the symbols. You'll also see the symbol server information right below the NuGet command line info, just above the release notes.

MyGet has both package and symbol downloads

When you go to push the package to NuGet, you need to make sure you check the "Include matching symbols packages" box.

Include matching symbols packages

When it makes it over to NuGet, NuGet also has corresponding "download package/download symbols" links.

We'll have to drop both into each release as artifacts instead of just the library package.

PR #1310, opened 2022-02-25
Update to use the latest Codecov Uploader and SHA validation

tillig:
The bash version of the Codecov Uploader [has been retired and the new version is a binary](https://docs.codecov.com/docs/codecov-uploader) you can download and check SHA hashes to verify integrity.

I've updated the build script to use this new uploader and fail if the checksums aren't correct.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -66,9 +66,19 @@ try {
 
     if ($env:CI -eq "true") {
         # Generate Coverage Report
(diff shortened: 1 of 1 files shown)


alistairjevans (approved):
Looks like codecov have taken steps to protect against the supply chain attacks.

PR #1311, opened 2022-02-28
Corrected links, removed deprecated package listings.

tillig:
- Direct link to the Autofac core package with a separate link to the set of all packages - get people started quickly with maybe less confusion.
- Release notes are on GitHub now, and while you can get to _all_ the package release notes through our documentation site, now that the README is also in the NuGet package it seemed like sending folks directly to the Autofac notes made sense.
- Removed links to packages that are in maintenance mode and really aren't supported anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,9 +9,9 @@ Autofac is an [IoC container](http://martinfowler.com/articles/injection.html) f
 
 ## Get Packages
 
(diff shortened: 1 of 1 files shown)


alistairjevans (approved):
🧾

PR #1316, opened 2022-03-18
Use base type for generic argument matching

### LINKED ISSUE #1315: ComponentNotRegisteredException when type parameter name doesn't match the base class
## Describe the Bug

Registering a component with `RegisterGeneric` when the service is a class where the type parameter name is different,
Autofac throws a `ComponentNotRegisteredException`, unless the component has been resolved explicitly first.

` With_as_self_and_explicit_resolve` succeeds because `container.Resolve<NHibernateRepository<Task>>();` is called before `container.Resolve<Repository<Task>>();`.

If `Repository<T>` is an interface, then it succeeds.

If the type parameter in `NHibernateRepository<TDomain>` is renamed to match `Repository<T>`, it also succeeds.

## Steps to Reproduce

```c#
public class ReproTest
{
    public class Task
    {
    }

    // Changing this to an interface works without problems.
    // Making it an abstract class also fails
    public class Repository<T>
    {
    }

    // Renaming TDomain to T also fixes the problem
    public class NHibernateRepository<TDomain> : Repository<TDomain>
    {
    }

    [Fact]
    public void With_as_self_and_explicit_resolve()
    {

        builder
            .RegisterGeneric(typeof(NHibernateRepository<>))
            .AsSelf()
            .As(typeof(Repository<>));

        // Resolving the component explicitly fixes the problem
        container.Resolve<NHibernateRepository<Task>>();
        container.Resolve<Repository<Task>>();
    }

    [Fact]
    public void With_as_self_and_without_explicit_resolve()
    {

        builder
            .RegisterGeneric(typeof(NHibernateRepository<>))
            .AsSelf()
            .As(typeof(Repository<>));

        container.Resolve<Repository<Task>>();
    }

    [Fact]
    public void Without_as_self_and_without_explicit_resolve()
    {

        builder
            .RegisterGeneric(typeof(NHibernateRepository<>))
            .As(typeof(Repository<>));

        container.Resolve<Repository<Task>>();
    }
}

```

## Expected Behavior

Resolving `Repository<T>` should return `NHibernateRepository<TDomain>` in all 3 cases.

## Exception with Stack Trace

```text
Autofac.Core.Registration.ComponentNotRegisteredException
The requested service 'Autofac.Test.ReproTest+Repository`1[[Autofac.Test.ReproTest+Task, Autofac.Test, Version=1.0.0.0, Culture=neutral, PublicKeyToken=17863af14b0044da]]' has not been registered. To avoid this exception, either register a component to provide the service, check for service registration using IsRegistered(), or use the ResolveOptional() method to resolve an optional dependency.
   at Autofac.ResolutionExtensions.ResolveService(IComponentContext context, Service service, IEnumerable`1 parameters) in C:\Projects\Autofac\src\Autofac\ResolutionExtensions.cs:line 874
   at Autofac.ResolutionExtensions.Resolve(IComponentContext context, Type serviceType, IEnumerable`1 parameters) in C:\Projects\Autofac\src\Autofac\ResolutionExtensions.cs:line 338
   at Autofac.ResolutionExtensions.Resolve[TService](IComponentContext context, IEnumerable`1 parameters) in C:\Projects\Autofac\src\Autofac\ResolutionExtensions.cs:line 290
   at Autofac.ResolutionExtensions.Resolve[TService](IComponentContext context) in C:\Projects\Autofac\src\Autofac\ResolutionExtensions.cs:line 273
   at Autofac.Test.ReproTest.Without_as_self() in C:\Projects\Autofac\test\Autofac.Test\ReproTest.cs:line 59
```

## Dependency Versions

Autofac: 6.3.0
Autofac: The develop branch as of March 18th 2022

tillig:
Fix #1315.

When you call `GetInterfaces()` on a `Type`, it returns the set of implemented interfaces but has already swapped the generic type argument names in for the ones used on the implementing `Type`. For example, if you have:

```c#
public interface IThing<T1> { }
public class Thing<TFirst> : IThing<TFirst> { }
```

...then when you call `typeof(Thing<>).GetInterfaces()` the result will be `IThing<TFirst>` - the name of the argument matches the implementer's name, not the one in the actual interface definition.

We use this naming to line up the type parameters on generic types and service type interface implementations.

For classes, that's not what we were doing.

```c#
public class BaseClass<T1> { }
public class DerivedClass<TFirst> : BaseClass<TFirst> { }
```

In the class case, we were basically doing `typeof(DervivedClass<>).GetGenericArguments()` and `typeof(BaseClass<>).GetGenericArguments()` and _then_ trying to line up names. The problem is, the name had changed in the implementation so things wouldn't line up.

Instead of doing `typeof(BaseClass<>).GetGenericArguments()`, we need to _walk back in the class hierarchy_ to find where `DerivedClass<TFirst>` actually derives from `BaseClass<>` so we can make use of the compiler having already renamed all the generic arguments for us... _then_ we can match on name.

This PR does exactly that - walks back the inheritance hierarchy until it finds the derived class and then does the argument matching.

I added tests to show it'll work more than just one derived class deep, proving the updated name mapping works all the way down the stack.

### DIFF
--- a/src/Autofac/Features/OpenGenerics/OpenGenericServiceBinder.cs
+++ b/src/Autofac/Features/OpenGenerics/OpenGenericServiceBinder.cs
@@ -173,7 +173,18 @@ internal static class OpenGenericServiceBinder
 
         if (!serviceType.IsInterface)
         {
-            return TryFindServiceArgumentsForImplementation(implementationType, serviceGenericArguments, serviceTypeDefinition.GetGenericArguments());
+            var baseType = GetGenericBaseType(implementationType, serviceTypeDefinition);
+            if (baseType == null)
+            {
+                // If it's not an interface, the implementation type MUST have derived from
+                // the generic service type at some point or there's no way to cast.
+                return Array.Empty<Type>();
+            }
+
+            return TryFindServiceArgumentsForImplementation(
+                    implementationType,
+                    serviceGenericArguments,
+                    baseType.GenericTypeArguments);
         }
 
         var availableArguments = GetInterfaces(implementationType, serviceType)
@@ -197,6 +208,22 @@ internal static class OpenGenericServiceBinder
             .Select(implementationGenericArgumentDefinition => TryFindServiceArgumentForImplementationArgumentDefinition(
                 implementationGenericArgumentDefinition, serviceArgumentDefinitionToArgumentMapping))
             .ToArray();
+    }
+
+    private static Type? GetGenericBaseType(Type implementationType, Type serviceTypeDefinition)
+    {
+        var baseType = implementationType.BaseType;
+        while (baseType != null)
+        {
+            if (baseType.IsGenericType && baseType.GetGenericTypeDefinition() == serviceTypeDefinition)
+            {
+                break;
+            }
+
+            baseType = baseType.BaseType;
+        }
+
+        return baseType;
     }
 
     private static Type[] GetInterfaces(Type implementationType, Type serviceType) =>
--- a/src/Autofac/Features/OpenGenerics/OpenGenericServiceBinder.cs
+++ b/src/Autofac/Features/OpenGenerics/OpenGenericServiceBinder.cs
@@ -179,7 +179,12 @@ internal static class OpenGenericServiceBinder
              *
              * If you have a class BaseClass<T1, T2> and a derived class
              * DerivedClass<A1, A2> : BaseClass<A2, A1> then then arguments need
-             * to line up name-wise like A2 -> T1 A1 -> T2
+             * to line up name-wise like...
+             * A2 -> T1
+             * A1 -> T2
+             *
+             * Having those symbols aligned means that, later, we can do a
+             * name-based match to populate the generic arguments.
              *
              * If you do a typeof(BaseClass<,>) then the generic arguments are
              * named T1 and T2, which doesn't help us line up the arguments in
@@ -187,8 +192,7 @@ internal static class OpenGenericServiceBinder
              * However, if you start at the derived class and walk backwards up
              * the inheritance hierarchy then instead of getting the original
              * T1/T2 naming, we get the names A1/A2 - the symbols as lined up by
-             * the compiler. Having those symbols aligned means that, later, we
-             * can do a name-based match to populate the generic arguments.
+             * the compiler.
              *
              * This is the same reason why Type.GetInterfaces() "just works" -
              * it's the generic in relation to the derived/implementing type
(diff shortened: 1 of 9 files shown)


tillig:
Oh, I also found a typo in some of the test classes' namespaces `Contraint` instead of `Constraint` so I fixed that. The PR looks a little bigger than it actually is because of that.

src/Autofac/Features/OpenGenerics/OpenGenericServiceBinder.cs
```diff
@@ -173,7 +173,18 @@ internal static class OpenGenericServiceBinder
 
         if (!serviceType.IsInterface)
         {
-            return TryFindServiceArgumentsForImplementation(implementationType, serviceGenericArguments, serviceTypeDefinition.GetGenericArguments());
+            var baseType = GetGenericBaseType(implementationType, serviceTypeDefinition);
```
alistairjevans, line 176: Can I request a comment here explaining 'why' we have to do this? This generic stuff is already pretty complex, and without scouting through the tests/issues, not sure it will be clear (perhaps just tag the issue number in a comment)?

alistairjevans (approved):
Approved, but with two minor queries if you're up for them. Generics are complicated....

src/Autofac/Features/OpenGenerics/OpenGenericServiceBinder.cs
```diff
@@ -173,7 +173,18 @@ internal static class OpenGenericServiceBinder
...
+            if (baseType == null)
+            {
+                // If it's not an interface, the implementation type MUST have derived from
+                // the generic service type at some point or there's no way to cast.
+                return Array.Empty<Type>();
```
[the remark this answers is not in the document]
tillig: I don't know if we can unless something had gone buggy in the checks where we determine if the thing in RegisterGeneric and As are compatible but I can try walking backwards through it to see if we'd ever get there.

src/Autofac/Features/OpenGenerics/OpenGenericServiceBinder.cs
tillig: Totally fair and excellent point. I'll get something in there.

tillig: Minor clarifying manifesto added to explain what's going on here and why.

alistairjevans: Thanks, very useful comment.

alistairjevans (approved):
🐲

tillig:
Thanks! I couldn't figure out a good way to test that one code path yet. My gut says it'll never get hit. Not sure I'll have time to figure a test in the super near future, so good to not hold it up.

alistairjevans:
Yeah, probably don't need a test for what is sort of an Assert statement that only trips if something else has gone horribly wrong.

PR #1318, opened 2022-03-24
Upgrade AsyncInterfaces package to v6.0.0

rcdailey:


alistairjevans:
Why is this change needed? Is there an issue you're trying to resolve? 

The package you're changing is only needed for netstandard20, all subsequent versions of the SDK come with it in the box.

alistairjevans:
Why is this change needed? Is there an issue you're trying to resolve? 

The package you're changing is only needed for netstandard20, all subsequent versions of the SDK come with it in the box.

rcdailey:
I have a series of .NET 4.8 projects. Because Autofac uses v1.0.0 of this package, and another project uses v6.0.0, I now have mixed versions. I won't go into more detail than that, but having Autofac use the latest available version resolves that issue *and* I assert just makes sense in general. I think the question should be, "Why is this change *not* needed?".

Thanks for taking the time to review this!

alistairjevans:
This change as proposed forces a dependency on an unnecessary package for _all_ consumers of the package, because you've taken it out of the conditional netstandard20 block.  As per the [nuget readme](https://www.nuget.org/packages/Microsoft.Bcl.AsyncInterfaces/), the package is not required at all for more up-to-date consumers of the package.

Upgrading to 6.0.0 of package doesn't bring any tangible benefit; the package is simply a collection of interfaces forwarded to runtime types in later versions of the 4.x Framework, so there are no bug fixes to be gained from upgrading as far as I'm aware. Also, upping the dependency version to 6.0.0 means for users of the package that _are_ on .NET Framework, they have to now reference that latest version.

If memory serves, you can use binding redirects quite easily in your projects to redirect AsyncInterfaces 1.1.0 to 6.0.0, and your problem goes away, without changing Autofac dependencies.

tillig:
It's _always_ "why do we need it?" vs "why not?" for compatibility reasons. [There's an FAQ on it](https://autofac.readthedocs.io/en/latest/faq/binding-redirect.html) and it's [in our contributors guide.](https://github.com/autofac/.github/blob/master/CONTRIBUTING.md#dependencies-and-upgrades)

rcdailey:
I elided details about my situation because I didn't feel they were relevant, but certain assertions are being made here that I'm not sure help me out, so I'll share a bit more detail.

The two projects I referred to are not connected dependencies. They are independent of each other, but the conflict comes in because the binaries for both applications are sent to the same output directory. So because of this, there's a race condition between AsyncInterfaces v1.0 and v6.0. I don't believe binding redirects solve this particular issue.

Yes this is a "me" problem. But I figured upgrading wouldn't hurt here, and it was easy to do. If it's just a collection of interfaces, I'm not sure what the harm is in upgrading it. That's not me trying to selfishly push for it, but rather just an attempt at being practical. Obviously I'm wrong here, as pointed out in the FAQ linked above.

As for the removal of the condition, that was not intentional. When I used Jetbrains Rider to upgrade the version of the package, it took the liberty of removing the condition for whatever reason. I did notice that, but I didn't realize it would be a problem.

I'm going to close this PR as obviously I don't have a "technical reason" for upgrading the package. I also have very limited time due to already having too many projects I contribute to, so I'll end things here and close the PR.

I appreciate everyone taking the time to roast me ;-)

alsami:
> I have a series of .NET 4.8 projects. Because Autofac uses v1.0.0 of this package, and another project uses v6.0.0, I now have mixed versions. I won't go into more detail than that, but having Autofac use the latest available version resolves that issue _and_ I assert just makes sense in general. I think the question should be, "Why is this change _not_ needed?".
> 
> Thanks for taking the time to review this!

As already mentioned, the solution for this is to add binding redirects to your dotnet framework project.
This has been a problem with legacy dotnet framework projects since like forever and it's the only and right way to work around it. 

https://docs.microsoft.com/en-us/dotnet/framework/configure-apps/redirect-assembly-versions

rcdailey:
> > I have a series of .NET 4.8 projects. Because Autofac uses v1.0.0 of this package, and another project uses v6.0.0, I now have mixed versions. I won't go into more detail than that, but having Autofac use the latest available version resolves that issue _and_ I assert just makes sense in general. I think the question should be, "Why is this change _not_ needed?".
> > Thanks for taking the time to review this!
> 
> As already mentioned, the solution for this is to add binding redirects to your dotnet framework project. This has been a problem with legacy dotnet framework projects since like forever and it's the only and right way to work around it.
> 
> [docs.microsoft.com/en-us/dotnet/framework/configure-apps/redirect-assembly-versions](https://docs.microsoft.com/en-us/dotnet/framework/configure-apps/redirect-assembly-versions)

I'm trying to tell you this isn't the solution to my problem. Go back and *carefully* read my last comment again.

tillig:
Try adding a direct reference to the 6.0.0 interfaces package in your project.

PR #1321, opened 2022-03-31
Add generic delegate lambda overloads, and associated tests

### LINKED ISSUE #1320: Feature Proposal - Generic Overloads for Register to make delegate registration more concise
## Problem Statement

Today I was registering some Autofac components, and realised I wanted to capture a configuration object, and pass that into my component constructor, along with it's dependencies. 

So, something like this:

```csharp
interface IMyDependency1 { } 
interface IMyDependency2 { }

class ConfigOptions { }

class MyComponent 
{
  public MyComponent(ConfigOptions opt, IMyDependency1 dep1, IMyDependency2 dep2) 
  {
  }
}

// Registering (in a module, but that's an implementation detail)
var myOpts = new ConfigOptions 
{
   // Populate my options from wherever
};

builder.Register(ctxt => new MyComponent(
  myOpts, 
  ctxt.Resolve<IMyDependency1>(), 
  ctxt.Resolve<IMyDependency2>());
```

All good, works as expected, but constructing that instance of `MyComponent` feels reallly verbose, since I need to call `ctxt.Resolve` each time.  It also smacks of service location in a way, because we're accessing an autofac service locator directly:

```csharp
// Ugh, noisy, more generic args, etc.
builder.Register(ctxt => new MyComponent(
  myOpts, 
  ctxt.Resolve<IMyDependency1>(), 
  ctxt.Resolve<IMyDependency2>());
```

## Desired Solution

What I'd like to be able to do is this:

```csharp
// Just declare your dependencies as arguments to the delegate, just as you would with a constructor.
builder.Register((IMyDependency1 dep1, IMyDependency2 dep2) 
  => new MyComponent(myOpts, dep1, dep2));
```

Personally I feel like this brings things closer to feeling like a normal factory method, plus it has the advantage that I can treat any method that accepts dependencies as a factory method, without leaking Autofac types into it.

I also have to know less Autofac 'stuff', like a component context, to get this working.

But, I'd like to retain the option to access the full component context if I need to do something a bit more complex:

```csharp
builder.Register((IComponentContext ctxt, IMyDependency2 dep2) 
  => new MyComponent(ctxt.ResolveKeyed<IMyDependency1>("key"), dep2));
```

## Implementation

The implementation is actually pretty simple; we pick a maximum number of generic parameters we want to support; let's say 10 to start, we then create a set of methods for each number of parameters we want to support. Here are the new extension methods for my example with 2 parameters:

```csharp
public static IRegistrationBuilder<TComponent, SimpleActivatorData, SingleRegistrationStyle>
    Register<TParam1, TParam2, TComponent>(
        this ContainerBuilder builder,
        Func<TParam1, TParam2, TComponent> @delegate)
    where TParam1 : notnull
    where TParam2 : notnull
    where TComponent : notnull
{
    if (@delegate == null)
    {
        throw new ArgumentNullException(nameof(@delegate));
    }

    return builder.Register((c, p) => @delegate(c.Resolve<TParam1>(), c.Resolve<TParam2>()));
}

// Note, here we can include an `IComponentContext` parameter first so users can blend both types if they need to.
public static IRegistrationBuilder<TComponent, SimpleActivatorData, SingleRegistrationStyle>
    Register<TParam1, TParam2, TComponent>(
        this ContainerBuilder builder,
        Func<IComponentContext, TParam1, TParam2, TComponent> @delegate)
    where TParam1 : notnull
    where TParam2 : notnull
    where TComponent : notnull
{
    if (@delegate == null)
    {
        throw new ArgumentNullException(nameof(@delegate));
    }

    return builder.Register((c, p) => @delegate(c, c.Resolve<TParam1>(), c.Resolve<TParam2>()));
}
```

Granted, it would be a fair number of extra overloads, but it doesn't seem too onerous, since the methods themselves are lightweight. There are prior examples of this in the runtime (e.g. [HashCode.Combine](https://docs.microsoft.com/en-us/dotnet/api/system.hashcode.combine?view=net-6.0)), and I've done it in a couple of private projects.

I tested the edges of this solution a little to see if there were obvious pitfalls.  One of the nice things (of sorts) is that the compiler won't just 'infer' delegate argument types automatically, you need to specify them.  On top of that, if you specify one, you specify all. So, this largely prevents clashes with our existing overloads:

```csharp
// Existing overloads don't break, because a) non-generic methods are preferred, and b) the type isn't specified as an argument.
builder.Register(ctxt => new Component(ctxt.ResolveKeyed<IFancyDep>("val"));

// Compiler error here, types cannot be inferred, must be specified explicitly. 
// I kind of like it because it makes it mandatory to specify your dependencies.
builder.Register((dep1, dep2, dep3) => new Component(dep1, dep2, dep3));

// Compiler error here because arg1 is IComponentContext and arg2 is IEnumerable<Parameter>, because I didn't specify,
// so I got the non-generic form.
builder.Register((arg1, arg2) => new Component(arg1, arg2));

// Compiler error here, can't mix and match.
builder.Register((dep1, IMyDependency2 dep2, dep3) => new Component(dep1, dep2, dep3));
```

## Alternatives You've (not) Considered

It's entirely possible that this has been suggested before by someone else, and dismissed because of some glaring thing I've missed.

alistairjevans:
Closes #1320 

I'll look at the optimisation for the reflection activator with matching constructor later, thought I'd get this up there.

Aside; I accidentally pushed these changes directly to develop first, and had to revert. 🤦‍♂️


src/Autofac/RegistrationExtensions.Delegates.cs
```diff
@@ -0,0 +1,817 @@
...
+using Autofac.Features.Scanning;
+using Autofac.Util;
+
+namespace Autofac;
+
+/// <summary>
+/// Adds registration syntax to the <see cref="ContainerBuilder"/> type.
+/// </summary>
+[SuppressMessage("Microsoft.Maintainability", "CA1506:AvoidExcessiveClassCoupling")]
+public static partial class RegistrationExtensions
```
tillig, line 23: I don't have a strong preference, but we had to split that one giant `RegistrationExtensions` class into files because we couldn't break binary compatibility and actually put them in different classes. Given these are net-new, is it worth actually having a different extensions class? I mean, _technically_ it doesn't matter, I guess, just thinking out loud.

src/Autofac/RegistrationExtensions.Delegates.cs
```diff
@@ -0,0 +1,817 @@
...
+            Func<IComponentContext, TDependency1, TComponent> @delegate)
+        where TDependency1 : notnull
+        where TComponent : notnull
+    {
+        if (@delegate == null)
+        {
+            throw new ArgumentNullException(nameof(@delegate));
+        }
+
+        return builder.Register((c, p) => @delegate(
```
tillig, line 45: As I read this I wonder if it's worth doing anything with the parameters in this new way. Working with parameters is sort of a pain, too. Like, it'd be cool to do this:

```c#
builder.Register((IDependency1 dep) => new Thing(dep));
```

And also have it work with this:

```c#
container.Resolve<Thing>(TypedParameter.From(myDep));
```

That is, right now the dependencies are _only_ resolved, where a full resolution operation also generally takes parameters into account.

That would obviously impact perf in a way such that the majority case (no parameters) would always be checking for something that isn't there and then falling back to do the right thing here. Probably something "opt-in" would be best if that feature were to be added.

```c#
builder.Register(
  (IDependency1 dep1) => new Thing(dep1),
  includeParameters: true);
```

Something like that maybe.

tillig (commented):
Looks good. A couple of questions in there in the spirit of brainstorming.

Also curious if it'd be worth putting something in the benchmark suite to compare reflection vs this method of registration to demonstrate that lambda is still faster (or even if there's some tipping point, like once you hit eight dependencies then reflection starts being equal or winning).  Definitely not required for this to go in, we could always add benchmarks later.

src/Autofac/RegistrationExtensions.Delegates.cs
alistairjevans: I think the reason they should be in the same class is that they have the same name (`Register`) as other extension methods, because of overload ambiguity detection; if the extension methods are in the same class, the compiler will add an error to the method definition that the overloads clash, but if they're in another class, the compiler will raise an "ambiguous" error on the caller of the method, which seems bad.

src/Autofac/RegistrationExtensions.Delegates.cs
alistairjevans: Oof; yeah, I see what you mean...and it would make it more consistent. Not super-straightforward though, but I'll take a look. There's already another overload that takes `IEnumerable<Parameter>()`.  

What I sort of want to happen (in a general sense) is that if I call `Resolve<TSomething>()` on a component context with parameters in it, that it looks at those parameters for a matching typed parameter first...but that would clash with the reflection activator, and wouldn't handle named delegate parameters. Hmm.

src/Autofac/RegistrationExtensions.Delegates.cs
tillig: Oh, that makes good sense. I didn't think of that.

src/Autofac/RegistrationExtensions.Delegates.cs
tillig: Well, the parameters only go to the thing being activated, so in this case it'd be the `Thing`. I imagine it'd be something like:

```c#
public static IRegistrationBuilder<TComponent, SimpleActivatorData, SingleRegistrationStyle>
  Register<TDependency1, TComponent>(
    this ContainerBuilder builder,
    Func<TDependency1, TComponent> @delegate,
    bool includeParameters = false)
  where TDependency1 : notnull
  where TComponent : notnull
{
  if (@delegate == null)
  {
      throw new ArgumentNullException(nameof(@delegate));
  }

  return builder.Register((c, p) => 
  {
    TDependency1 dep1;
    if(includeParameters && p.FirstOrDefault(param => param.CanSupplyValue(/*....*/, out valueProvider)) is Parameter val)
    {
      dep1 = valueProvider();
    }
    else
    {
      dep1 = c.Resolve<TDependency1>();
    }

    @delegate(dep1);
}
```

Doing that from memory, not compiler. But that's the basic concept; Ideally the bool check wouldn't cost much in the majority case but folks could opt in to supporting parameters.

I suppose it could default to always check the list of parameters, but that wouldn't be as fast if you know you're never calling `Resolve<T>(params)`.

alistairjevans: OK; I've been having a bit of an experiment this morning, and I think I've got a good solution for handing parameters that doesn't require an additional flag, and doesn't detract from the performance.

Here's the benchmark code:

```csharp
[GlobalSetup]
public void Setup()
{

    builder.RegisterType<MyDependency1>();

    // Existing method, capture the argument directly.
    builder.Register((MyDependency1 dep1) => new MyComponent(dep1, "arg"))
           .As<IServiceExistingMethodCapturedArg>();

    // New parameter-supporting method, but with capturing the argument.
    // Name is different just so I can run both side-by-side for now.
    builder.RegisterWithParamSupport((MyDependency1 dep1) => new MyComponent(dep1, "arg"))
           .As<IServiceNewMethodCapturedArg>();

    // New parameter-supporting method, but the argument comes from a resolve parameter.
    // Name is different just so I can run both side-by-side for now.
    builder.RegisterWithParamSupport((MyDependency1 dep1, string arg1) => new MyComponent(dep1, arg1))
           .As<IServiceNewMethodOpenArg>();

    // Check against the reflection way as well.
    builder
       .RegisterType<MyComponent>()
       .UsingConstructor(typeof(MyDependency1), typeof(string))
       .WithParameter(TypedParameter.From("arg"))
       .As<IServiceFromReflection>();

    _container = builder.Build();
}

[Benchmark]
public object ResolveExistingWithCapturedParameter()
{
    return _container.Resolve<IServiceExistingMethodCapturedArg>();
}

[Benchmark]
public object ResolveParamSupportWithCapturedParameter()
{
    return _container.Resolve<IServiceNewMethodCapturedArg>();
}

[Benchmark]
public object ResolveParamSupportWithProvidedTypedParameter()
{
    return _container.Resolve<IServiceNewMethodOpenArg>(TypedParameter.From("arg"));
}

[Benchmark]
public object ResolveParamSupportWithProvidedNamedParameter()
{
    return _container.Resolve<IServiceNewMethodOpenArg>(new NamedParameter("arg1", "value"));
}

[Benchmark]
public object ResolveReflectionWithRegisteredParameter()
{
    return _container.Resolve<IServiceFromReflection>();
}
```

And the results:

```markdown
|                                        Method |       Mean |   Error |  StdDev |  Gen 0 | Allocated |
|---------------------------------------------- |-----------:|--------:|--------:|-------:|----------:|
|          ResolveExistingWithCapturedParameter |   795.0 ns | 2.42 ns | 2.02 ns | 0.2861 |      1 KB |
|      ResolveParamSupportWithCapturedParameter |   795.5 ns | 5.75 ns | 4.49 ns | 0.2861 |      1 KB |
| ResolveParamSupportWithProvidedTypedParameter |   960.8 ns | 3.97 ns | 3.10 ns | 0.3548 |      1 KB |
| ResolveParamSupportWithProvidedNamedParameter |   958.9 ns | 3.21 ns | 3.00 ns | 0.3548 |      1 KB |
|      ResolveReflectionWithRegisteredParameter | 1,009.7 ns | 2.19 ns | 1.83 ns | 0.4025 |      2 KB |
```

As you can see, the new method that supports parameters is no slower than the existing method when the set of parameters is empty.

The downside? It's a good bit more code per-method (holding types, conditional checks, etc).  I'm tempted to add a little private Source Generator to auto-generate the methods for us...I've done a few source generators now, so it wouldn't be particularly tricky.

alistairjevans: Cool; I'll put a generator together over the weekend.

alistairjevans:
I've created a new project `Autofac.CodeGen` under a new `codegen` folder (src seemed the wrong place for it).

This is an "incremental" generator, so requires the latest SDK, which you'd need anyway to build Autofac; as far as I can tell it works fine under VSCode+OmniSharp and VS2022.  You may have problems in VS2019.

The _actual_ generated code files are not committed into source control, however I have added Verify-based snapshot testing for the generated code in a separate test project, `Autofac.Test.CodeGen`. There's only the one test, but it makes sure that as the build system/roslyn updates, our generated code stays the same. I've used this approach before for generators, and works pretty well.

Latest benchmarks with the new, generated code:

```markdown
|                                   Method |       Mean |    Error |   StdDev |  Gen 0 | Allocated |
|----------------------------------------- |-----------:|---------:|---------:|-------:|----------:|
|      ResolveManuallyWithComponentContext |   850.7 ns | 10.57 ns | 12.59 ns | 0.2861 |      1 KB |
|             ResolveWithCapturedParameter |   874.1 ns |  9.86 ns |  8.23 ns | 0.2861 |      1 KB |
|        ResolveWithProvidedTypedParameter | 1,081.7 ns | 11.02 ns |  9.77 ns | 0.3548 |      1 KB |
|        ResolveWithProvidedNamedParameter | 1,074.9 ns |  8.37 ns |  7.42 ns | 0.3548 |      1 KB |
| ResolveReflectionWithRegisteredParameter | 1,105.8 ns |  8.79 ns |  7.34 ns | 0.4025 |      2 KB |
```

bench/Autofac.Benchmarks/LambdaResolveBenchmark.cs
```diff
@@ -0,0 +1,99 @@
...
+           .WithParameter(TypedParameter.From("arg"))
+           .As<IServiceFromReflection>();
+
+        _container = builder.Build();
+    }
+
+    [Benchmark]
+    public object ResolveManuallyWithComponentContext()
+    {
+        return _container.Resolve<IServiceExistingMethodCapturedArg>();
```
tillig, line 41: I kinda like this over the `GC.KeepAlive` we have in other tests. I think we had `GC.KeepAlive` from before when BenchmarkDotNet supported returning objects.

tillig:
This could be "someone moved my cheese" level complaining, so if I'm just being grumpy or whatever definitely just say so - no offense will be taken. I'm just noticing that debugging generated code is sort of a pain, at least in VS Code. I can't find the generated source (it's not checked in and seems to maybe be in some sort of temp location?) so the only way to see it is to go to the unit tests and hit F12 to "Go To Definition." I can then get to the `Register` methods, but I can't get to the `DelegateInvokers` holder class from there.

No definition found for DelegateInvokers

I also tried to get some analyzer or compiler warnings out of the generated code and it was pretty hard. I never did quite get the order of operations down that needed to happen. I can see you have to change the analyzer and build (I think?) and _then_ change `Autofac.RegistrationExtensions` and build _again_ (I think?) but I never did really get it figured.

I did finally get it to issue an error during compilation, but when you do get an error, you can't navigate to the source to see what went wrong.

Trying to get a compiler/analyzer warning

**Is the experience better in Visual Studio?**

It may be worth adding, say, a tiny README to the Autofac.CodeGen folder that explains a little more about how it works or even just links to appropriate documentation. Where does the generated source _go_? What's the order of steps to make this regenerate the code? When should we expect to see compiler/analyzer warnings?

Again, if this is just me having to get used to something that other folks find pretty straightforward, let me know. I'm still hanging on to T4 templates and [CodeSmith Generator](https://www.codesmithtools.com/product/generator) and Reflection.Emit so this is pretty new to me.

alistairjevans:
Personally I've not used source generators much at all in VSCode.  Visual Studio is evidently somewhat better than VSCode, in that it does allow you to navigate the generated code quite well. It's still not good at picking up changes to the source generator.

Can you try adding `<EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>` to the Autofac csproj? It should put the generated files on-disk in the obj folder, might make it easier for OmniSharp to navigate to them?

I've found that the most effective/correct manner to definitely getting errors out of generated source is to run `dotnet build`; the CLI build is very accurate in building the latest source generator and using it in the referencing projects.

No problem adding a README file to the CodeGen project.

alistairjevans:
Oh, also, you didn't say in your comment, but if you are debugging, and you put a breakpoint in the (not generated) test method, can you step-in to the generated code in VSCode if you keep stepping in?

test/Autofac.Test.Compilation/Autofac.Test.Compilation.csproj
```diff
@@ -26,8 +26,11 @@
   </ItemGroup>
 
   <ItemGroup>
+    <Compile Remove="Helpers\**" />
```
tillig, line 29: It looks like the generator project used to be in here but maybe moved and resulted in some cruft showing up?

I admit, I had to go digging in git history to remember that `Autofac.Test.Compilation` was mostly about nullable references than about _compiling Autofac_ so maybe some clarification here might help at some point in the form of README files. I was about to suggest we combine the two since they're sort of both about compilation but I do understand they're about the _consumer of Autofac_ compiling vs _Autofac itself_ compiling so having them split also makes some sense.

test/Autofac.Test.CodeGen/Autofac.Test.CodeGen.csproj
```diff
@@ -0,0 +1,62 @@
...
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.1.0" />
+    <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.406">
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
+    <PackageReference Include="xunit" Version="2.4.1" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
+    </PackageReference>
+    <PackageReference Include="Verify.SourceGenerators" Version="1.3.0" />
```
tillig, line 54: Definitely not obvious that the `*.verified.cs` files [get ignored by the `Verify` package](https://github.com/VerifyTests/Verify.SourceGenerators/blob/b6f663bdcfad76a825f70994db5106a25456cce9/src/Verify.SourceGenerators/build/Verify.SourceGenerators.targets#L4). VS Code, at least, doesn't really like it if you open the `verified.cs` files because it then tries to analyze them and things like `System.Diagnostics.CodeAnalysis` aren't in `ImplicitUsings`. Which isn't a big deal, it just took me a bit to realize why I'd get errors if I _opened_ the file but not on _build_.

codegen/Autofac.CodeGen/DelegateRegisterGenerator.cs
```diff
@@ -0,0 +1,329 @@
...
+    /// <summary>
+    /// Register a delegate as a component.
+    /// </summary>");
+            WriteTypeParamDocs(sb, argCount, 1);
+            sb.AppendLine($@"    /// <typeparam name=""TComponent"">The type of the instance.</typeparam>
+    /// <param name=""builder"">Container builder.</param>
+    /// <param name=""delegate"">The delegate to register.</param>
+    /// <returns>Registration builder allowing the registration to be configured.</returns>
+    public static IRegistrationBuilder<TComponent, SimpleActivatorData, SingleRegistrationStyle>
+        Register<{methodGenericTypeList}> (
```
tillig, line 130:
```suggestion
        Register<{methodGenericTypeList}>(
```

codegen/Autofac.CodeGen/DelegateRegisterGenerator.cs
```diff
@@ -0,0 +1,329 @@
...
+
+        private static void GenerateDelegateInvokerWithComponentContext(StringBuilder sb, string typeName, int argCount)
+        {
+            var reusableStrBuilder = new StringBuilder();
+            var delegateGenericTypeList = GetTypeParamList(reusableStrBuilder, withComponentContext: true, argCount);
+
+            sb.AppendLine($@"
+    public sealed class {typeName}<{GetTypeParamList(reusableStrBuilder, withComponentContext: false, argCount)}> : BaseGenericResolveDelegateInvoker");
+            WriteTypeConstraints(sb, argCount, 2);
+            sb.Append($@"    {{
```
tillig, line 196:
```suggestion
            sb.Append($@"    {{
```

test/Autofac.Test.CodeGen/Helpers/ModuleInitializer.cs
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    private static ConversionResult ConvertDriver(GeneratorDriver target, IReadOnlyDictionary<string, object> context)
+    {
+        return ConvertRunResult(target.GetRunResult(), context);
+    }
+
+    private static Target SourceToTarget(GeneratedSourceResult source)
+    {
+        var data = $@"//HintName: {source.HintName}
```
tillig, line 64:
```suggestion
        var data = $@"// HintName: {source.HintName}
```
Not sure if we have control over this, but the analyzers would like it better if there was a space.

tillig (changes requested):
I'm definitely learning a lot about source generators. Pros and cons, dev experience, that sort of thing. No real major findings here, mostly just little one-liners.

codegen/Autofac.CodeGen/AnalyzerReleases.Shipped.md
```diff
@@ -0,0 +1,9 @@
+﻿; Shipped analyzer releases
```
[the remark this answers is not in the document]
alistairjevans: I think it's required...will check. To be honest this is my first source generator with no analyser messages, I lifted this from a common template I've used before.

src/Autofac/ResolveRequest.cs
```diff
@@ -11,6 +11,11 @@ namespace Autofac;
 /// </summary>
 public class ResolveRequest
 {
+    /// <summary>
+    /// Shared constant value defining an empty set of parameters.
+    /// </summary>
+    internal static readonly IEnumerable<Parameter> NoParameters = Enumerable.Empty<Parameter>();
```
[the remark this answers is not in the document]
alistairjevans: I think as soon as it became public/internal, I didn't really want to introduce direct coupling between the `BaseGenericDelegateInvoker` and the `RegistrationExtensions` class. It felt like it introduced a dependency loop between the two classes in a way I didn't love.

I was close to creating a new `Parameters` internal class just to hold the constant but thought that may be overkill.

So, sort of because it was "kinda weird", I guess.

tillig:
It definitely looks like an OmniSharp thing where navigating to generated code won't treat that generated code like it's part of a project. It treats it more like... if you opened that file in a whole disconnected context so it doesn't tie symbols together. [I've dropped a comment on the open issue for that in OmniSharp](https://github.com/OmniSharp/omnisharp-vscode/issues/4215#issuecomment-1087896434).

I _am_ able to set a breakpoint and step into the generated code, and once I'm there I can set breakpoints inside that generated code just like I can with SourceLink. It's just getting started and having to step in there that's a challenge, but it's not insurmountable, just not _obvious_.

Setting `<EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>` does get the generated source to start showing up in `obj/Debug/netXXX/generated/...` which is nice, but it doesn't appear to fix the "F12 navigate to definition" sort of thing in VS Code. It doesn't matter whether that property is true, VS Code _debugging_ into generated code (like from a unit test) works regardless. It's just a navigation problem, which is something I can live with.

codegen/Autofac.CodeGen/DelegateRegisterGenerator.cs
```diff
@@ -0,0 +1,329 @@
...
+    {
+        public const int NumberOfGenericArgs = 10;
+
+        public const int SpacesPerIndent = 4;
+
+        public void Initialize(IncrementalGeneratorInitializationContext context)
+        {
+            // Set up an incremental generator that regenerates when the 'RegistrationExtensions' class changes.
+            // Capture the INamedTypeSymbol when it does.
+            IncrementalValuesProvider<INamedTypeSymbol> classDeclarations = context.SyntaxProvider
```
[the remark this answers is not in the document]
alistairjevans: I found that it was super easy to confuse `IncrementalValuesProvider` and `IncrementalValueProvider`, to the extent that it made it much easier to comprehend if you leave the type names.

test/Autofac.Test.CodeGen/Helpers/ModuleInitializer.cs
alistairjevans: I think a space may not be allowed, but will check. Why does the analyzer not understand the comment is a string...?

codegen/Autofac.CodeGen/DelegateRegisterGenerator.cs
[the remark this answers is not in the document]
tillig: Fair enough.

test/Autofac.Test.CodeGen/Autofac.Test.CodeGen.csproj
alistairjevans: Will add something to a README about this.

test/Autofac.Test.Compilation/Autofac.Test.Compilation.csproj
alistairjevans: I did indeed initially add them to this test project, then realised that the two projects were coming at it from completely different ends, as you say.

test/Autofac.Test.CodeGen/TestExtensionMethodCodeGen.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) Autofac Project. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Autofac.Test.CodeGen.Helpers;
+
+namespace Autofac.Test.CodeGen;
+
+[UsesVerify]
+public class TestExtensionMethodCodeGen
```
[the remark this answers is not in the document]
alistairjevans: No; will fix.

test/Autofac.Test.CodeGen/Helpers/ModuleInitializer.cs
tillig: Analyzers want a space between `//` and the comment text. Admittedly it's not that big of a deal since it seems mostly to support the `Verify` framework and it doesn't show up (at least in VS Code) unless you actually open that file since it's excluded from compilation. It was just nice to be able to go down the list of analyzers and go, "Yup, the generated code adheres here, doesn't here... OK, we know how to tweak it." If we can't change this, it's not going to break my heart. 💔

codegen/Autofac.CodeGen/DelegateRegisterGenerator.cs
```diff
@@ -0,0 +1,329 @@
...
+
+            sb.AppendLine($@"
+}}");
+
+            spc.AddSource($"Autofac.{holdingTypeName}.g.cs", sb.ToString());
+        }
+
+        private static void GenerateDelegateInvokerWithComponentContext(StringBuilder sb, string typeName, int argCount)
+        {
+            var reusableStrBuilder = new StringBuilder();
```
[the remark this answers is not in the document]
alistairjevans: It's used again a few lines further down on 194; name isn't great though, I admit, just wanted to have a StringBuilder I could re-use for misc purposes.

codegen/Autofac.CodeGen/AnalyzerReleases.Shipped.md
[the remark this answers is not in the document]
alistairjevans: Turns out not needed; will delete.

test/Autofac.Specification.Test/Registration/LambdaGenericOverloadRegistrationTests.cs
```diff
@@ -0,0 +1,588 @@
...
+            Arg1 = arg1;
+            Arg2 = arg2;
+        }
+
+        public string Arg1 { get; }
+
+        public string Arg2 { get; }
+    }
+
+    private static readonly Type[] _allDeps = new[]
```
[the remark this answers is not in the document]
alistairjevans: Will update the name; I'd suggest that if an editorconfig is down as a suggestion, but it would fail at code review, it should be a warning rather than a suggestion.

I read suggestions as "personal preference on not doing X, or if you know there's a good reason not to do X", rather than "don't do X".

I've updated the editorconfig to set `constant_fields_should_be_pascal_case, `static_fields_should_be_camel_case`, `static_readonly_fields_should_be_pascal_case` and `camel_case_for_private_internal_fields` to `warning`, and no additional warnings have been raised, so seems like we should do that.

src/Autofac/Core/BaseGenericResolveDelegateInvoker.cs
```diff
@@ -0,0 +1,68 @@
+﻿// Copyright (c) Autofac Project. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Reflection;
+using System.Runtime.CompilerServices;
+
+namespace Autofac.Core;
```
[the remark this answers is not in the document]
alistairjevans: Moved; also moved the generated DelegateInvoker implementations into the same namespace.

src/Autofac/Core/BaseGenericResolveDelegateInvoker.cs
```diff
@@ -0,0 +1,68 @@
...
+
+/// <summary>
+/// Provides a base implementation for the dynamically generated DelegateInvoker classes.
+/// </summary>
```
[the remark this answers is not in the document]
alistairjevans: Ugh, get a warning trying to use a cref to the generator; probably because the analyzer project is not a referenced output assembly of the project.  I'll pop it in a remarks block instead without the `<see />` tag.

src/Autofac/Core/BaseGenericResolveDelegateInvoker.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// </summary>
+    /// <param name="parameters">The parameters.</param>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    protected static bool AnyParameters(IEnumerable<Parameter> parameters)
+    {
+        // The by-far most common way you'll end up with no parameters
+        // is by invoking a Resolve() function call that doesn't accept parameters, so the readonly NoParameters shared value is
+        // used.
+        // A referenceequals comparison here handles that neatly, and is significantly faster in benchmarks than doing the Any()
+        // call in every case.
```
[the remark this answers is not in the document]
alistairjevans: Ooh, that is a good extension.

test/Autofac.Test.CodeGen/Helpers/ModuleInitializer.cs
alistairjevans: Looks like a space is fine; so you can rest easy.

src/Autofac/Core/BaseGenericResolveDelegateInvoker.cs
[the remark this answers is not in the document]
tillig: Yeah, that's why I put it in `<c></c>` ("code") tags rather than a `<see />`. I just wanted to illustrate in the docs that it wasn't something arbitrarily capitalized, but is actually a name.

Oh, and I messed up the `<see>` tag, it should be `<seealso>` - you can't have `<see>` outside a main block.

test/Autofac.Test.CodeGen/Helpers/CompilationVerifier.cs
```diff
@@ -0,0 +1,38 @@
...
+    public static Task Verify(string source)
+    {
+        var compilation = Create(source);
+
+        var generator = new DelegateRegisterGenerator();
+
+        GeneratorDriver driver = CSharpGeneratorDriver.Create(generator);
+
+        // Once through for the generated code diagnostics, where we update the compilation.
+        driver = driver.RunGeneratorsAndUpdateCompilation(compilation, out var outputCompilation, out var outputDiagnostics);
```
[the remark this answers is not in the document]
tillig: I guess I don't feel _suuuuper_ strongly, but there is an analyzer warning about the allocation and unused values. Maybe a `Debug.Assert` or something to "use them" that won't affect the execution but will serve to keep them around? (Or maybe a `Debug.Assert` on what you think the values should be, to help validate what you'd be checking anyway?)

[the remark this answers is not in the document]
alistairjevans: Yeah, it is a test after all, I'll assert on the expected outputs.

.editorconfig
```diff
@@ -12,6 +12,16 @@ insert_final_newline = true
 
 ; .NET Code - almost, but not exactly, the same suggestions as corefx
 ; https://github.com/dotnet/corefx/blob/master/.editorconfig
+dotnet_style_operator_placement_when_wrapping = beginning_of_line
```
tillig, line 15: Make sure the `dotnet` stuff falls under something other than `[*]`. I've used [the Roslyn repo `.editorconfig`](https://github.com/dotnet/roslyn/blob/main/.editorconfig) as a pattern before - not only for starting settings, but ordering. It can get really hard to make sure duplicate conflicting entries aren't added if there isn't some sort of ordering. I think `dotnet` directives in there are under `[*.{cs,vb}]`. In fact, I think if you diff ours vs the Roslyn repo one, it might line up pretty close.

We also can't include `end_of_line` here for the Mac/Linux folks. I think the only time I've seen that used successfully is when it's under `[*.sh]` because it matters, but then it also needs to get coordinated with `.gitattributes` to make sure Git auto line endings match up. Generally best to just not specify line endings if it doesn't matter and let the client deal with it.

A blank line above section headers like `[*.cs]` also helps more easily locate those section headers.

tillig:
Looks like the stuff left:

- README to kinda explain source generator stuff
- Reusable `StringBuilder`
- Updates to `.editorconfig`
- `Debug.Assert` on those `out var` items.

Re: .editorconfig, we don't have to tackle that whole thing right now. I thought you were just going to mark those naming items as `warning` and call it a day. You _can_ update everything if you want, for sure, but we don't have to get in the weeds there if you don't want to.

.editorconfig
alistairjevans: Ugh, I didn't add any of the extra changes manually; it was that new-fangled UI editor for editorconfig in VS2022 (which I really don't like); it opened it and must have asserted all this extra stuff. I'll just take it out, I only mean to change the warning level on the naming conventions.

tillig: You VS users and your fancy tooling. 😆

codegen/Autofac.CodeGen/DelegateRegisterGenerator.cs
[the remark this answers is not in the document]
alistairjevans: All good points, I was probably optimising too early; source generators _can_ run every time you type a character in the editor, so speed is important. The incremental generator stuff is pretty new, and will probably help reduce the impact (source generators used to just run on every roslyn compilation pass, so was more dramatic).

I guess we can always revisit if it becomes noticeable.

alistairjevans:
OK READMEs are up, think I've resolved everything outstanding now.

tillig (approved):
🦄

alistairjevans:
👍 I'll work on a PR for the docs for the new feature.

PR #1323, opened 2022-04-07
Change sync disposal warning message adjective from efficient -> inefficient

alistairjevans:
Noticed this typo while updating docs.


tillig (approved):
🦚

PR #1325, opened 2022-04-09
Single constructor reflection optimisation

alistairjevans:
We didn't actually need to generate extra delegates using the mechanism added in #1321, I've added fast-paths through the reflection activator for single-constructor and matched-sig-constructor selection that does "early binding" of sorts on the constructors where possible.

The approach taken actually means I don't need to change the binding exception handling at all, and the existing resolve check behaviour stays.

We get noticeable perf and memory usage improvements across the board. 

Some examples:

#### DeepGraphResolveBenchmark
```markdown
# develop
|  Method |     Mean |    Error |   StdDev |  Gen 0 | Allocated |
|-------- |---------:|---------:|---------:|-------:|----------:|
| Resolve | 13.97 μs | 0.061 μs | 0.054 μs | 4.1351 |     17 KB |

# optimised
|  Method |     Mean |    Error |   StdDev |   Median |  Gen 0 | Allocated |
|-------- |---------:|---------:|---------:|---------:|-------:|----------:|
| Resolve | 12.82 μs | 0.254 μs | 0.451 μs | 12.45 μs | 3.5553 |     15 KB |
```

#### ChildScopeResolveBenchmark
```markdown
# develop
|  Method |     Mean |    Error |   StdDev |  Gen 0 |  Gen 1 | Allocated |
|-------- |---------:|---------:|---------:|-------:|-------:|----------:|
| Resolve | 32.08 μs | 0.195 μs | 0.173 μs | 7.4463 | 1.4648 |     31 KB |

# optimised 
|  Method |     Mean |    Error |   StdDev |  Gen 0 |  Gen 1 | Allocated |
|-------- |---------:|---------:|---------:|-------:|-------:|----------:|
| Resolve | 31.47 μs | 0.084 μs | 0.070 μs | 7.3853 | 1.4648 |     30 KB |
```

#### EnumerableResolveBenchmark
```markdown
# develop
|               Method |     Mean |     Error |    StdDev |  Gen 0 | Allocated |
|--------------------- |---------:|----------:|----------:|-------:|----------:|
|   ResolveIEnumerable | 3.066 μs | 0.0223 μs | 0.0186 μs | 0.9727 |      4 KB |
| ResolveIReadOnlyList | 3.070 μs | 0.0072 μs | 0.0064 μs | 0.9804 |      4 KB |
|         ResolveArray | 3.114 μs | 0.0106 μs | 0.0094 μs | 0.9727 |      4 KB |

# optimised
|               Method |     Mean |     Error |    StdDev |   Median |  Gen 0 | Allocated |
|--------------------- |---------:|----------:|----------:|---------:|-------:|----------:|
|   ResolveIEnumerable | 2.747 μs | 0.0122 μs | 0.0114 μs | 2.749 μs | 0.8125 |      3 KB |
| ResolveIReadOnlyList | 2.795 μs | 0.0189 μs | 0.0167 μs | 2.791 μs | 0.8202 |      3 KB |
|         ResolveArray | 2.787 μs | 0.0557 μs | 0.1175 μs | 2.704 μs | 0.8125 |      3 KB |
```

#### Multi-Constructor Benchmarks
```markdown
# develop
|             Method |       Mean |   Error |  StdDev | Ratio |  Gen 0 | Allocated |
|------------------- |-----------:|--------:|--------:|------:|-------:|----------:|
|             Single |   813.9 ns | 2.08 ns | 1.84 ns |  1.00 | 0.3557 |      1 KB |
|                Two | 1,345.0 ns | 3.56 ns | 3.33 ns |  1.65 | 0.5455 |      2 KB |
| TwoValidOneInvalid | 1,684.5 ns | 6.15 ns | 5.75 ns |  2.07 | 0.6866 |      3 KB |
|              Three | 1,944.2 ns | 6.86 ns | 6.08 ns |  2.39 | 0.7782 |      3 KB |
|               Four | 2,568.4 ns | 8.19 ns | 7.66 ns |  3.16 | 1.0529 |      4 KB |

# optimised
|             Method |       Mean |    Error |  StdDev | Ratio | RatioSD |  Gen 0 | Allocated |
|------------------- |-----------:|---------:|--------:|------:|--------:|-------:|----------:|
|             Single |   735.8 ns |  2.58 ns | 2.16 ns |  1.00 |    0.00 | 0.3204 |      1 KB |
|                Two | 1,220.8 ns |  8.80 ns | 6.87 ns |  1.66 |    0.01 | 0.4921 |      2 KB |
| TwoValidOneInvalid | 1,555.0 ns |  4.23 ns | 3.96 ns |  2.11 |    0.01 | 0.6313 |      3 KB |
|              Three | 1,769.7 ns |  6.51 ns | 6.09 ns |  2.40 |    0.01 | 0.7000 |      3 KB |
|               Four | 2,359.6 ns | 10.00 ns | 8.86 ns |  3.21 |    0.02 | 0.9460 |      4 KB |
```

Improvements on the multi-constructor benchmarks (which use the old path) may be down to the use of the `ReferenceEquals` check against the `NoParameters` constant, rather than `Any()`.

### DIFF
--- a/src/Autofac/Core/Activators/Reflection/ConstructorBinder.cs
+++ b/src/Autofac/Core/Activators/Reflection/ConstructorBinder.cs
@@ -36,7 +36,7 @@ public class ConstructorBinder
         if (_illegalParameter is null)
         {
             // Build the invoker.
-            _factory = FactoryCache.GetOrAdd(constructorInfo, FactoryBuilder);
+            _factory = FactoryCache.GetOrAdd(Constructor, FactoryBuilder);
         }
     }
 
@@ -111,6 +111,11 @@ public class ConstructorBinder
         }
 
         return BoundConstructor.ForBindSuccess(this, _factory!, valueRetrievers);
+    }
+
+    internal Func<object?[], object> GetConstructorInvoker()
+    {
+        return FactoryCache.GetOrAdd(Constructor, FactoryBuilder);
     }
 
     private static Func<object?[], object> GetConstructorInvoker(ConstructorInfo constructorInfo)
--- a/src/Autofac/Core/Activators/Reflection/ConstructorBinder.cs
+++ b/src/Autofac/Core/Activators/Reflection/ConstructorBinder.cs
@@ -116,9 +116,12 @@ public class ConstructorBinder
     /// <summary>
     /// Get the constructor factory delegate.
     /// </summary>
-    internal Func<object?[], object> GetConstructorInvoker()
+    /// <remarks>Will return null if the constructor contains an invalid parameter.</remarks>
+    internal Func<object?[], object>? GetConstructorInvoker()
     {
-        return FactoryCache.GetOrAdd(Constructor, FactoryBuilder);
+        // Method is only used in zero-parameter constructors, so illegal parameters cannot happen.
+        // Adding the null check + throw just in case.
+        return _factory;
     }
 
     private static Func<object?[], object> GetConstructorInvoker(ConstructorInfo constructorInfo)
--- a/src/Autofac/Core/Activators/Reflection/ConstructorBinder.cs
+++ b/src/Autofac/Core/Activators/Reflection/ConstructorBinder.cs
@@ -113,6 +113,9 @@ public class ConstructorBinder
         return BoundConstructor.ForBindSuccess(this, _factory!, valueRetrievers);
     }
 
+    /// <summary>
+    /// Get the constructor factory delegate.
+    /// </summary>
     internal Func<object?[], object> GetConstructorInvoker()
     {
         return FactoryCache.GetOrAdd(Constructor, FactoryBuilder);
--- a/src/Autofac/Core/Activators/Reflection/ConstructorBinder.cs
+++ b/src/Autofac/Core/Activators/Reflection/ConstructorBinder.cs
@@ -119,8 +119,6 @@ public class ConstructorBinder
     /// <remarks>Will return null if the constructor contains an invalid parameter.</remarks>
     internal Func<object?[], object>? GetConstructorInvoker()
     {
-        // Method is only used in zero-parameter constructors, so illegal parameters cannot happen.
-        // Adding the null check + throw just in case.
         return _factory;
     }
 
--- a/src/Autofac/Core/Activators/Reflection/IConstructorSelectorWithEarlyBinding.cs
+++ b/src/Autofac/Core/Activators/Reflection/IConstructorSelectorWithEarlyBinding.cs
@@ -1,0 +1,28 @@
+﻿// Copyright (c) Autofac Project. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Autofac.Core.Activators.Reflection;
+
+/// <summary>
+/// Defines an interface that indicates a constructor selector can attempt to
+/// determine the correct constructor early, as the container is built,
+/// without needing to understand the set of parameters in each resolve.
+/// </summary>
+public interface IConstructorSelectorWithEarlyBinding : IConstructorSelector
+{
+    /// <summary>
+    /// Given the set of all found constructors for a registration, try and
+    /// select the correct single <see cref="ConstructorBinder"/> to use.
+    /// </summary>
+    /// <param name="constructorBinders">
+    /// The set of binders for all constructors found by
+    /// <see cref="IConstructorFinder"/> on the registration.
+    /// </param>
+    /// <returns>
+    /// The single, correct binder to use. If the method returns null, this
+    /// indicates the selector requires resolve-time parameters, and the default
+    /// <see cref="IConstructorSelector.SelectConstructorBinding"/> method will
+    /// be invoked.
+    /// </returns>
+    ConstructorBinder? SelectConstructorBinder(ConstructorBinder[] constructorBinders);
+}
--- a/src/Autofac/Core/Activators/Reflection/ReflectionActivator.cs
+++ b/src/Autofac/Core/Activators/Reflection/ReflectionActivator.cs
@@ -98,13 +98,16 @@ public class ReflectionActivator : InstanceActivator, IInstanceActivator
             UseSingleConstructorActivation(pipelineBuilder, binders[0]);
             return;
         }
-        else if (ConstructorSelector is MatchingSignatureConstructorSelector matchingSelector)
+        else if (ConstructorSelector is IConstructorSelectorWithEarlyBinding earlyBindingSelector)
         {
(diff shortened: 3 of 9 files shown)


src/Autofac/Core/Activators/Reflection/ConstructorBinder.cs
```diff
@@ -36,7 +36,7 @@ public ConstructorBinder(ConstructorInfo constructorInfo)
         if (_illegalParameter is null)
         {
             // Build the invoker.
-            _factory = FactoryCache.GetOrAdd(constructorInfo, FactoryBuilder);
+            _factory = FactoryCache.GetOrAdd(Constructor, FactoryBuilder);
```
tillig, line 39: Should this use the `GetConstructorInvoker` new method?

```suggestion
            _factory = GetConstructorInvoker();
```

src/Autofac/Core/Activators/Reflection/ReflectionActivator.cs
```diff
@@ -91,6 +93,20 @@ public void ConfigurePipeline(IComponentRegistryServices componentRegistryServic
             binders[idx] = new ConstructorBinder(availableConstructors[idx]);
         }
 
+        if (binders.Length == 1)
+        {
+            UseSingleConstructorActivation(pipelineBuilder, binders[0]);
+            return;
+        }
+        else if (ConstructorSelector is MatchingSignatureConstructorSelector matchingSelector)
```
tillig, line 101: It feels kind of... weird? bad?... to me that we have to "know" about a specific `IConstructorSelector` type here. Is this something where we should technically be looking at updating the `IConstructorSelector` interface?

src/Autofac/Core/Activators/Reflection/ReflectionActivator.cs
```diff
@@ -147,7 +213,8 @@ private object ActivateInstance(IComponentContext context, IEnumerable<Parameter
     private BoundConstructor[] GetAllBindings(ConstructorBinder[] availableConstructors, IComponentContext context, IEnumerable<Parameter> parameters)
     {
         // Most often, there will be no `parameters` and/or no `_defaultParameters`; in both of those cases we can avoid allocating.
-        var prioritisedParameters = parameters.Any() ? EnumerateParameters(parameters) : _defaultParameters;
+        // Do a reference compare with the NoParameters constant; faster than an Any() check for the common case.
```
tillig, line 216: I think this comment goes down in the `GetAllParameters` method since the associated code moved.

tillig (changes requested):
Couple minor things, question about whether we should change the `IConstructorSelector` interface.

src/Autofac/Core/Activators/Reflection/ReflectionActivator.cs
tillig: If we don't want to change it, we could do that additive thing like `IConstructorSelectorSupportsSomething` and check to see if the constructor selector supports the optional feature.

alistairjevans:
Added `IConstructorSelectorWithEarlyBinding`. Also made it optional for the `SelectConstructorBinder` function in that interface to return a `ConstructorBinder` instance. If it returns null, it falls back to the old selection mechanism at resolve time; means implementations can make "best effort" selection attempts.

src/Autofac/Core/Activators/Reflection/IConstructorSelectorWithEarlyBinding.cs
```diff
@@ -0,0 +1,28 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Autofac.Core.Activators.Reflection;
+
+/// <summary>
+/// Defines an interface that indicates a constructor selector can attempt to
+/// determine the correct constructor early, as the container is built,
+/// without needing to understand the set of parameters in each resolve.
+/// </summary>
+public interface IConstructorSelectorWithEarlyBinding : IConstructorSelector
```
tillig, line 11:
```suggestion
public interface IConstructorSelectorWithEarlyBinding
```

It seems like the way .NET Core did it, the "extension feature interface" [didn't inherit the interface of the thing it was extending](https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.DependencyInjection.Abstractions/src/ISupportRequiredService.cs). I don't know which way is right, I can see pros and cons to each.

Admittedly, you can't really _just_ implement `IConstructorSelectorWithEarlyBinding`, but then, you also can't _just_ implement `ISupportRequiredService`, either. They only make sense together.

tillig (changes requested):
Looking good. Only real question of substance is about whether `IConstructorSelectorWithEarlyBinding` should inherit `IConstructorSelector`.

src/Autofac/Core/Activators/Reflection/ConstructorBinder.cs
```diff
@@ -113,6 +113,17 @@ public BoundConstructor Bind(IEnumerable<Parameter> availableParameters, ICompon
...
     }
 
+    /// <summary>
+    /// Get the constructor factory delegate.
+    /// </summary>
+    /// <remarks>Will return null if the constructor contains an invalid parameter.</remarks>
+    internal Func<object?[], object>? GetConstructorInvoker()
+    {
+        // Method is only used in zero-parameter constructors, so illegal parameters cannot happen.
+        // Adding the null check + throw just in case.
```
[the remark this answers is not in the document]
alistairjevans: 🤦‍♂️

src/Autofac/Core/Activators/Reflection/IConstructorSelectorWithEarlyBinding.cs
alistairjevans: My feeling here is that in order to provide a constructor selector to Autofac at registration time, it _must_ implement `IConstructorSelector`; that's the type of the property in `ReflectionActivatorData`.  It feels like it would be confusing if someone created an implementation of `IConstructorSelectorWithEarlyBinding`, and then just couldn't provide it to Autofac?

tillig: Agreed. I just didn't know if I was missing a use case somewhere, like some future composability system like the `PartCatalog` or `HttpContext.Features` sort of thing in ASP.NET Core. I'm fine with it either way, we can leave it.

tillig (approved):
🚨

jimmycartrette:
> Improvements on the multi-constructor benchmarks (which use the old path) may be down to the use of the `ReferenceEquals` check against the `NoParameters` constant, rather than `Any()`.

Would it be because the multi-contructors still have dependencies, and those dependencies have single constructors which are now optimized?

alistairjevans:
Yes, that's also quite possible; I didn't really investigate the reason that closely.

PR #1326, opened 2022-04-11
Reflection cache clear support

### LINKED ISSUE #1324: Memory leak when using types in collectible AssemblyLoadContexts
I've run into something that is either a memory leak or an unsupported use case, and I'm looking for guidance on which.

My application has a plugin model that loads assemblies into collectible AssemblyLoadContexts. It has a base lifetime scope built from core types loaded into the default ALC (including the Autofac types), and a nested lifetime scope built from types in assemblies in the collectible ALC. Requests are served from a third level of short-lived lifetime scopes based off the "current" plugin lifetime scope. Under certain conditions a new collectible ALC and plugin lifetime scope are constructed to replace the existing one, which is then disposed. Unfortunately, this leaks.

The source of the leaks is Autofac's reflection caches:

Autofac.Core.Activators.Reflection.ConstructorBinder.FactoryCache
Autofac.Core.Activators.Reflection.DefaultConstructorFinder.DefaultPublicConstructorsCache
Autofac.Util.TypeExtensions.IsGenericEnumerableInterfaceCache
Autofac.Util.TypeExtensions.IsGenericListOrCollectionInterfaceTypeCache
Autofac.Util.TypeExtensions.IsGenericTypeDefinedByCache

There may well be others, but using reflection to clear these caches fixes the leak in my test harness, and my ALCs successfully unload. To elaborate slightly, these are all static ConcurrentDictionaries that cache reflection information from the collectible ALCs, holding references and preventing unloading.

Is the intention that Autofac be usable with types in collectible ACLs? Is my hacky fix safe?

I'm happy to provide a test or test harness if this is considered a bug.

alistairjevans:
Raising a PR that fixes #1324.

@mburnell, I'm partly raising this so we can get a MyGet build out; you can pull that down and see if that helps with your situation before we fully push a fix out.

### DIFF
--- a/src/Autofac/Core/Activators/Reflection/AutowiringPropertyInjector.cs
+++ b/src/Autofac/Core/Activators/Reflection/AutowiringPropertyInjector.cs
@@ -17,9 +17,9 @@ internal static class AutowiringPropertyInjector
     /// </summary>
     internal const string InstanceTypeNamedParameter = "Autofac.AutowiringPropertyInjector.InstanceType";
 
-    private static readonly ConcurrentDictionary<PropertyInfo, Action<object, object?>> PropertySetters = new();
+    private static readonly ReflectionCacheDictionary<PropertyInfo, Action<object, object?>> PropertySetters = new();
 
-    private static readonly ConcurrentDictionary<Type, PropertyInfo[]> InjectableProperties = new();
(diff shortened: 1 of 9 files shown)


alistairjevans:
Closing in favour of a new approach on the same branch.

PR #1333, opened 2022-07-22
Note about AWS sponsorship.

tillig:
AWS made a financial contribution and we thought it would be nice to mention that kind support in the README.


alistairjevans (approved):
👍

PR #1334, opened 2022-07-24
JC Circular dependencies 

jimmycartrette:
Circular dependencies only begin in depth check after _maxResolveDepth has been reached.

The existing approach is (n*m) complexity where n is the number of resolves and m is the average resolve depth. Our use case can be very deep resolve chains so this impacts us more.

This approach could improve performance on the hot, non-error path, but decrease performance on an error case.
Could circular dependencies become transient? This could miss those.

I set up a much deeper (8 level) unit test to match closer to our normal resolve operations, here are the results from that:

Deep nest (8 levels) with existing Circular Dependency
|  Method |     Mean |   Error |  StdDev |   Gen 0 |  Gen 1 | Allocated |
|-------- |---------:|--------:|--------:|--------:|-------:|----------:|
| Resolve | 261.1 us | 2.25 us | 1.75 us | 34.1797 | 8.3008 |    281 KB |

Deep nest (8 levels) with new limited Circular Dependency
|  Method |     Mean |   Error |  StdDev |   Gen 0 |  Gen 1 | Allocated |
|-------- |---------:|--------:|--------:|--------:|-------:|----------:|
| Resolve | 217.9 us | 2.13 us | 1.89 us | 31.4941 | 7.8125 |    259 KB |

Alternatives could be an opt-in/opt-out for this, or we would be happy with a complete opt out ability for CircularDependencyMiddleware (we'd enable in Debug, leave off in prod because we've had 0 ever happen in prod but it costs us a lot of time)


alistairjevans:
I've been mulling this over the last couple of days, and I can't help but feel like detecting circular references _late_, i.e. only after we've been through the loop a few times may have some undesirable effects.

1. Dependencies can indeed be transient, with subsequent registrations selected (perhaps via `IIndex` or `Func<>`) inside a given constructor; so it would be possible to have a _limited_ circular dependency that manages to break out of the loop before it hits your condition check, but still does 20 unnecessary allocations. This then gets quite thorny, because the user would have no way of knowing they were doing unnecessary resolve work or memory allocations without inspecting the entire resolve graph.

2. While ideally constructors would not have any side-effects, the truth is they often do (opening DB connections, doing stuff on the file system, etc). If we allow a circular dependency loop to get too far before we stop it, it could have unknown side-effects for users that expect (or at least expected at one point) us to protect them from.

I'm actually more in favour of, in a tested codebase, allowing advanced users to disable the expensive "search all prior registrations" behaviour that is the algorithmically slow part of the process.

We can't just remove the middleware entirely though, since the `requestStack` it populates is used by other parts of the resolve process to determine when a resolve flow starts and ends.

My suggestion (if you want to try it) would be to add a method to `CircularDependencyDetectorMiddleware`, something like `UnsafeDisableRegistrationChecks` (that `Unsafe` prefix seems to be a standard convention these days in the runtime), that turns off the search of the `requestStack`.  

This can then be accessed by code via `CircularDependencyDetectorMiddleware.Default.UnsafeDisableRegistrationChecks()`.

tillig:
Is there a way to reduce the cost of the lookup, like by using some sort of hashed set or other data structure? (No, I haven't thought that through, I'm just thinking if the expensive part is the check then O(1) rather than O(N) would let us have our cake and eat it, too.)

jimmycartrette:
Good idea, maybe reimplement the SegmentedStack backing with something that does O(1) lookup but still maintains order and ability to scope into segments. I assume a slight degradation in performance for very small resolution chains due to slightly more costly insert but still significant improvements in larger ones for hash lookup. I'll try a couple of implementations and see what the numbers come to.

jimmycartrette:
Here is the performance of this implementation. I've modified the Deep Graph resolve to be very specific to testing 16 resolves of depths 1..16 so the avg resolve depth should be 8 here and these don't compare to previous stats above:
|                                            Method |     Mean |     Error |    StdDev | Ratio | RatioSD |  Gen 0 |  Gen 1 | Allocated |
|-------------------------------------------------- |---------:|----------:|----------:|------:|--------:|-------:|-------:|----------:|
|                                     Resolve16Deep | 8.404 us | 0.0714 us | 0.0633 us |  1.00 |    0.00 | 1.0376 | 0.0153 |      9 KB |
| Resolve16DeepWithCircularDependencyChecksDisabled | 6.868 us | 0.1154 us | 0.1079 us |  0.82 |    0.02 | 0.9689 | 0.0153 |      8 KB |

Something like 18% faster for avg 8 depth resolves, with minimal loss of troubleshooting functionality. We'd use this in our production builds.

test/Autofac.Specification.Test/Features/CircularDependencyTests.cs
```diff
@@ -75,6 +76,23 @@ public void DetectsCircularDependencies()
...
+        builder.RegisterType<D>().As<ID>();
+        builder.RegisterType<A>().As<IA>();
+        builder.RegisterType<BC>().As<IB, IC>();
+        var target = builder.Build();
+        var de = Assert.Throws<DependencyResolutionException>(() => target.Resolve<ID>());
+
+        // The exception should contain a loop because we have not caught it
+        Assert.Contains("Autofac.Specification.Test.Features.CircularDependency.D -> Autofac.Specification.Test.Features.CircularDependency.A -> Autofac.Specification.Test.Features.CircularDependency.BC -> Autofac.Specification.Test.Features.CircularDependency.A", de.Message);
+        Assert.DoesNotContain("component dependency", de.Message);
+        DefaultMiddlewareConfiguration.EnableProactiveCircularDependencyChecks();
```
alistairjevans, line 93: Can you put this in a finally block so if the test fails the configuration is still reverted?

src/Autofac/Core/Resolving/Middleware/DefaultMiddlewareConfiguration.cs
```diff
@@ -0,0 +1,28 @@
...
+
+/// <summary>
+/// Allows configuration of default middleware.
+/// </summary>
+public static class DefaultMiddlewareConfiguration
+{
+    /// <summary>
+    /// Universally disables the circular dependency check on every resolution. Instead,
+    /// a circular dependency may be caught after many loops and may cause undesirable
+    /// side effects, however resolution performance should improve with deep resolution chains.
```
alistairjevans, line 14: Can we add a comment here (perhaps in a `<remarks>` block) to the effect that this should only be done after extensive testing of the application to ensure no circular dependencies exist (or words to that effect).

bench/Autofac.Benchmarks/DeepGraphResolveBenchmark.cs
```diff
@@ -1,59 +1,147 @@
...
+        public A(B b) { }
+    }
+
+    internal class B
+    {
+        public B(C c) { }
     }
 
-    internal class B1
+
```
alistairjevans, line 77: Remove extra blank line.

bench/Autofac.Benchmarks/DeepGraphResolveBenchmark.cs
```diff
@@ -1,59 +1,147 @@
...
+    internal class C
     {
-        public B1(B2 b2, C1 c1, C2 c2) { }
+        public C(D d) { }
+    }
+
+
```
alistairjevans, line 83: Remove extra blank line.

bench/Autofac.Benchmarks/DeepGraphResolveBenchmark.cs
```diff
@@ -1,59 +1,147 @@
...
+    internal class D {
```
alistairjevans, line 84: Move the brace onto the next line for consistency.

alistairjevans (changes requested):
A few minor things, but generally looks good.

alistairjevans:
@tillig, regarding your suggestions around data structures, I remember doing some extensive experiments around the best ways to implement the circular check over a segmented set of data structures at the time I put it in during "the great pipeline refactor". I came to the conclusion that fundamentally we need to iterate over the set of all registrations resolved so far in the hierarchy in order to effectively do pro-active circular dependency checks in the way users were used to, that stays efficient at small hierarchy depths.  Manipulating a `O(1)` set as the resolve takes place is much slower than adding to a stack, especially if that stack has already been "warmed up" by allocating the default initial size of 16 (`SegmentedStack.InitialCapacity`).

The performance cost at 'typical' hierarchy depths isn't extreme, but obviously does get worse at deep hierarchy depths as more and more iterations are required to evaluate each registration.

I'm inclined to take this change as a workable solution for advanced perf cases with deep hierarchies; while it is a static method (`DefaultMiddlewareConfiguration.UnsafeDisableProactiveCircularDependencyChecks()`), so not container specific, I took a look at alternatives, and making the option container-specific would require a whole bunch of annoying changes, so this option is lower maintenance cost for us too. I'm not sure I want to add the surface area and maintenance overheads needed to expose such a setting directly out of the container (or even make it directly intellisense-accessible from the container). 

Let me know if you feel differently, appreciate your thoughts.

tillig:
@alistairjevans Yeah, I remember quite a bit of work going into figuring out how to track things in The Great Pipeline Refactor of '20. I admit some of the details are lost to time in my currently scattered mind, but I do remember that.

I think the unsafe disable option is a reasonable solution for this admittedly pretty advanced case.

I think it'd be worthwhile to have answers for things like...

- If I can enable/disable circular dependencies at the registration level, how come I can't opt in/out of the circular dependency checking at that level?
- Similar question: why can't I opt in/out at a lifetime scope level?
- How does the [per-registration `PropertyWiringOptions.AllowCircularDependencies`](https://autofac.readthedocs.io/en/latest/advanced/circular-dependencies.html) interact with this? Does setting the unsafe detection check mean that every registration is effectively opted into allow circular dependencies?

I mention it because I've seen questions about circular dependencies and feature opt in/out for other features that are similar to the above questions.

Somewhat related to that enable/disable at the registration level... while I get "throw the giant global switch" is the easiest answer, would it be better to be able to tie into per-registration semantics and the `AllowCircularDependencies` sort of behavior? That is, if it's metadata on the registration, it could be propagated down through the resolve pipeline and the thing that gets check to see if it's enabled/disabled is a function of the registration, not a global switch.

Implementing the global switch could then be a middleware / module sort of thing that just updates the resolve pipeline with that metadata.

I don't want to overcomplicate it if "global switch" is better, but I thought the per-registration idea (especially with respect to how it interacts with other per-registration things that exist today) might be worth considering.

alistairjevans:
Hmmmm, good point about `AllowCircularDependencies`, but just to follow the thread, as its owner type `PropertyWiringOptions` indicates, it does only apply to properties, and is mostly sugar used only in the registration extension methods. It also never stores that value against the registration.

`AllowCircularDependencies` doesn't actually disable circular dependency checks for the registration, it just defers the property injection until the end of the resolve request rather than inline it.  If that registration has constructor circular dependencies, they will still hit.  In a sense it's actually safer for property injection, because it's not just that it 'allows' them, it executes the resolve outside the executing resolve context, so no (slim?) chance of stack overflow.

--- 

If we did track a flag like this more generally (with a new `DisableCircularDependencyChecks` value to the `RegistrationOptions` enum, for example) we would need to understand what setting the flag on a registration means for registrations deeper in the hierarchy. Are they implicitly also allowed to circular reference as well, or does it only apply just to the registration? 

I can see an ironically circular reference of Github issues when modules/complex graphs are involved:

> I've disabled circular reference detection, but I'm still seeing exceptions?

> I haven't turned off circular reference detection on this dependency, but the constructor is still executing multiple times?

Not saying it's impossible to figure out or anything, but the maintenance cost on us will likely be higher than the option on the table.

More philosophically, we haven't generally exposed _performance-related_ flags on registrations or the scopes (I believe), only ones that are related to tweaking _behaviour_. Personally I think we should continue to make generally opinionated perf-related decisions in favour of user safety, and conceal these potentially dangerous options away from the registration or lifetime scope intellisense, for the super-advanced use-cases.

tillig:
> Personally I think we should continue to make generally opinionated perf-related decisions in favour of user safety

I think I read somewhere that every configuration option you provide is a design decision you failed to make. That seems to be the case here. I mean, I get there's a perf aspect to it, but - devil's advocate - maybe _that's just OK and we don't actually need to allow folks to turn it off._ Again, just devil's advocate.

Hiding stuff from IntelliSense can be done by marking it `[EditorBrowsable(false)]`, if I'm not mistaken. I'm fine with hiding the option from the average user, just noting that if it was on a registration or scope, that could still be accomplished.

Off the top of my head, I don't know about any real performance related flags _at all_ that we've exposed, so there's not really a precedent for it that I'm aware of. It seems like "don't check for circular dependencies" is a sort of cross between behavior and perf.

I don't think we're going to escape from some level of question or another if we introduce this. I think any change here may benefit a very small number of advanced folks but will be a long-term maintenance burden for us, simply because we'll have to answer questions about it _and_ make sure we're not breaking it when we add updated/improved circular dependency handling on the next major refactor. I mean, it's almost like "you must have the environment variable `TOTALLY_UNDOCUMENTED_AND_MAY_CHANGE_OUT_FROM_UNDER_YOU` set to `SomeCrazyMagicValue`" might be a better option. 😆 

I recognize that internally the `AllowCircularDependencies` doesn't work the same as this, but from an API standpoint they're in the same general vicinity. The average consumer isn't going to differentiate between the property setting and this new unsafe flag thing. I guarantee that in six months I, myself, will forget the differences between the two things and have to re-research the whole thing when answering a question about it. That's why I figured I'd raise the potential cognitive load / mental grouping about the two features. They _sound the same_ (for slightly different purposes) so naturally someone is going to wonder if flipping one switch is going to affect the other switch, and why-didn't-you-do-it-the-way-I-personally-expected-it is going to be the outcome.

Honestly I don't feel _super strongly_ about it, just thinking out the precedent, the fact that we'll have this global flag (or registration flag, or scope flag, or whatever) to support, and what that might mean long-term for newbies shooting themselves in the foot. As long as the expected behavior is very clearly documented - including in all the places we talk about allowing (or not) circular dependency support - I'm fine.

It does make me wonder if that means we should have a specific `FeatureFlags` or `PerformanceFlags` class or something for this, such that it's not necessarily tied to middleware configuration. That is, if we need to expand what "unsafe disable circular dependency" means internally, like skipping doing something in the property registration thing for `AllowCircularDependencies` or something, then maybe having a more agnostic name would be of value.

Man, really not looking forward to the "I did this, but I also set the advanced flag, so why isn't it reading my mind?" questions. Something is sticking with me about how this will interact with that `AllowCircularDependencies` flag that I can't quantify, which could just be FUD around the whole thing. I'm admittedly already at my personal limit with the amount of support burden I'm willing to take on and I'm burning out fast. [Glances at the list of PRs that still need review/approval, releases that need to be cut post-PR, etc.]

OK, last thing: If this was something we tracked at the container / registration / scope level, that might open up additional things like:

- Ability to include that information in traces/diagnostics.
- Additional flags for perf/behavior at the container/registration/scope/whatever level rather than global all-or-nothing.

Which may have their own side-effects, like additional memory allocations for tracking that stuff through the pipeline or different-but-equally-complex end user support and questions.

Yeah, that's a lot of waffling and TL;DR. Sorry. Day job is getting in the way and my brain is pretty scattered.

alistairjevans:
I'm not sure this is something we need to rush; adding flags of this sort is something or a line in the sand for us, so we definitely don't want to cross it without thought.

I don't want to add to your load right now, so I'll take this one away and give it a little more thought, see if there's a more general approach for "options" that flows down through the container.

tillig:
I chatted some with @alistairjevans recently about how this has been kind of sitting for like a year now with no action. I agree that there's potentially some good performance benefits here, but I'm still pretty iffy on adding a flag for it - from the 80/20 standpoint, this definitely falls into the 20, and it means we have to support, test, and generally own this forever if it comes in. I still have some reservations on that. It'd be better if things "just worked," like...

- Are there general perf improvements that would make such a flag not required?
- Could we do something different that might equate to the same thing, like "you can change the stack depth for circular dependency detection" - if you know you'll only have a dependency depth of 4, you can set it to 5; or if you want to "disable" you could set it to -1. (I'm not recommending this, just brainstorming.)
- Could we show how to find/replace the CircularDependencyMiddleware with a custom version where you could add your own "skip everything" flag?

I also wonder if there are more sweeping things that could/should be done to normalize how circular dependencies are handled so we can better optimize that path. For example, if `AllowCircularDependencies` just changes the order of property injection, maybe that should be the default and then we can optimize _the general resolution path_. Again, I don't know if that's even a possibility at the moment, I'm just brainstorming.

@alistairjevans do you have anything to add here?

tillig:
This issue has been open for quite some time without any traction. While I think there's value in the PR, I'm not sure we have the bandwidth to support something outside that 80/20 area. Perhaps there's a different/better solution to this that wouldn't involve global flags and opt in/out - something that would support people well by default, such that it "just works." I'm worried a lot about people opting out and then shooting themselves in the foot, which results in support problems.

PR #1339, opened 2022-08-15
Improve scan perf

RogerKratz:
When, like us, have multiple RegisterAssemblyTypes (and large assemblies) and creating loads of container when running our test suites, the "general type checks" IsClass, IsAbstract and IsDelegate that is called over and over again on the very same type takes noticeable time.

I used old test AssemblyScanningPerformanceTests.MeasurePerformance as a benchmark. It now runs "too fast" so fyi increased the loop with *10 as part of this PR.

Note! Everything should work as before except...
In `ScanningRegistrationExtensions.RegisterTypes(RegisterTypes, params Type[])`, it's different behavior now. the "general type checks" mentioned above is no longer called, as that method calls `ScanTypes` directly. Not sure if these "general type checks" needs to be added explicitly here? I didn't change it in this PR as all tests are green, better listen with you first.

### DIFF
--- a/src/Autofac/Util/AssemblyExtensions.cs
+++ b/src/Autofac/Util/AssemblyExtensions.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Autofac Project. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+using System.Collections.Concurrent;
 using System.Reflection;
 
 namespace Autofac.Util;
@@ -21,7 +22,7 @@ public static class AssemblyExtensions
     /// <exception cref="System.ArgumentNullException">
     /// Thrown if <paramref name="assembly" /> is <see langword="null" />.
     /// </exception>
-    public static IEnumerable<Type> GetLoadableTypes(this Assembly assembly)
+    private static IEnumerable<Type> GetLoadableTypes(this Assembly assembly)
     {
         // Algorithm from StackOverflow answer here:
         // https://stackoverflow.com/questions/7889228/how-to-prevent-reflectiontypeloadexception-when-calling-assembly-gettypes
@@ -39,4 +40,16 @@ public static class AssemblyExtensions
             return ex.Types.Where(t => t is not null)!;
         }
     }
+
+    private static readonly ConcurrentDictionary<Assembly, IEnumerable<Type>> PossibleAssemblyTypes = new();
+
+    /// <summary>
+    /// A first filter for possible assembly types to register.
+    /// </summary>
+    /// <param name="assembly">The <see cref="System.Reflection.Assembly"/> from which to load types.</param>
+    public static IEnumerable<Type> PossibleScanTypes(this Assembly assembly) =>
+        PossibleAssemblyTypes.GetOrAdd(assembly, ass =>
+            ass.GetLoadableTypes().Where(t => t.IsClass &&
+                                              !t.IsAbstract &&
+                                              !t.IsDelegate()).ToArray());
 }
--- a/src/Autofac/Features/Scanning/OpenGenericScanningRegistrationExtensions.cs
+++ b/src/Autofac/Features/Scanning/OpenGenericScanningRegistrationExtensions.cs
@@ -92,7 +92,7 @@ internal static partial class ScanningRegistrationExtensions
             rb.RegistrationData.Services.OfType<IServiceWithType>().All(swt =>
                 t.IsOpenGenericTypeOf(swt.ServiceType)));
 
-        var types = assemblies.SelectMany(a => a.GetLoadableTypes())
+        var types = assemblies.SelectMany(a => a.PossibleScanTypes())
             .Where(t => t.IsGenericTypeDefinition)
             .CanBeRegistered(rb.ActivatorData);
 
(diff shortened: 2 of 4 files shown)


src/Autofac/Util/AssemblyExtensions.cs
```diff
@@ -39,4 +40,16 @@ public static IEnumerable<Type> GetLoadableTypes(this Assembly assembly)
             return ex.Types.Where(t => t is not null)!;
         }
     }
+
+    private static readonly ConcurrentDictionary<Assembly, IEnumerable<Type>> PossibleAssemblyTypes = new();
```
tillig, line 44: I think this is the only part of the solution that I'm a tiny bit concerned with - a static dictionary that will be used during registration but never after, but also won't ever get flushed. Do we know the memory impact here? (Also probably should move the definition to the top of the class, but that's more cosmetic than functional.) @alistairjevans am I overthinking it?

alistairjevans: This also relates to issue #1324, an assembly referenced in this way can never be unloaded.  We haven't really agreed a solution over there, but perhaps this is a good opportunity to revisit, since I'm not sure we want to make the problem "worse".

As for memory usage, this will be a list as long as the number of valid types in the assembly, which is an unknown size. I'm pretty sure a bunch of use-cases involve assemblies with a **lot** of types in them.

Personally I'd imagine defining an `IReflectionCache` interface that lets existing code 'register' cached reflection datastores in a single location, and put an instance of that interface on the `IComponentRegistryBuilder` interface. This means the scanning registration extensions can reference that `IReflectionCache` instance.  That reflection cache can then be cleared on-demand, and even passed through into the built container if needed.

I guess this is sort of an extension of the 'static' approach I started in #1326.  Maybe it's actually time to make the jump and create an instance to hold a reflection cache.

tillig: Man, I'm glad I asked, I totally forgot about that other issue about assembly unloading. Excellent point.

Agreed, some sort of instance for this that can be disposed/cleared later would maybe be better and allow the assembly unloading to work.

RogerKratz: >  That reflection cache can then be cleared on-demand, and even passed through into the built container if needed.

If letting this state go our of scope/cleared (which of course would be nicer), having a possibility to reuse the state between containers as you write above would be important. When creating one container, this PR's perf gain doesn't matter too much to us, it comes into play in our large test suites when _many_ containers are created. So keeping this state/cache between containers (tests in our case) is what would help us.

Thanks.

alistairjevans: Good point @RogerKratz, it suggests we would need a constructor overload on `ContainerBuilder` that allows you to specify an existing `IReflectionCache`, and possibly expose the `IReflectionCache` from the `Container`?

RogerKratz: Why expose it?

Also, maybe it would be needed also if used for a single container (?), but to allow it to work in scenarios like ours when containers are used in parallel, in needs to be impl thread safe.

tillig: Seems like there are other options if the issue is having a bunch of common things that get registered for every test but each test then has overrides - like using a common base container with the reflection registered stuff (do it once) and then nested lifetime scopes for the overrides (one per test).

I'm a little torn on `ContainerBuilder` constructor params. I get it, but I also wonder about precedent and future need to add things. Maybe a `ContainerBuilderContext` that has a property in it, so we could add to the `ContainerBuilderContext` later without changing/adding constructors on `ContainerBuilder`. Default zero-param would create the default context with whatever default cache in it.

I also know we have the `Properties` dictionary that gets created; maybe a well-known property is a place to set that.

```c#
public static void SetReflectionCache(this ContainerBuilder builder, IReflectionCache cache)
{
  builder.Properties["SomeWellKnownKey"] = cache;
}

public static IReflectionCache GetReflectionCache(this ContainerBuilder builder)
{
  return builder.Properties["SomeWellKnownKey"] as IReflectionCache;
}

var b = new ContainerBuilder();
b.SetReflectionCache(myCache);
b.RegisterAssemblyTypes(...);
```

RogerKratz: > Seems like there are other options if the issue is having a bunch of common things that get registered for every test but each test then has overrides - like using a common base container with the reflection registered stuff (do it once) and then nested lifetime scopes for the overrides (one per test).

I've been thinking doing it like this in our tests before. In my head this could make nasty bugs hard to find though (eg if developer has incorrectly put some mutable state on services/components etc). Sure, a fix like this PR is also in a sense "leaking state" between tests but.... ;)

`new ContainerBuilder().SetReflectionCache(myCache)`
sounds fine to me.

alistairjevans: Yep, that all makes sense.`SetReflectionCache` and `GetReflectionCache` make sense to me.  What would the default instance be?

There are caches in the `TypeExtensions` class, currently static, that would also likely fall into "things we would put in an `IReflectionCache`", to address #1324 at the same time.

However, if we do move those into an `IReflectionCache`, and default to having a new instance per  `ContainerBuilder`, people's existing unit test performance would plummet as each container has a new set of caches.  We don't want people really to have to add `SetReflectionCache` everywhere to get the old performance back.

It almost feels like `IReflectionCache` should have a shared static `Default` instance to mimic current behaviour, and if someone wants to use a custom instance that can be reset on-demand, then they can do so as described, but then that leads to the same memory leak problems as before.

tillig: Maybe the `Get` checks to see if the cache doesn't exist and instantiates/sets a default value on the fly?

alistairjevans:
@RogerKratz, I've pulled your assembly scanning caching into #1341, which is nearing merge. No additional "share cache between these builders" settings needed, so this PR has probably not been superseded?

RogerKratz:
I'm out of office atm and cannot really look at code, but if I understand you correctly - just drop this PR. 
Great work with #1341!

PR #1340, opened 2022-08-15
Ignore revs in git blame.

tillig:
Large scale changes that modify trivial things (tab-to-space conversion, file-scoped namespaces, etc.) can be ignored in git blame in git 2.23+. I think this will help with, as noted, the file-scoped namespace conversion, to see the "real" history of lines.

It does require some git client config, which I noted in the file. Community convention seems to be `.git-blame-ignore-revs` as the filename but this isn't a built-in convention like `.gitignore` or `.gitattributes`.

Based on [this blog article](https://akrabat.com/ignoring-revisions-with-git-blame/) and [this blog article](https://michaelheap.com/git-ignore-rev/).

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,11 @@
+# Ignore revisions in git blame - set your git config to use the file by convention:
+# git config --global blame.ignoreRevsFile .git-blame-ignore-revs
+#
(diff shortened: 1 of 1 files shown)


alistairjevans (approved):
LGTM

PR #1347, opened 2022-09-12
Remove misleading comment

OwnageIsMagic:
This code may produces leaks: if `ValueTask` backed by `IValueTaskSource` `ValueTaskSource<TResult>.GetResult(token)` won't get called, so it can't free result associated with that token.

And compiler generated code already checks for `IsCompleted`: 
https://sharplab.io/#v2:D4AQDABCCMCsDcBYAUCkBmKAmCBhCA3ihCVJiABwQBqAhgDYCuApiAGwQCyAFHU6xwAuASkLFSEiAEsAZhG6CAdAEkAzrgD2AWwAO9ZoOYATAMqMAxuearVMxvXoBPYeMluQAdiTI3pEAE4IQW8JAF8UcNRkIA==

### DIFF
--- a/src/Autofac/Core/Activators/ProvidedInstance/ProvidedInstanceActivator.cs
+++ b/src/Autofac/Core/Activators/ProvidedInstance/ProvidedInstanceActivator.cs
@@ -94,14 +94,23 @@ public class ProvidedInstanceActivator : InstanceActivator, IInstanceActivator
     }
 
     /// <inheritdoc />
-    protected override async ValueTask DisposeAsync(bool disposing)
+    protected override ValueTask DisposeAsync(bool disposing)
     {
         if (disposing && DisposeInstance && !_activated)
         {
             // If the item implements IAsyncDisposable we will call its DisposeAsync Method.
             if (_instance is IAsyncDisposable asyncDisposable)
             {
-                await asyncDisposable.DisposeAsync().ConfigureAwait(false);
+                var vt = asyncDisposable.DisposeAsync();
+                if (vt.IsCompletedSuccessfully)
+                {
+                    return vt;
+                }
+                else
+                {
+                    static async ValueTask Awaiter(ValueTask vt) => await vt.ConfigureAwait(false);
+                    return Awaiter(vt);
+                }
             }
             else if (_instance is IDisposable disposable)
             {
--- a/src/Autofac/Core/Activators/ProvidedInstance/ProvidedInstanceActivator.cs
+++ b/src/Autofac/Core/Activators/ProvidedInstance/ProvidedInstanceActivator.cs
@@ -106,6 +106,7 @@ public class ProvidedInstanceActivator : InstanceActivator, IInstanceActivator
                 {
                     return vt;
                 }
+
                 static async ValueTask Awaiter(ValueTask vt) => await vt.ConfigureAwait(false);
                 return Awaiter(vt);
             }
@@ -113,9 +114,10 @@ public class ProvidedInstanceActivator : InstanceActivator, IInstanceActivator
             {
                 // Call the standard Dispose.
                 disposable.Dispose();
-                return default;
             }
         }
+
+        return default;
 
         // Do not call the base, otherwise the standard Dispose will fire.
     }
--- a/src/Autofac/Core/Activators/ProvidedInstance/ProvidedInstanceActivator.cs
+++ b/src/Autofac/Core/Activators/ProvidedInstance/ProvidedInstanceActivator.cs
@@ -106,16 +106,14 @@ public class ProvidedInstanceActivator : InstanceActivator, IInstanceActivator
                 {
                     return vt;
                 }
-                else
-                {
-                    static async ValueTask Awaiter(ValueTask vt) => await vt.ConfigureAwait(false);
-                    return Awaiter(vt);
-                }
+                static async ValueTask Awaiter(ValueTask vt) => await vt.ConfigureAwait(false);
+                return Awaiter(vt);
             }
             else if (_instance is IDisposable disposable)
             {
                 // Call the standard Dispose.
                 disposable.Dispose();
+                return default;
             }
         }
 
--- a/src/Autofac/Core/Activators/ProvidedInstance/ProvidedInstanceActivator.cs
+++ b/src/Autofac/Core/Activators/ProvidedInstance/ProvidedInstanceActivator.cs
@@ -101,13 +101,7 @@ public class ProvidedInstanceActivator : InstanceActivator, IInstanceActivator
             // If the item implements IAsyncDisposable we will call its DisposeAsync Method.
             if (_instance is IAsyncDisposable asyncDisposable)
             {
-                var vt = asyncDisposable.DisposeAsync();
-
-                // Don't await if it's already completed (this is a slight gain in performance of using ValueTask).
-                if (!vt.IsCompletedSuccessfully)
-                {
-                    await vt.ConfigureAwait(false);
-                }
+                await asyncDisposable.DisposeAsync().ConfigureAwait(false);
             }
             else if (_instance is IDisposable disposable)
             {
--- a/src/Autofac/Core/Disposer.cs
+++ b/src/Autofac/Core/Disposer.cs
@@ -87,13 +87,7 @@ internal class Disposer : Disposable, IDisposer
                     // If the item implements IAsyncDisposable we will call its DisposeAsync Method.
                     if (item is IAsyncDisposable asyncDisposable)
                     {
-                        var vt = asyncDisposable.DisposeAsync();
-
-                        // Don't await if it's already completed (this is a slight gain in performance of using ValueTask).
-                        if (!vt.IsCompletedSuccessfully)
-                        {
-                            await vt.ConfigureAwait(false);
-                        }
+                        await asyncDisposable.DisposeAsync().ConfigureAwait(false);
                     }
                     else if (item is IDisposable disposable)
                     {
--- a/src/Autofac/Core/Registration/ComponentRegistration.cs
+++ b/src/Autofac/Core/Registration/ComponentRegistration.cs
@@ -330,15 +330,17 @@ public class ComponentRegistration : Disposable, IComponentRegistration
                 {
                     return vt;
                 }
+
                 static async ValueTask Awaiter(ValueTask vt) => await vt.ConfigureAwait(false);
                 return Awaiter(vt);
             }
             else
             {
                 Activator.Dispose();
-                return default;
             }
         }
+
+        return default;
 
         // Do not call the base, otherwise the standard Dispose will fire.
     }
--- a/src/Autofac/Core/Registration/ComponentRegistration.cs
+++ b/src/Autofac/Core/Registration/ComponentRegistration.cs
@@ -325,13 +325,7 @@ public class ComponentRegistration : Disposable, IComponentRegistration
             // Check the Activator type so we don't have to force Activators to implement IAsyncDisposable if they don't need it.
             if (Activator is IAsyncDisposable asyncDispose)
             {
-                var vt = asyncDispose.DisposeAsync();
-
-                // Don't await if it's already completed (this is a slight gain in performance of using ValueTask).
-                if (!vt.IsCompletedSuccessfully)
-                {
-                    await vt.ConfigureAwait(false);
-                }
+                await asyncDispose.DisposeAsync().ConfigureAwait(false);
             }
             else
             {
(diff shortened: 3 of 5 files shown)


alistairjevans:
Have you confirmed the assertion about ValueTask  behaviour re:IsCompleted is true on all our supported framework versions? I believe that the optimisation for value task completion certainly was true in earlier versions of the framework, perhaps 4.6.2 when this code was added.

OwnageIsMagic:
This kind of optimization is valid (if implemented properly) in non-async functions where you can check if task is already completed and proceed without allocating IAsyncStateMachine and other async machinery.

I didn't really understand what you asked exactly. If your question is about automatic check of IsCompleted, then this is Roslyn feature and it generally doesn't differentiate codegen based on tfm. 
I'm quite sure that all SDK versions currently shipping by Microsoft have recent enough version of compiler (even Async CTP [spec](https://www.microsoft.com/en-us/download/details.aspx?id=23753) mentions this as possible optimiztion).
Anyway, not awaiting ValueTask is really an error.

alistairjevans:
The optimisation you mention is there based on guidance from Stephen Toub from 2018 (https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/#valid-consumption-patterns-for-valuetasks), with an example that _is_ inside an async method. It's possible the guidance has changed/the compiler doesn't need this any more, but it would be good to understand _why_ that guidance is no longer valid. 

I'm also away from keyboard right now, so will pick this up properly when I get back.

OwnageIsMagic:
@alistairjevans citing:
> The short rule is this: with a ValueTask or a ValueTask<TResult>, you should either await it directly (optionally with .ConfigureAwait(false)) or call AsTask() on it directly, and then never use it again, e.g.

Storing in local variable is not directly awaiting, so you stepping away from safe land.
The only example using `IsCompleted*` is 1) it consuming task (reading `.Result`) 2) it describes another optimization about avoiding calling to `RegisterCancellation` (`CancellationToken.Register` allocates)

alistairjevans:
Right, to remove opinion from this, I've put a benchmark together to compare the existing optimised code with the proposed switch to a straight async:

```csharp
/// <summary>
/// Verify the performance of methods that await a method that completes synchronously.
/// </summary>
public class ValueTaskCompletionBenchmark
{
    public long _incrementing = 0;

    [Benchmark(Baseline = true)]
    public async ValueTask Optimised()
    {
        await OptimisedCompletionWait();
    }

    [Benchmark]
    public async ValueTask Regular()
    {
        await RegularCompletionWait();
    }

    private async ValueTask OptimisedCompletionWait()
    {
        var vt = IncrementAsync();

        if (vt.IsCompletedSuccessfully)
        {
            return;
        }

        await vt.ConfigureAwait(false);
    }

    private async ValueTask RegularCompletionWait()
    {
        await IncrementAsync().ConfigureAwait(false);
    }

    private async ValueTask IncrementAsync()
    { 
        // Do something.
        var result = Interlocked.Increment(ref _incrementing);

        // Never true, but the compiler doesn't know that.
        if (result < 0)
        {
            await Task.Delay(1);
        }
    }
}
```

This yields the following benchmarks:

```

BenchmarkDotNet=v0.13.1, OS=Windows 10.0.19043.2006 (21H1/May2021Update)
Intel Core i7-4770 CPU 3.40GHz (Haswell), 1 CPU, 8 logical and 4 physical cores
.NET SDK=6.0.300-preview.22154.4
  [Host]     : .NET 6.0.2 (6.0.222.6406), X64 RyuJIT
  Job-YQYJFM : .NET 6.0.2 (6.0.222.6406), X64 RyuJIT
  Job-VALRBK : .NET Core 3.1.22 (CoreCLR 4.700.21.56803, CoreFX 4.700.21.57101), X64 RyuJIT
  Job-WDUCVJ : .NET Framework 4.8 (4.8.4515.0), X64 RyuJIT


|    Method |        Job |            Runtime |     Toolchain |      Mean |    Error |   StdDev | Ratio | Allocated |
|---------- |----------- |------------------- |-------------- |----------:|---------:|---------:|------:|----------:|
| Optimised | Job-YQYJFM |           .NET 6.0 |        net6.0 |  68.25 ns | 0.218 ns | 0.204 ns |  0.36 |         - |
|   Regular | Job-YQYJFM |           .NET 6.0 |        net6.0 |  77.11 ns | 0.183 ns | 0.171 ns |  0.40 |         - |
| Optimised | Job-VALRBK |      .NET Core 3.1 | netcoreapp3.1 | 102.67 ns | 0.348 ns | 0.325 ns |  0.54 |         - |
|   Regular | Job-VALRBK |      .NET Core 3.1 | netcoreapp3.1 | 117.75 ns | 0.269 ns | 0.239 ns |  0.62 |         - |
| Optimised | Job-WDUCVJ | .NET Framework 4.8 |         net48 | 191.17 ns | 0.473 ns | 0.442 ns |  1.00 |         - |
|   Regular | Job-WDUCVJ | .NET Framework 4.8 |         net48 | 197.11 ns | 0.315 ns | 0.263 ns |  1.03 |         - |
```

As you can see, the un-optimised proposed changes are slower by significantly more than the margin of error.

Also worth noting that all the places in which we apply this optimisation are the non-generic form of `ValueTask`, as opposed to one that holds a result.

OwnageIsMagic:
I'm sorry that I didn't stated that explicitly enough: this is about correctness, not about performance. Your `Optimised` version gains advantage by violating `ValueTask` contract. If you don't call `IValueTaskSource.GetResult` (through `ValueTask.Result` or some other way) implementation can't recycle pooled object and it will exhaust object pool (if pool is bounded) or leak (if it's unbounded).

---
If you want to gain performance, you should remove `async`/`await` in hot path.
```c#
    [Benchmark]
    public async ValueTask Optimised2()
    {
        await OptimisedProperlyCompletionWait();
    }

    private ValueTask OptimisedProperlyCompletionWait()
    {
        var vt = IncrementAsync();

        if (vt.IsCompletedSuccessfully)
        {
            return vt;
        }

        return Await(vt);

        static async ValueTask Await(ValueTask valueTask) => await valueTask.ConfigureAwait(false);
    }
```
```
BenchmarkDotNet=v0.13.2, OS=Windows 10 (10.0.19044.1526/21H2/November2021Update)
Intel Core i5-8500 CPU 3.00GHz (Coffee Lake), 1 CPU, 6 logical and 6 physical cores
.NET SDK=7.0.100-preview.7.22377.5
  [Host]             : .NET 6.0.7 (6.0.722.32202), X64 RyuJIT AVX2
  .NET 6.0           : .NET 6.0.8 (6.0.822.36306), X64 RyuJIT AVX2
  .NET Core 3.1      : .NET Core 3.1.28 (CoreCLR 4.700.22.36202, CoreFX 4.700.22.36301), X64 RyuJIT AVX2
  .NET Framework 4.8 : .NET Framework 4.8 (4.8.4515.0), X64 RyuJIT VectorSize=256


|     Method |                Job |            Runtime |      Mean |    Error |   StdDev | Ratio | Allocated | Alloc Ratio |
|----------- |------------------- |------------------- |----------:|---------:|---------:|------:|----------:|------------:|
|  Optimised |           .NET 6.0 |           .NET 6.0 |  59.86 ns | 0.318 ns | 0.298 ns |  1.00 |         - |          NA |
| Optimised2 |           .NET 6.0 |           .NET 6.0 |  47.92 ns | 0.210 ns | 0.186 ns |  0.80 |         - |          NA |
|    Regular |           .NET 6.0 |           .NET 6.0 |  69.76 ns | 0.420 ns | 0.393 ns |  1.17 |         - |          NA |
|            |                    |                    |           |          |          |       |           |             |
|  Optimised |      .NET Core 3.1 |      .NET Core 3.1 |  88.76 ns | 0.604 ns | 0.471 ns |  1.00 |         - |          NA |
| Optimised2 |      .NET Core 3.1 |      .NET Core 3.1 |  66.37 ns | 0.322 ns | 0.301 ns |  0.75 |         - |          NA |
|    Regular |      .NET Core 3.1 |      .NET Core 3.1 |  99.80 ns | 0.699 ns | 0.654 ns |  1.12 |         - |          NA |
|            |                    |                    |           |          |          |       |           |             |
|  Optimised | .NET Framework 4.8 | .NET Framework 4.8 | 173.09 ns | 0.360 ns | 0.300 ns |  1.00 |         - |          NA |
| Optimised2 | .NET Framework 4.8 | .NET Framework 4.8 | 125.89 ns | 0.410 ns | 0.320 ns |  0.73 |         - |          NA |
|    Regular | .NET Framework 4.8 | .NET Framework 4.8 | 180.32 ns | 0.411 ns | 0.364 ns |  1.04 |         - |          NA |
```

alistairjevans:
To be perfectly honest I'm still not sure I understand the 'correctness' argument, since when I decompile the code, and look at both the Increment method, and the caller, I can see nothing that pulls from a pool or allocates anything in the synchronous path.  Same goes with the `AsyncValueTaskMethodBuilder`.  If I go and look at something that actually implements `IValueTaskSource`, e.g. `AwaitableSocketEventArgs`, if a method in there completes synchronously, it _also_ [does not allocate/retrieve from pool anything in the synchronous path.](https://source.dot.net/#System.Net.Sockets/System/Net/Sockets/Socket.Tasks.cs,c90dbe3bcfe58d71)

Also, your further-optimised version (while clearly faster due to entirely removing the async state machine in the synchronous path), doesn't call `.Result` in the synchronous path either, so it is hard to argue that it is more correct somehow. It also passes `ValueTask` to a method, which makes me honestly more squeamish than the original (I'm not saying it's incorrect though, from what I can tell this will also function as intended, provided the original `ValueTask` is never accessed after passing it to `await`).

OwnageIsMagic:
`IValueTaskSource` used to implement amortized zero allocation for async path by reusing pooled `IValueTaskSource` objects. As stated in first post we are talking about `IValueTaskSource` backed `ValueTask`s. `Increment` method doesn't do this obviously. Operations completed synchronously neither (they use `T value` constructor of `ValueTask`).

---
You missed one important thing:
We can't check if `ValueTask` was completed synchronously, only that it is already completed at time of check.
`AwaitableSocketEventArgs` on async path calls [Release()](https://source.dot.net/#System.Net.Sockets/System/Net/Sockets/Socket.Tasks.cs,1361) to reset itself and return to the cache.

---
In my example since we don't need result (it is value-less task), so we are delegating consuming of `ValueTask` to caller (and caller as it receives `ValueTask` must consume it exactly once (no more, no less) or pass it through as we did it). You can replace it with
            vt.GetA
            waiter().GetResult();
            return default;
to consume it here.

alistairjevans:
Aaahhh, I think I get it now:

> We can't check if ValueTask was completed synchronously, only that it is already completed at time of check.

That now makes sense in terms of correctness.

Can I ask you to update your PR to use the correct optimised form, then we can merge it in?

Thanks for sticking with this through the questions.

OwnageIsMagic:
Can you confirm that this 10 nanoseconds win will show up in any non synthetic benchmark?

OwnageIsMagic:
I update where it was feasible.

_github codespace is really a thing
committing to open source is one `.` key press away
hope they will add `--amend` some day_

alistairjevans:
> Can you confirm that this 10 nanoseconds win on dispose path will show up in any non synthetic benchmark?

On ASP.NET Core, each request's service scope is disposed using `DisposeAsync` if the service scope supports it (which ours does).  However, a big majority of services that are disposable within our service scope will likely complete disposal synchronously (our default implementation of `Disposable.DisposeAsync` just calls the synchronous `Dispose`). 

So, yes, I can perceive that a sufficiently complex object graph, perhaps utilising nested unit-of-work scopes per API  request, may show some benefit from this optimisation.

OwnageIsMagic:
I mean it won't show up in e2e, because disposal happens after request is finished. To really see it you need request payload of just disposing preallocated scopes (because if you're allocating, this 10 ns advantage will be shadowed by the next gc pause).
Disposal is not on hot path.

Anyway, codecov is now blocking merge, and I don't want to create tests.

OwnageIsMagic:
I think you don't really understand what nanosecond scale is about. Non-cached read from RAM takes about 100 ns.

alistairjevans (approved):
I am fully aware that 10ns is not a lot, and would not provide any obvious gains on single-request timing; but when you're one of our users with a heavily loaded server handling 1000s of requests a second, every CPU cycle we're taking up on one of their threads disposing of a request scope is a CPU cycle that could be used to work on a request.

At the scale of Autofac usage, we get some very interesting performance scenarios bought to us for which parts of the code their profiler indicates is taking up time, so where there is a "known" optimisation we can make for the 90% case such as a fast sync dispose path when async-dispose path is in use, we will take it, even if in isolation it's a tiny amount of performance gain.  Sure we won't necessarily go back and revisit huge swathes of the codebase and break everything we do, but we certainly bear it in mind. 

Anyway, I'm happy with the proposed changes, and the coverage lines are only warning where you now have `return default`, so I'm happy to merge.

PR #1351, opened 2022-10-02
Add support for replacing an instance activator in IComponentRegistration

### LINKED ISSUE #1337: Hook/Extension point modifying IRegistrationBuilder
(I'm not 100% sure that this isn't possible today but afaik it isn't...)

## Problem Statement

We use our own little aop fw for our services. To use this in combination with autofac we use the same technique as eg autofac.extras.dynamicproxy does, having an extension method that replaces the implemented type of the service.
It works fine but as we want this for all our services (replace registered service type with our generated one), it would be nice if there was any hook/extension point or similar in autofac that would be called for every registered component so we could skip this ext method call for each and every registration (many, many, many registrations in our system).

## Desired Solution

Don't know the inner working of autofac good enough to have a good suggestion. `AttachToRegistrationSource` and `Registered` event seem to happen "too late".
Would it be possible to inject something to ContainerBuilder that will be called for each registration maybe?

## Alternatives You've Considered

We can create a composition for ContainerBuilder and extend its methods behavior there. Or we could create our own extension methods directly on ContainerBuilder.
Still, would be great if a "registration hook" would be available in autofac itself, especially for us with a big legacy system with many different processes and containerbuilder instances.

alistairjevans:
Have added a `ReplaceActivator` method that throws exceptions if it is called _after_ the `Registered` event, or if the registration is the `Target` of another. 

It feels unwise to allow modification of a registration's activator if the registration's activator is actually from somewhere else.

Targeting registrations are always created _after_ the targeting registration has been built, so no chance of it losing a reference to the 'correct' activator.

Fixes #1337.

### DIFF
--- a/src/Autofac/Core/IComponentRegistration.cs
+++ b/src/Autofac/Core/IComponentRegistration.cs
@@ -78,4 +78,18 @@ public interface IComponentRegistration : IDisposable, IAsyncDisposable
     /// </summary>
     /// <param name="registryServices">The available services.</param>
     void BuildResolvePipeline(IComponentRegistryServices registryServices);
+
+    /// <summary>
+    /// Replaces the activator for the registration.
+    /// </summary>
+    /// <remarks>
+    /// You can only invoke this method inside
+    /// a <see cref="IComponentRegistryBuilder.Registered"/> handler.
+    /// </remarks>
+    /// <param name="activator">The new activator.</param>
+    /// <exception cref="InvalidOperationException">
+    /// Thrown if <see cref="ReplaceActivator"/> is invoked
+    /// after the registration pipeline has been built, or if this registration targets a different one.
+    /// </exception>
+    void ReplaceActivator(IInstanceActivator activator);
 }
--- a/test/Autofac.Test/Core/ComponentRegistrationTests.cs
+++ b/test/Autofac.Test/Core/ComponentRegistrationTests.cs
@@ -86,4 +86,71 @@ public class ComponentRegistrationTests
 
         Assert.Throws<InvalidOperationException>(() => registration.PipelineBuilding += (s, p) => { });
     }
+
+    [Fact]
+    public void CanReplaceActivatorBeforePipelineHasBeenBuilt()
+    {
+        var services = new Service[] { new TypedService(typeof(object)) };
+
+        var registration = Factory.CreateSingletonRegistration(services, Factory.CreateProvidedInstanceActivator(new object()));
+
+        var newActivator = Factory.CreateProvidedInstanceActivator(new object());
+
+        registration.ReplaceActivator(newActivator);
+
+        Assert.Same(registration.Activator, newActivator);
+    }
+
+    [Fact]
+    public void CannotReplaceActivatorAfterRegistrationHasBeenBuilt()
+    {
+        var services = new Service[] { new TypedService(typeof(object)) };
+
+        var registration = Factory.CreateSingletonRegistration(services, Factory.CreateProvidedInstanceActivator(new object()));
+
+        var builder = Factory.CreateEmptyComponentRegistryBuilder();
+        builder.Register(registration);
+
+        builder.Build();
+
+        Assert.Throws<InvalidOperationException>(() => registration.ReplaceActivator(Factory.CreateProvidedInstanceActivator(new object())));
+    }
+
+    [Fact]
+    public void CannotReplaceActivatorIfTargetingAnotherRegistration()
+    {
+        var services = new Service[] { new TypedService(typeof(object)) };
+
+        var registration = Factory.CreateSingletonRegistration(services, Factory.CreateProvidedInstanceActivator(new object()));
+
+        var targetingRegistration = Factory.CreateTargetingRegistration(registration);
+
+        Assert.Throws<InvalidOperationException>(() => targetingRegistration.ReplaceActivator(Factory.CreateProvidedInstanceActivator(new object())));
+    }
+
+    [Fact]
+    public void ReplacementActivatorLimitTypeMustSatisfyAllServices()
+    {
+        var services = new Service[] { new TypedService(typeof(IServiceA)), new TypedService(typeof(IServiceB)) };
+
+        var registration = Factory.CreateSingletonRegistration(services, Factory.CreateProvidedInstanceActivator(new ComponentAB()));
+
+        Assert.Throws<ArgumentException>(() => registration.ReplaceActivator(Factory.CreateProvidedInstanceActivator(new ComponentA())));
+    }
+
+    private interface IServiceA
+    {
+    }
+
+    private interface IServiceB
+    {
+    }
+
+    public class ComponentA : IServiceA
+    {
+    }
+
+    public class ComponentAB : IServiceA, IServiceB
+    {
+    }
 }
--- a/src/Autofac/Core/Registration/ComponentRegistration.cs
+++ b/src/Autofac/Core/Registration/ComponentRegistration.cs
@@ -165,7 +165,7 @@ public class ComponentRegistration : Disposable, IComponentRegistration
     /// <summary>
     /// Gets the activator for the registration.
     /// </summary>
-    public IInstanceActivator Activator { get; }
+    public IInstanceActivator Activator { get; private set; }
 
     /// <summary>
     /// Gets the lifetime associated with the component.
@@ -242,6 +242,44 @@ public class ComponentRegistration : Disposable, IComponentRegistration
         }
 
         ResolvePipeline = BuildResolvePipeline(registryServices, _lateBuildPipeline);
+    }
+
+    /// <inheritdoc />
+    public void ReplaceActivator(IInstanceActivator activator)
+    {
+        if (activator is null)
+        {
+            throw new ArgumentNullException(nameof(activator));
+        }
+
+        if (_builtComponentPipeline is not null)
+        {
+            throw new InvalidOperationException(ComponentRegistrationResources.ReplaceActivatorPipelineAlreadyBuilt);
+        }
+
+        if (Target != this)
+        {
+            throw new InvalidOperationException(ComponentRegistrationResources.ReplaceActivatorTargetingRegistration);
+        }
+
+        var limitType = activator.LimitType;
+        if (limitType != typeof(object))
+        {
+            foreach (var ts in Services)
+            {
+                if (ts is not IServiceWithType asServiceWithType)
+                {
+                    continue;
+                }
+
+                if (!asServiceWithType.ServiceType.IsAssignableFrom(limitType))
+                {
+                    throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, ComponentRegistrationResources.ReplaceActivatorComponentDoesNotSupportService, limitType, ts));
+                }
+            }
+        }
+
+        Activator = activator;
     }
 
     /// <summary>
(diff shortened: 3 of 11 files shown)


src/Autofac/Core/IComponentRegistration.cs
```diff
@@ -78,4 +78,18 @@ public interface IComponentRegistration : IDisposable, IAsyncDisposable
...
+    /// Replaces the activator for the registration.
+    /// </summary>
+    /// <remarks>
+    /// You can only invoke this method inside
+    /// a <see cref="IComponentRegistryBuilder.Registered"/> handler.
+    /// </remarks>
+    /// <param name="activator">The new activator.</param>
+    /// <exception cref="InvalidOperationException">
+    /// Thrown if <see cref="ReplaceActivator"/> is invoked
+    /// after the registration pipeline has been built, or if this registration targets a different one.
```
tillig, line 92:
```suggestion
    /// after the registration pipeline has been built, or if
    /// this registration has a <see cref="Target"/> of a different registration.
```

Minor clarification to differentiate the previous object "registration pipeline" from the thing being referenced here. It took me a couple of seconds to get it.

test/Autofac.Test/Core/ComponentRegistrationTests.cs
```diff
@@ -86,4 +86,71 @@ public void AttachingToPipelineBuildingShouldFailIfAlreadyBuilt()
...
+    [Fact]
+    public void ReplacementActivatorLimitTypeMustSatisfyAllServices()
+    {
+        var services = new Service[] { new TypedService(typeof(IServiceA)), new TypedService(typeof(IServiceB)) };
+
+        var registration = Factory.CreateSingletonRegistration(services, Factory.CreateProvidedInstanceActivator(new ComponentAB()));
+
+        Assert.Throws<ArgumentException>(() => registration.ReplaceActivator(Factory.CreateProvidedInstanceActivator(new ComponentA())));
+    }
+
```
tillig, line 140: It may just be me being paranoid, but we use services for things like `Startable` that aren't typed. It might be worth adding a test to make sure `AutoActivate` on a non-`IStartable` item still works. Possibly also adapters/decorators, just to close the loop. I mean, I don't see any reason they shouldn't, but the weird edge cases freak me out. "I replaced the activator on my decorator type and now nothing works WHY DID YOU BREAK MY STUFF AUTOFAC?"

test/Autofac.Test/Core/ComponentRegistrationTests.cs
```diff
@@ -86,4 +86,71 @@ public void AttachingToPipelineBuildingShouldFailIfAlreadyBuilt()
 
         Assert.Throws<InvalidOperationException>(() => registration.PipelineBuilding += (s, p) => { });
     }
+
```
tillig, line 89: Would these be part of `Autofac.Specification.Test` since it'd be supported public API?

tillig (changes requested):
Minor doc update and a couple of questions about tests.

test/Autofac.Test/Core/ComponentRegistrationTests.cs
alistairjevans: I don't have a super-strong feeling about it; is that the rule, that any tests for publicly visible API endpoints should be in `Specification.Test`? I pretty much put those tests in here because they were compact, could use the `Factory` class and tested _only_ `ComponentRegistration`, without involving the `Registered` event.

I'm going to add a new test file to `Specification.Test` to cover off `IStartable`, decorators, etc. Perhaps that's enough for defining the public surface?

tillig: Generally what I was trying to do with the `Specification.Test` is separate out "tests _for_ the public API tested _by_ the public API like an end user would consume them" from the other kinds of tests ("tests for X that use internals" or "tests for internal APIs") so we would know during a refactor which tests _should_ be sort of safe to update vs. which ones really shouldn't be touched and should continue to work. Also, from an end user perspective, it's easier to search and find "real examples" if you don't have to determine whether we were using something internal to show the workings of things.

Tests like "here's how to consume this thing from a module" that are almost documentation seemed like they fit under the public API spec test sort of thing.

We never did actually get all of the tests separated out. It was too big of a job to do all at once, so it was more iterative - "if you're touching this thing or adding new tests, put them in the right spot so we don't have to figure it out later."

alistairjevans: Noted, ok, will move some things around.

test/Autofac.Test/Core/ComponentRegistrationTests.cs
alistairjevans: Adding tests to `Specification.Test` for a variety of these features + activator replacement, and just wanted to note that you cannot replace the Activator for an Adapter, because it targets a different registration.  Only registrations that do not target others are permitted.

That restriction is mostly in there to stop people replacing the Activator for the `ExternalComponentRegistration` incorrectly, and to stop them implementing custom `Lazy<T>`, `Meta<T>` activators, but has also caused this problem with adapters (have I said how much I hate adapters before?).

Using `Targets` may be a bad approach though; it may be that I need to think of another way of preventing this, e.g. virtual property `CanReplaceActivator` or something that allows `ExternalComponentRegistration` to say "no".

One of the challenges I face is that the `Registered` event gets called for _all_ registrations added to child scopes, _including_ registrations from an external registry:

```csharp
public void ReplacementActivatorLimitCanReplaceInChildScope()
{

    builder.ComponentRegistryBuilder.Registered += (sender, args) =>
    {
        // I added a `ProvidesService` method to make this code less horrible.
        if (args.ComponentRegistration.ProvidesService<ServiceA>())
        {
            // This runs twice, fails the second time because the registration is an `ExternalComponentRegistration`.
            args.ComponentRegistration.ReplaceActivator(
                new CustomActivator<ServiceA>(_ => new ServiceA(createdCustom: true)));
        }
    };

    builder.RegisterType<ServiceA>();

    var scope = container.BeginLifetimeScope(builder => { });

    var instance = scope.Resolve<ServiceA>();

    Assert.True(instance.CreatedCustom);
}
```

I can imagine this causing people some substantial confusion.  Should the `Registered` event even fire for `ExternalComponentRegistration` instances?

Decorators work fine because they don't create targeting registrations. Don't love the syntax required to extract the decorated service though, but it is there:

```csharp
builder.ComponentRegistryBuilder.Registered += (sender, args) =>
{
    if (args.ComponentRegistration.Activator.LimitType == typeof(ServiceADecorator))
    {
        // Don't love this syntax much, was sort of expecting the automatic 'inject parameters'
        // functionality in the ReflectionActivator to 'just work', but does not.
        args.ComponentRegistration.ReplaceActivator(
            new CustomActivator<ServiceADecorator>(ctxt => new ServiceADecorator(
                (IServiceA)ctxt.Parameters.OfType<TypedParameter>().First(x => x is TypedParameter t && t.Type == typeof(IServiceA)).Value,
                createdCustom: true)));
    }
};
```

tillig: I'm not going to lie, if I step back from this whole thing and think, "Yeah, let's allow folks to replace the activators as part of an event for this edge case over here" I also kind of want to /headdesk because it seems like it's a lot of somewhat confusing work to support this thing that will be a total foot-gun to anyone trying to consume it outside of this one very specific edge case. I mean, maybe the answer really is a custom `RegisterMyWrappedType<T>` extension that then does the AOP or whatever else and registers the proxied type directly instead of trying to replace the activator. Or maybe the answer is to just use the right DynamicProxy decorators and not support some extra custom AOP thing. And, yeah, I'm sorry that means there's no global one-place-to-hook-into-literally-everything-except-the-stuff-that-doesn't-allow-it hook. I'd be OK not having that. It feels like Autofac-as-the-instrument-of-AOP is just as questionable as Autofac-as-the-instrument-of-orcehstrating-your-app-startup-logic.

Anyway, I digress, and I'm a little hangry at the moment.

Re: adapters... uh... maybe you don't get to replace the activator for an adapter and that's just a shortcoming of the first iteration of this feature. I'd be fine with that. (I'll take adapters over ACTNARS any day of the week.)

Re: should `ExternalComponentRegistration` fire `Registered`... I don't have a good answer for that. It seems like it's getting brought in from the external registry so it is kinda getting registered since it's from a different context and this is the first time this context has seen it. But I get that it's also kinda _not_ getting registered because nothing has really changed and it's simply bringing in something that already existed. I'd be afraid if we _didn't_ raise that event, then someone is going to want the `ExternalComponentBroughtInJustNow` event so they can do something with just these special registrations and I'm just not interested in having that. Maybe the `Registered` event needs more info in the args, like whether it's... uh... "inherited" or not. Is inherited the right word? You know what I mean.

alistairjevans: Also thinking about it some more now you're questioning it, I wonder if it wouldn't be better for this use-case to add custom middleware pre-activator to the registration, instead of actually changing the activator.

That means that you can choose to invoke the existing activator if you want by calling `next()` in the middleware, or, if you want to completely replace it, set `ctxt.Instance` and never call `next()`.

The current approach in the PR means you _cannot_ access the existing activator.

Fundamentally activators are just middleware, right? So then this may not even need code changes, since you can already add pipeline middleware to all registrations (e.g. log4net example).

alistairjevans: re:IsExternalRegistration on the Registered event, no superb way to do it to be honest, perhaps checking `registration is ExternalComponentRegistration`? That's internal though.

To be honest, while it may seem a little odd, I don't recall ever seeing an issue where someone complained that the event handlers were doubled up?

tillig: I mean, if it can be done via middleware and all the information is there... 🤷 that's kinda what the middleware approach was all about, right? Maybe it's just a matter of documenting more examples of middleware usage, beyond the log4net one?

alistairjevans: I'll double-check my thought once I'm back at a keyboard, but yeah, maybe we can solve this with docs.

alistairjevans:
It "just works"™ using middleware in front of the activator. Should have realised originally, trust in the pipeline, it is our friend! Closing this PR, will go back on the original issue and close that there as well.

```csharp
public class ActivatorReplacementWithMiddleware
{
    [Fact]
    public void CanReplaceActivatorWithMiddleware()
    {

        builder.ComponentRegistryBuilder.Registered += (sender, args) =>
        {
            if (args.ComponentRegistration.Activator.LimitType == typeof(ServiceA))
            {
                args.ComponentRegistration.PipelineBuilding += (sender, pipeline) =>
                {
                    pipeline.Use(new ActivatorReplacementMiddleware(() => new ServiceA(createdCustom: true)));
                };
            }
        };

        builder.RegisterType<ServiceA>();

        var instance = container.Resolve<ServiceA>();

        Assert.True(instance.CreatedCustom);
    }
}

class ActivatorReplacementMiddleware : IResolveMiddleware
{
    private readonly Func<object> _getInstance;

    public PipelinePhase Phase => PipelinePhase.Activation;

    public ActivatorReplacementMiddleware(Func<object> getInstance)
    {
        this._getInstance = getInstance;
    }

    public void Execute(ResolveRequestContext context, Action<ResolveRequestContext> next)
    {
        // If you want, you can call the existing activator by invoking next,
        // but you don't need to.
        // next(context);

        context.Instance = _getInstance();
    }
} 
```

PR #1353, opened 2022-10-25
Fix memory leak regression

botinko:
Fix memory leak due to use of ConcurrentBag for registrations by ensuring we empty the bag when we're done with it.

Bug was fixed in https://github.com/autofac/Autofac/pull/1257 
AsyncDispose was not aware of this issue. 

image


alistairjevans (approved):
Aaargh, good catch! Thanks for the change. 👍

botinko:
@alistairjevans Sure. Is it possible to make a new release with this fix?

PR #1358, opened 2022-12-10
Use net7 SDK

alistairjevans:
I'm preparing to look into the `required` property functionality, and that will require a net7.0 target, but thought I'd raise a separate PR to get net7 SDK builds underway with the appropriate TFM first.


global.json
```diff
@@ -1,11 +1,12 @@
...
-    "version": "6.0.200",
+    "version": "7.0.100",
     "rollForward": "latestFeature"
   },
 
   "additionalSdks": [
     "3.1.416",
-    "5.0.405"
+    "5.0.405",
+    "6.0.200"
```
tillig, line 10: It won't let me apply a suggestion because it thinks the current `5.0.405` line was deleted 😩 

We should probably build against the latest respective SDK versions:

- 3.1.425
- 5.0.408
- 6.0.403

src/Autofac/Autofac.csproj
```diff
@@ -71,10 +71,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.3">
-      <PrivateAssets>all</PrivateAssets>
-    </PackageReference>
-    <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="6.0.0">
+    <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="7.0.0">
```
tillig, line 74: I _think_ we don't actually need to reference the `NetAnalyzers` package after .NET 6 - it comes "built in" with the compilation.

test/Autofac.Test.Compilation/Autofac.Test.Compilation.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net6.0;net5.0;netcoreapp3.1;</TargetFrameworks>
+    <TargetFrameworks>net6.0;netcoreapp3.1;</TargetFrameworks>
```
tillig, line 4: Should this include net7?

tillig (changes requested):
A couple of minor questions and changes, pretty straightforward stuff.

Something to be aware of - [the OmniSharp issue with .NET 7 is still open](https://github.com/OmniSharp/omnisharp-vscode/issues/5463) so once this goes in, my ability to work in core Autofac becomes somewhat reduced until it's resolved since my primary dev environment is MacOS. It's more a cosmetic than functional problem, but it makes any sort of IntelliSense or pre-compile warnings worthless because you can't tell if it's a real problem or part of the bug.

global.json
alistairjevans: Actually need to take out the `5.0`, it's out of support now, so I'm proposing we no longer target/support it either.  Will up the remaining versions though.

alistairjevans:
Seems odd that the Omnisharp issue hasn't received more attention to be honest. This doesn't have to be merged to develop just yet, I can create a second branch off this one to delay it.

tillig (approved):
🍞

PR #1366, opened 2023-02-04
Mark RegisterGeneratedFactory obsolete

### LINKED ISSUE #1297: Discussion: `RegisterGeneratedFactory` becoming obsolete
We have [a set of extensions](https://github.com/autofac/Autofac/blob/f52dc4d877aaa7428b93ccf799a0e4a265a88c85/src/Autofac/Builder/RegistrationExtensions.cs) for `ContainerBuilder.RegisterGeneratedFactory<T>` that are noted as being effectively obsolete as of Autofac v2. The functionality of these extensions is built into a combination of the `Func<T>` relationship and support for delegate factories. There's no longer any need to manually register a generated factory.

We'd like to mark these extensions as `[Obsolete]` in the next major release (v7.0.0) and then remove them in a subsequent major release (ideally v8.0.0). This will help reduce confusion over why those extensions exist as well as remove some legacy code we won't need to maintain anymore.

**If you are currently using `RegisterGeneratedFactory` and the other factory support doesn't work, please let us know what use cases we're missing!** If there's some compelling reason to keep it, we totally can, but we need to understand what the feature gaps are with `Func<T>` and delegate factory support that is being bridged with `RegisterGeneratedFactory`.

tillig:
Fix #1297.

This marks `RegisterGeneratedFactory` extensions and `GeneratedFactoryActivatorData` with an `[Obsolete]` attribute indicating folks should upgrade to use `Func<T>` or delegate factories. Tests for the feature remain, but the obsolete warnings are suppressed around the tests so we don't see them during the build.

While I was here I picked up a couple of optimizations based on Roslyn analyzers:

- Use conditional assignment
- Use local function
- Make struct readonly

### DIFF
--- a/src/Autofac/Features/Scanning/ScanningRegistrationExtensions.cs
+++ b/src/Autofac/Features/Scanning/ScanningRegistrationExtensions.cs
@@ -86,9 +86,9 @@ internal static partial class ScanningRegistrationExtensions
             rb.RegistrationData.Services.OfType<IServiceWithType>().All(swt =>
                 swt.ServiceType.IsAssignableFrom(t)));
 
-        IRegistrationBuilder<object, ConcreteReflectionActivatorData, SingleRegistrationStyle> TypeBuilderFactory(Type type) => RegistrationBuilder.ForType(type);
+        static IRegistrationBuilder<object, ConcreteReflectionActivatorData, SingleRegistrationStyle> TypeBuilderFactory(Type type) => RegistrationBuilder.ForType(type);
 
-        void SingleComponentRegistration(IComponentRegistryBuilder cr, IRegistrationBuilder<object, ConcreteReflectionActivatorData, SingleRegistrationStyle> scanned) => RegistrationBuilder.RegisterSingleComponent(cr, scanned);
+        static void SingleComponentRegistration(IComponentRegistryBuilder registry, IRegistrationBuilder<object, ConcreteReflectionActivatorData, SingleRegistrationStyle> data) => RegistrationBuilder.RegisterSingleComponent(registry, data);
 
         ScanTypesTemplate(closedTypes, cr, rb, TypeBuilderFactory, SingleComponentRegistration);
     }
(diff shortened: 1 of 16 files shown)


src/Autofac/Features/Scanning/ScanningRegistrationExtensions.cs
```diff
@@ -86,13 +86,11 @@ private static void ScanTypes(IEnumerable<Type> types, IComponentRegistryBuilder
             rb.RegistrationData.Services.OfType<IServiceWithType>().All(swt =>
                 swt.ServiceType.IsAssignableFrom(t)));
 
-        Func<Type, IRegistrationBuilder<object, ConcreteReflectionActivatorData, SingleRegistrationStyle>> scannedConstructor =
-            (type) => RegistrationBuilder.ForType(type);
+        IRegistrationBuilder<object, ConcreteReflectionActivatorData, SingleRegistrationStyle> TypeBuilderFactory(Type type) => RegistrationBuilder.ForType(type);
 
-        Action<IComponentRegistryBuilder, IRegistrationBuilder<object, ConcreteReflectionActivatorData, SingleRegistrationStyle>> register =
-            (cr, scanned) => RegistrationBuilder.RegisterSingleComponent(cr, scanned);
+        void SingleComponentRegistration(IComponentRegistryBuilder cr, IRegistrationBuilder<object, ConcreteReflectionActivatorData, SingleRegistrationStyle> scanned) => RegistrationBuilder.RegisterSingleComponent(cr, scanned);
```
alistairjevans, line 91: Can these be declared as static local methods?

alistairjevans (changes requested):
Looks great, only thing I noticed was a possible instance of a missing `static` on two of the new local methods.

src/Autofac/Features/Scanning/ScanningRegistrationExtensions.cs
tillig: Oh, I see what happened - the analyzer considered `cr` in the body of the function to be a closure over the `ScanTypes` _parameter_ rather than its own local parameter of the same name. I can mark it static, and I'll update the parameter names to disambiguate.

alistairjevans (approved):
👩‍🚀

PR #1369, opened 2023-03-06
Ignore required properties in the default autowiring property selector.

alistairjevans:
Noticed during the docs creation for required properties, before this change any registrations with `PropertiesAutowired` applied would duplicate property injection with required properties.

I decided to just blanket ignore required properties in the `DefaultPropertySelector`. My logic for this was as follows:

- Reflection-activated components will already have applied the mandatory property injection we've just added.
- Lambda-activated components will have to create an instance of the object, so the compiler will force them to set those properties.
- Instance components will have to have been created externally, so the compiler will have forced those properties to be set.

_If_ someone inside a lambda uses `Activator.CreateInstance` directly (for some reason) to bypass the required properties, those required properties won't be injected; but then, why aren't you using the reflection activator?

### DIFF
--- a/src/Autofac/Core/Activators/DefaultPropertySelector.cs
+++ b/src/Autofac/Core/Activators/DefaultPropertySelector.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System.Reflection;
+using System.Runtime.CompilerServices;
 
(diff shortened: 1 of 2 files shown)


tillig (approved):
The logic (re: use reflection vs lambda, etc.) is sound. I dig it. So let it be written, so shall it be done.

PR #1370, opened 2023-03-06
Add additional test for enumerables in load context

alistairjevans:
Add a test to verify that a parent and loaded scope can create a single enumerable that combines components from both scopes.

Thought this was worth adding, while writing the docs.


tillig (approved):
🪘

PR #1374, opened 2023-03-23
Optimize Autofac.Features.OpenGenerics.OpenGenericServiceBinder.TryBindOpenGenericTypedService

SergeiPavlov:
Optimizations:
* Use `foreach` instead of LINQ `.Any()` to avoid closure allocation in `.Any(s => s.Equals(definitionService))` expression
* Memoize `type.GetGenericTypeDefinition()` in static ConcurrentDictionary
* Avoid anonymous temporary objects allocation in `.Select(s => new { ServiceWithType = s, GenericService = s.ServiceType.MakeGenericType(serviceGenericArguments) })`

### DIFF
--- a/src/Autofac/Builder/RegistrationBuilder.cs
+++ b/src/Autofac/Builder/RegistrationBuilder.cs
@@ -132,7 +132,7 @@ public static class RegistrationBuilder
         RegistrationData data,
         IInstanceActivator activator,
         IResolvePipelineBuilder pipelineBuilder,
-        IReadOnlyList<Service> services)
+        Service[] services)
     {
         return CreateRegistration(id, data, activator, pipelineBuilder, services, null);
     }
@@ -155,7 +155,7 @@ public static class RegistrationBuilder
         RegistrationData data,
         IInstanceActivator activator,
         IResolvePipelineBuilder pipelineBuilder,
-        IReadOnlyList<Service> services,
+        Service[] services,
         IComponentRegistration? target)
     {
         if (activator == null)
--- a/src/Autofac/Builder/RegistrationBuilder.cs
+++ b/src/Autofac/Builder/RegistrationBuilder.cs
@@ -132,7 +132,7 @@ public static class RegistrationBuilder
         RegistrationData data,
         IInstanceActivator activator,
         IResolvePipelineBuilder pipelineBuilder,
-        Service[] services)
+        IReadOnlyList<Service> services)
     {
         return CreateRegistration(id, data, activator, pipelineBuilder, services, null);
     }
@@ -155,7 +155,7 @@ public static class RegistrationBuilder
         RegistrationData data,
         IInstanceActivator activator,
         IResolvePipelineBuilder pipelineBuilder,
-        Service[] services,
+        IReadOnlyList<Service> services,
         IComponentRegistration? target)
     {
         if (activator == null)
--- a/src/Autofac/Features/OpenGenerics/OpenGenericServiceBinder.cs
+++ b/src/Autofac/Features/OpenGenerics/OpenGenericServiceBinder.cs
@@ -69,10 +69,7 @@ internal static class OpenGenericServiceBinder
         return false;
     }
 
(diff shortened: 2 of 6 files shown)


tillig (changes requested):
This is cool! I'd just like to avoid the breaking API change if possible - can we add overloads instead of changing the signature?

src/Autofac/Builder/RegistrationBuilder.cs
```diff
@@ -132,7 +132,7 @@ public static class RegistrationBuilder
         RegistrationData data,
         IInstanceActivator activator,
         IResolvePipelineBuilder pipelineBuilder,
-        Service[] services)
+        IReadOnlyList<Service> services)
```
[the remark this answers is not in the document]
SergeiPavlov: I think it is better to stay with `Service[]`. 
Saving one allocation does not worth to change public API

src/Autofac/Features/OpenGenerics/OpenGenericServiceBinder.cs
```diff
@@ -86,24 +92,33 @@ internal static class OpenGenericServiceBinder
         [NotNullWhen(returnValue: true)] out Func<IComponentContext, IEnumerable<Parameter>, object>? constructedFactory,
         [NotNullWhen(returnValue: true)] out Service[]? constructedServices)
     {
-        if (serviceWithType.ServiceType.IsGenericType && !serviceWithType.ServiceType.IsGenericTypeDefinition)
+        var serviceWithTypeServiceType = serviceWithType.ServiceType;
+        if (serviceWithTypeServiceType.IsGenericType && !serviceWithTypeServiceType.IsGenericTypeDefinition)
         {
-            var definitionService = (IServiceWithType)serviceWithType.ChangeType(serviceWithType.ServiceType.GetGenericTypeDefinition());
-            var serviceGenericArguments = serviceWithType.ServiceType.GetGenericArguments();
+            var definitionService = (IServiceWithType)serviceWithType.ChangeType(GetGenericTypeDefinition(serviceWithTypeServiceType));
```
tillig, line 98: Sorry, I did a fast pass at this before - I noticed that there is some type-related caching going on now with this `GenericTypeDefinitionByType` dictionary.

I think that's going to break [the new work that got done to isolate caching so we can unload it and allow AssemblyLoadContexts to be unloaded](https://github.com/autofac/Autofac/blob/2e622bda29521333fa4b6d93466dfbc768f77a0e/src/Autofac/Core/Lifetime/LifetimeScope.cs#L208)

If there's a cache of type info it'll need to be [stored in the `ReflectionCacheSet`](https://github.com/autofac/Autofac/blob/develop/src/Autofac/Core/ReflectionCacheSet.cs) so it can be unloaded if needed.

For example, [we cache all the `IEnumerable<T>` we figure out](https://github.com/autofac/Autofac/blob/86c36ab0a70b0f1f43e7821b5aea73cf9a4e8136/src/Autofac/Util/InternalTypeExtensions.cs#L148) so we can unload those references. This seems similar. I'm unclear if we have tests that cover open generics unloading in isolated scopes since there wasn't any caching like this before.

It may warrant [a handy internal accessor](https://github.com/autofac/Autofac/blob/86c36ab0a7/src/Autofac/Core/InternalReflectionCaches.cs) to make it fast to retrieve.

tillig (changes requested):
Just noticed the caching (which... yeah, don't know how I missed it) and I think we'll need to be able to unload the cache to allow for `AssemblyLoadContext` support. Sorry I didn't see it first time around.

src/Autofac/Features/OpenGenerics/OpenGenericServiceBinder.cs
SergeiPavlov: done

tillig (approved):
This is great! Thanks for the help!

tillig:
Looks like there may be a couple build/style warnings popping up.

tillig (approved):
🎉

PR #1389, opened 2023-08-04
Re-enable RegsiterTypes filtering; better message for RegisterType issues

### LINKED ISSUE #1388: RegisterTypes will attempt to register abstract types (undocumented breaking change in v6.5.0)
## Describe the Bug

In Autofac v6.4.0, a scanning registration using `RegisterTypes` that includes types that cannot be constructed (e.g. interfaces, abstract classes) will work fine.

In Autofac v6.5.0 onwards, it will attempt to register those non-constructible types resulting in an `Autofac.Core.Activators.Reflection.NoConstructorsFoundException` upon building the container.

This is a breaking change that I couldn't find documented anywhere.

## Steps to Reproduce

```c#
public class ReproTest
{
    interface IInterface { }

    class Class : IInterface
    {
    }

    [Fact]
    public void Repro()
    {
        var containerBuilder = new ContainerBuilder();

        containerBuilder
            .RegisterTypes(typeof(IInterface), typeof(Class))
            .AssignableTo<IInterface>();

        var exception = Record.Exception(() => containerBuilder.Build());

        Assert.Null(exception);
    }
}
```

## Expected Behavior

No exception, test passes (v6.4.0 behaviour)

## Exception with Stack Trace

v6.5.0 onwards throws:

```text
Autofac.Core.Activators.Reflection.NoConstructorsFoundException: No accessible constructors were found for the type 'ReproTestProject.ReproTest+IInterface'.
   at Autofac.Core.Activators.Reflection.DefaultConstructorFinder.GetDefaultPublicConstructors(Type type)
   at Autofac.Core.Activators.Reflection.DefaultConstructorFinder.FindConstructors(Type targetType)
   at Autofac.Core.Activators.Reflection.ReflectionActivator.ConfigurePipeline(IComponentRegistryServices componentRegistryServices, IResolvePipelineBuilder pipelineBuilder)
   at Autofac.Core.Registration.ComponentRegistration.BuildResolvePipeline(IComponentRegistryServices registryServices, IResolvePipelineBuilder pipelineBuilder)
   at Autofac.Core.Registration.ComponentRegistration.BuildResolvePipeline(IComponentRegistryServices registryServices)
   at Autofac.Core.Registration.ComponentRegistryBuilder.Build()
   at Autofac.ContainerBuilder.Build(ContainerBuildOptions options)
   at ReproTestProject.ReproTest.<>c__DisplayClass2_0.<Repro>b__0()
```

## Dependency Versions

Autofac: v6.5.0

## Additional Info

A workaround is for the user to filter out abstract types from the types passed to `RegisterTypes` themselves.

In trying to figure out why I was getting this exception after upgrading to v6.5.0 from v6.4.0 I took a look at the diff between the two versions, and the removal of `!t.IsAbstract` in `CanBeRegistered` in [this commit](https://github.com/autofac/Autofac/commit/d3dee0932ca0ce5c6c6dfce44f575ebda1cb0de4#diff-f4569fdec63a956e7026105d7be86ff8371180353b544c9c0fc16b911ab8d720L145) was something that stood out to me that _might_ be the cause.

tillig:
- Fixes #1388 - puts back the type filtering that was on `RegisterTypes` so if "invalid" types (types that generally can't be created via reflection - interfaces, abstract classes, etc.) get passed in, they will be ignored and not registered.
- Refactored some of the methods in the scanning namespace so they are in more appropriately named classes and have better/more descriptive names themselves. That way it's easier to see what they're for and who is using them.
- Adds more up-front error handling for `RegisterType(Type t)` and `RegisterType<T>()` calls so if someone tries to use a type in there that generally can't be created via reflection, an `ArgumentException` will come up and explain the issue rather than waiting for the container build to occur and hitting a "no constructors found" exception.
- Centralized the "check" for what may be generally registerable so it can be shared between assembly scanning and ACTNARS.
- Housekeeping:
  - Fix `parameterised` => `parameterized`. Given the fallback language on the assemblies is `en-US` it seemed like fixing up docs and internal naming to also be `en-US` was reasonable. Also allows spell checkers set to that dictionary to find fewer things.
  - Fix `ctxt` => `context`. Another spell checker issue. Community practice generally either uses `ctx` or `context`, with most .NET code in the framework being a fully spelled-out `context`. Updated references to go with the fully spelled-out version.

I'll add some comments to the important bits for extra attention.

### DIFF
--- a/src/Autofac/Builder/RegistrationBuilder.cs
+++ b/src/Autofac/Builder/RegistrationBuilder.cs
@@ -6,6 +6,7 @@ using Autofac.Core;
 using Autofac.Core.Activators.Delegate;
 using Autofac.Core.Registration;
 using Autofac.Core.Resolving.Pipeline;
+using Autofac.Util;
 
 namespace Autofac.Builder;
 
@@ -61,6 +62,12 @@ public static class RegistrationBuilder
     public static IRegistrationBuilder<TImplementer, ConcreteReflectionActivatorData, SingleRegistrationStyle> ForType<TImplementer>()
         where TImplementer : notnull
     {
+        // Open generics can't be generic type parameters so we don't have to check for that here.
+        if (!typeof(TImplementer).MayAllowReflectionActivation(allowCompilerGenerated: true))
+        {
+            throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, RegistrationBuilderResources.OnlyRegisterableTypesAllowed, typeof(TImplementer)));
+        }
+
         return new RegistrationBuilder<TImplementer, ConcreteReflectionActivatorData, SingleRegistrationStyle>(
             new TypedService(typeof(TImplementer)),
             new ConcreteReflectionActivatorData(typeof(TImplementer)),
@@ -74,6 +81,16 @@ public static class RegistrationBuilder
     /// <returns>A registration builder.</returns>
     public static IRegistrationBuilder<object, ConcreteReflectionActivatorData, SingleRegistrationStyle> ForType(Type implementationType)
     {
+        if (implementationType is null)
+        {
+            throw new ArgumentNullException(nameof(implementationType));
+        }
+
+        if (!implementationType.MayAllowReflectionActivation(allowCompilerGenerated: true) || implementationType.IsGenericTypeDefinition)
+        {
+            throw new ArgumentException(string.Format(CultureInfo.CurrentCulture, RegistrationBuilderResources.OnlyRegisterableTypesAllowed, implementationType));
+        }
+
         return new RegistrationBuilder<object, ConcreteReflectionActivatorData, SingleRegistrationStyle>(
             new TypedService(implementationType),
             new ConcreteReflectionActivatorData(implementationType),
--- a/src/Autofac/Builder/RegistrationBuilderResources.resx
+++ b/src/Autofac/Builder/RegistrationBuilderResources.resx
@@ -1,17 +1,17 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <root>
-  <!-- 
-    Microsoft ResX Schema 
-    
+  <!--
+    Microsoft ResX Schema
+
     Version 2.0
-    
-    The primary goals of this format is to allow a simple XML format 
-    that is mostly human readable. The generation and parsing of the 
-    various data types are done through the TypeConverter classes 
+
+    The primary goals of this format is to allow a simple XML format
+    that is mostly human readable. The generation and parsing of the
+    various data types are done through the TypeConverter classes
     associated with the data types.
-    
+
     Example:
-    
+
     ... ado.net/XML headers & schema ...
     <resheader name="resmimetype">text/microsoft-resx</resheader>
     <resheader name="version">2.0</resheader>
@@ -26,36 +26,36 @@
         <value>[base64 mime encoded string representing a byte array form of the .NET Framework object]</value>
         <comment>This is a comment</comment>
     </data>
-                
-    There are any number of "resheader" rows that contain simple 
+
+    There are any number of "resheader" rows that contain simple
     name/value pairs.
-    
-    Each data row contains a name, and value. The row also contains a 
-    type or mimetype. Type corresponds to a .NET class that support 
-    text/value conversion through the TypeConverter architecture. 
-    Classes that don't support this are serialized and stored with the 
+
+    Each data row contains a name, and value. The row also contains a
+    type or mimetype. Type corresponds to a .NET class that support
+    text/value conversion through the TypeConverter architecture.
+    Classes that don't support this are serialized and stored with the
     mimetype set.
-    
-    The mimetype is used for serialized objects, and tells the 
-    ResXResourceReader how to depersist the object. This is currently not 
+
+    The mimetype is used for serialized objects, and tells the
+    ResXResourceReader how to depersist the object. This is currently not
     extensible. For a given mimetype the value must be set accordingly:
-    
-    Note - application/x-microsoft.net.object.binary.base64 is the format 
-    that the ResXResourceWriter will generate, however the reader can 
+
+    Note - application/x-microsoft.net.object.binary.base64 is the format
+    that the ResXResourceWriter will generate, however the reader can
     read any of the formats listed below.
-    
+
     mimetype: application/x-microsoft.net.object.binary.base64
-    value   : The object must be serialized with 
+    value   : The object must be serialized with
             : System.Runtime.Serialization.Formatters.Binary.BinaryFormatter
             : and then encoded with base64 encoding.
-    
+
     mimetype: application/x-microsoft.net.object.soap.base64
-    value   : The object must be serialized with 
+    value   : The object must be serialized with
             : System.Runtime.Serialization.Formatters.Soap.SoapFormatter
             : and then encoded with base64 encoding.
 
     mimetype: application/x-microsoft.net.object.bytearray.base64
-    value   : The object must be serialized into a byte array 
+    value   : The object must be serialized into a byte array
             : using a System.ComponentModel.TypeConverter
             : and then encoded with base64 encoding.
     -->
@@ -119,5 +119,8 @@
   </resheader>
   <data name="ComponentDoesNotSupportService" xml:space="preserve">
     <value>The type '{0}' is not assignable to service '{1}'.</value>
+  </data>
+  <data name="OnlyRegisterableTypesAllowed" xml:space="preserve">
+    <value>The type '{0}' is not concrete. Only concrete types (no interfaces, abstract classes, open generics, etc.) can be registered by type.</value>
   </data>
 </root>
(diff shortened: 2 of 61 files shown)


src/Autofac/Builder/RegistrationBuilder.cs
```diff
@@ -61,6 +62,12 @@ public static class RegistrationBuilder
     public static IRegistrationBuilder<TImplementer, ConcreteReflectionActivatorData, SingleRegistrationStyle> ForType<TImplementer>()
         where TImplementer : notnull
     {
+        // Open generics can't be generic type parameters so we don't have to check for that here.
```
tillig, line 65: `RegistrationBuilder.ForType` (both overloads) is where all the `RegisterType<T>` and `RegisterType(Type t)` calls come through. It also gets used a few places directly like in ACTNARS and assembly/type scanning.

Adding the check here means we can catch any issues where someone is trying to do something that probably won't work with reflection activation and provide a more proactive error.

It does mean, though, that we may do the same checks twice on a given type. For example, _while scanning the set of types_ to register, we'll filter off things that don't match this criteria; then when the registration gets created it will be checked _again_ here. Ideally it means we won't get any of these errors during any sort of scanning, but it may have container build performance impacts due to the double-check.

src/Autofac/Builder/RegistrationBuilderResources.resx
```diff
@@ -120,4 +120,7 @@
   <data name="ComponentDoesNotSupportService" xml:space="preserve">
     <value>The type '{0}' is not assignable to service '{1}'.</value>
   </data>
-</root>
\ No newline at end of file
+  <data name="OnlyRegisterableTypesAllowed" xml:space="preserve">
+    <value>The type '{0}' is not concrete. Only concrete types (no interfaces, abstract classes, open generics, etc.) can be registered by type.</value>
```
tillig, line 124: This is the exception message that will be thrown if you register types that generally can't be activated by reflection.

src/Autofac/Features/ResolveAnything/AnyConcreteTypeNotAlreadyRegisteredSource.cs
```diff
@@ -55,9 +56,7 @@ public AnyConcreteTypeNotAlreadyRegisteredSource(Func<Type, bool> predicate)
         }
 
         var serviceType = ts.ServiceType;
-        if (!serviceType.IsClass ||
-            serviceType.IsSubclassOf(typeof(Delegate)) ||
-            serviceType.IsAbstract ||
+        if (!serviceType.MayAllowReflectionActivation(allowCompilerGenerated: true) ||
```
tillig, line 59: ACTNARS now uses the shared logic for filtering.

src/Autofac/Features/Scanning/AssemblyExtensions.cs
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) Autofac Project. All rights reserved.
```
tillig, line 1: This all came out of the OpenGenericScanningRegistrationExtensions and ScanningRegistrationExtensions classes.

src/Autofac/Features/Scanning/TypeExtensions.cs
```diff
@@ -0,0 +1,168 @@
+// Copyright (c) Autofac Project. All rights reserved.
```
tillig, line 1: This all came out of the OpenGenericScanningRegistrationExtensions and ScanningRegistrationExtensions classes.

src/Autofac/Features/Scanning/OpenGenericScanningRegistrationExtensions.cs
```diff
@@ -13,7 +13,7 @@ namespace Autofac.Features.Scanning;
 /// <summary>
 /// Helper methods to assist in scanning registration.
 /// </summary>
-internal static partial class ScanningRegistrationExtensions
+internal static class OpenGenericScanningRegistrationExtensions
```
tillig, line 16: Having this just One Big Class was causing some of the confusion with what was consuming what.

src/Autofac/Util/AssemblyExtensions.cs
```diff
@@ -40,22 +40,4 @@ public static IEnumerable<Type> GetLoadableTypes(this Assembly assembly)
             return ex.Types.Where(t => t is not null)!;
         }
     }
-
```
tillig, line 43: This is _only_ used in the scanning namespace, so I moved it there.

src/Autofac/Util/InternalTypeExtensions.cs
```diff
@@ -214,6 +213,33 @@ public static bool IsOpenGenericTypeOf(this Type @this, Type type)
...
+    /// scanning based on its reflection attributes; otherwise <see
+    /// langword="false"/>.
+    /// </returns>
+    // Issue #897: For back compat reasons we can't filter out
+    // non-public types here. Folks use assembly scanning on their
+    // own stuff, so encapsulation is a tricky thing to manage.
+    // If people want only public types, a LINQ Where clause can be used.
+    //
+    // Run IsCompilerGenerated check last due to perf. See AssemblyScanningPerformanceTests.MeasurePerformance.
+    internal static bool MayAllowReflectionActivation(this Type? type, bool allowCompilerGenerated = false) => type != null && type.IsClass && !type.IsAbstract && !type.IsDelegate() && (allowCompilerGenerated || !type.IsCompilerGenerated());
```
tillig, line 241: This gets used a lot - ACTNARS, `RegisterTypes`, `RegistrationBuilder.FromType`, assembly scanning, open generic assembly scanning. The permutations are challenging because sometimes (open generic scanning) you want open generics, but usually you don't; sometimes (ACTNARS) you want compiler generated, but usually you don't.

But at least this gives us a single spot for doing these checks and ensuring the proper order.

test/Autofac.Specification.Test/Registration/TypeRegistrationTests.cs
```diff
@@ -96,6 +106,26 @@ public void RegisterTypeAsUnsupportedService()
         Assert.Throws<ArgumentException>(() => builder.Build());
     }
 
+    [Fact]
```
tillig, line 109: Added some tests for `RegisterType` and `RegisterTypes`, both.

test/Autofac.Specification.Test/Registration/TypeRegistrationTests.cs
```diff
@@ -108,6 +138,30 @@ public void RegisterTypesCanBeFilteredByAssignableTo()
...
+    {
+        var container = new ContainerBuilder().Build().BeginLifetimeScope(b =>
+            b.RegisterTypes(typeof(IMyService), typeof(MyDelegate), typeof(MyAbstractClass), typeof(MyOpenGeneric<>), typeof(MyOpenGeneric<int>), typeof(MyComponent)));
+
+        Assert.Equal(2, container.ComponentRegistry.Registrations.Count());
+        Assert.True(container.TryResolve(typeof(MyComponent), out object _));
+        Assert.True(container.TryResolve(typeof(MyOpenGeneric<int>), out object _));
+        Assert.False(container.TryResolve(typeof(IMyService), out _));
+        Assert.False(container.TryResolve(typeof(MyDelegate), out _));
+        Assert.False(container.TryResolve(typeof(MyAbstractClass), out _));
```
alistairjevans, line 152: Suggest adding a test case for value types?

test/Autofac.Specification.Test/Registration/TypeRegistrationTests.cs
```diff
@@ -96,6 +106,26 @@ public void RegisterTypeAsUnsupportedService()
...
+        Assert.Throws<ArgumentException>(() => builder.RegisterType<IMyService>());
+        Assert.Throws<ArgumentException>(() => builder.RegisterType<MyDelegate>());
+        Assert.Throws<ArgumentException>(() => builder.RegisterType<MyAbstractClass>());
+    }
+
+    [Theory]
+    [InlineData(typeof(MyDelegate))]
+    [InlineData(typeof(IMyService))]
+    [InlineData(typeof(MyAbstractClass))]
+    [InlineData(typeof(MyOpenGeneric<>))]
```
alistairjevans, line 122: As below, suggest a test case for value types.

alistairjevans (changes requested):
For what its worth, this change also improves the error generated if a value type is registered. Before this change, this test:

```csharp
[Fact]
public void ValueTypeActivation()
{
    cb.RegisterType<A>().As<IA>();
    cb.RegisterType<MyValueType>();
    var c = cb.Build();
    c.Resolve<MyValueType>();
}

private struct MyValueType
{
    public MyValueType(IA a)
    {
        A = a;
    }

    public IA A { get; }
}
```

Produces a strange and confusing error:

```
 Message: 
   System.ArgumentException : Expression of type 'Autofac.Specification.Test.Registration.TypeRegistrationTests+MyValueType' cannot be used for return type 'System.Object'

 Stack Trace: 
   Expression.ValidateLambdaArgs(Type delegateType, Expression& body, ReadOnlyCollection`1 parameters, String paramName)
   Expression.Lambda[TDelegate](Expression body, String name, Boolean tailCall, IEnumerable`1 parameters)
   Expression.Lambda[TDelegate](Expression body, Boolean tailCall, IEnumerable`1 parameters)
   Expression.Lambda[TDelegate](Expression body, ParameterExpression[] parameters)
   ConstructorBinder.GetConstructorInvoker(ConstructorInfo constructorInfo) line 165
```

(I believe we generally don't support activating value types, and this isn't just another error.)

Now the error will get raised at `RegisterType`.

alistairjevans (approved):
🐓

PR #1397, opened 2023-10-16
Optimization: convert `ResolveRequest` into readonly struct

SergeiPavlov:
This will save GC work (one allocation per resolving).
Also:
* No need for null-checking
* Remove `GetOrCreateInstanceThrowsArgumentNullExceptionWhenResolveRequestIsNull` test as not-applicable anymore
* Add `in` function argument specifier to pass pointer instead of struct contents.

PR #1401, opened 2023-11-07
Update build/target frameworks for .NET 8; remove netcoreapp3.1.

tillig:
This updates the build to use .NET 8 and adds targeting for .NET 8 on all assemblies.

- Breaking changes:
  - Removed support for .NET 3.1. .NET 3.1 has reached end of life as of December 2022.
  - Removed binary serialization support for `DependencyResolutionException`. [.NET has been steadily working on deprecating binary serialization](https://github.com/dotnet/docs/issues/34893) and as of .NET 8 things are getting caught by both `[Obsolete]` warnings and Roslyn analyzers telling you it's time to move. This both keeps Autofac in line with the trend toward serialization removal and reduces support overhead. People are urged to move away from binary serialization where possible; and where not, it's up to development teams to start creating mechanisms to marshal dependency resolution issues across the wire.
- Updated VS Code extension recommendations to use the new C# Dev Kit from Microsoft.
- Updated all dependencies to latest versions.
- Fixed analyzer warnings.
- Added suppressions for Roslyn analyzers that we can't obey (or, at least, not easily) in a multi-targeting scenario.


alistairjevans (approved):
Did you want to do the major version bump in here, since we'll merge #1397  as well before release?

tillig:
Good point; there's also breaking change due to the exception serialization thing. I'll add that.

alistairjevans (approved):

PR #1414, opened 2024-03-15
`FastGuid.NewGuid()` helper

SergeiPavlov:
It is well-known that standard `Guid.NewGuid()` is slow.
And even slower on Linux https://github.com/dotnet/runtime/issues/13628

If GUIDs are not published outside of process it is enough to use global 64-bit counter as part of GUID generator.

### DIFF
--- a/bench/Autofac.Benchmarks/Benchmarks.cs
+++ b/bench/Autofac.Benchmarks/Benchmarks.cs
@@ -29,5 +29,6 @@ public static class Benchmarks
         typeof(MultiConstructorBenchmark),
         typeof(LambdaResolveBenchmark),
         typeof(RequiredPropertyBenchmark),
+        typeof(RegistrationBenchmark),
     };
 }
--- a/bench/Autofac.Benchmarks/RegistrationBenchmark.cs
+++ b/bench/Autofac.Benchmarks/RegistrationBenchmark.cs
@@ -1,0 +1,46 @@
+﻿namespace Autofac.Benchmarks;
+
+public class RegistrationBenchmark
+{
+    [Benchmark]
+    public void Register()
+    {
+        var builder = new ContainerBuilder();
+        builder.RegisterType<A>();
+        builder.RegisterType<B1>();
+        builder.RegisterType<B2>().InstancePerMatchingLifetimeScope("request");
+        builder.RegisterType<C2>().InstancePerLifetimeScope();
+        builder.RegisterType<D1>().SingleInstance();
+        builder.RegisterType<D2>().SingleInstance();
+        var _container = builder.Build();
+    }
+
+    internal class A
+    {
+        public A(B1 b1, B2 b2) { }
+    }
+
+    internal class B1
+    {
+        public B1(B2 b2, C1 c1, C2 c2) { }
+    }
+
+    internal class B2
+    {
+        public B2(C1 c1, C2 c2) { }
+    }
+
+    internal class C1
+    {
+        public C1(C2 c2, D1 d1, D2 d2) { }
+    }
+
+    internal class C2
+    {
+        public C2(D1 d1, D2 d2) { }
+    }
+
+    internal class D1 { }
+
+    internal class D2 { }
+}
--- a/src/Autofac/Builder/DeferredCallback.cs
+++ b/src/Autofac/Builder/DeferredCallback.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using Autofac.Core.Registration;
+using Autofac.Util;
 
 namespace Autofac.Builder;
 
@@ -25,7 +26,7 @@ public class DeferredCallback
     /// </exception>
     public DeferredCallback(Action<IComponentRegistryBuilder> callback)
     {
-        Id = Guid.NewGuid();
+        Id = FastGuid.NewGuid();
         Callback = callback ?? throw new ArgumentNullException(nameof(callback));
     }
 
--- a/src/Autofac/Builder/SingleRegistrationStyle.cs
+++ b/src/Autofac/Builder/SingleRegistrationStyle.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using Autofac.Core;
+using Autofac.Util;
 
 namespace Autofac.Builder;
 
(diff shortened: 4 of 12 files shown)


alistairjevans:
Thanks, interesting suggestion; can you please run our benchmarks and post before and after results here?

botinko:
Benchmark: 

```
| Method  | Mean      | Error     | StdDev    |
|-------- |----------:|----------:|----------:|
| Default | 34.354 ns | 0.2249 ns | 0.2104 ns |
| Fast    |  8.337 ns | 0.0624 ns | 0.0553 ns |

```

```
    [Benchmark]
    public Guid Default() => Guid.NewGuid();

    [Benchmark]
    public Guid Fast() => FastGuid.NewGuid();
```

For different OS it could be different. In mentioned dotnet runtime issue they say ~900ns on Linux. On Linux `Guid.NewGuid()` uses `/dev/urandom` as a source for randomness. `/dev/urandom`  serve as cryptographically secure pseudorandom number generators. They allow access to a CSPRNG that is seeded with entropy from environmental noise, collected from device drivers and other sources. It's known fact, that it have limited trougput, especially considering that in Kubernetes many containers share one physical machine and possible same `/dev/urandom` stream.

alistairjevans:
Ok, can you compare the existing Autofac benchmarks as well, the ones in our "bench" folder (https://github.com/autofac/Autofac/tree/develop/bench/Autofac.Benchmarks)?

Want to understand the overall impact of the change on Autofac, based on how many guids we generate.

SergeiPavlov:
What are specific Benchmarks which depends on `.NewGuid()`

I ran the entire set Before & After this change and don't see performance improvement
But benchmarks mostly test `Resolve*()` methods, not registrations where GUID may be generated

alistairjevans:
Interesting that it's coming out worse...have you benchmarked dictionary lookups when Guids from FastGuid are used as a key? I'm wondering (without evidence) if the GetHashCode distribution of the guids are not spread out in the same way as with Guid.NewGuid?

SergeiPavlov:
I have tried to mix bits of `Guid`:
```
    public static Guid NewGuid()
    {
        var v = Interlocked.Increment(ref variablePart);
        var h = (int)v * 521;               // one-to-one function to improve Guid's hash distribution
        return new(h, (short)(v >> 32), (short)(v >> 48), BasePart);
    }
```

But it does not show visible improvement.
`Guid.GetHashCode()` XORs all 32-bit parts of Guid, so the distribution should be uniform

SergeiPavlov:
Thanks to @snaumenko-st for further improvement of this PR: https://github.com/autofac/Autofac/pull/1414/commits/88272de910e152b983e5f684c5e16b1beb39e0d8

Benchmark results look not worse than before.

tillig:
Trying to figure out the state of this so I can get a 0.1.0 release out for #1415. I may just release that one without this, regardless.

However, looking at the chain here, it sounds like:

- The first run of benchmarks showed that things actually got _worse_.
- Subsequent runs (not shown) indicate things aren't worse, but also aren't any better.
- The primary use case here is to improve _registration_ time, which happens effectively once per application lifecycle.

It's not interesting to add an optimization that doesn't do anything, so I'm not sure where that leaves us.

If it's something where we can show some _really marked improvement_ like "it used to take 10 seconds to start this app and now it takes half a second" or whatever, I'm all for it. But I'm not super interested in optimizing the "we told Autofac to scan and register every type in the entire app and all assemblies" use case - folks shouldn't be doing that, and the pain may be a forcing factor to tell them they're doing something wrong, especially if we're still supporting `AnyConcreteTypeNotAlreadyRegisteredSource`.

SergeiPavlov:
> * The primary use case here is to improve _registration_ time, which happens effectively once per application lifecycle.

In our case most of restrations happen at App start, but a few additional happen at every Web Request.
That is why we are interested in this improvement.

tillig:
I think I still need to see some sort of improvement, which means this PR possibly needs to include adding some benchmarks that show the affected path and can show that there's improvement. It'd just be nice to see that it's not, like, "2ns improvement" or something basically meaningless. I saw the benchmark of raw GUID creation, but it'd be good to benchmark it in the context of the whole execution path.

SergeiPavlov:
Added a `RegistrationBenchmark`:

Before:
| Method   | Mean     | Error    | StdDev   | Gen0   | Gen1   | Allocated |
|--------- |---------:|---------:|---------:|-------:|-------:|----------:|
| Register | 53.76 us | 0.656 us | 0.548 us | 0.3052 | 0.2441 |  26.77 KB |

After:
| Method   | Mean     | Error    | StdDev   | Gen0   | Gen1   | Allocated |
|--------- |---------:|---------:|---------:|-------:|-------:|----------:|
| Register | 49.55 us | 0.648 us | 0.575 us | 0.3052 | 0.2441 |  27.82 KB |

tillig:
Let me verify I'm reading this right:

After adding the fast GUID optimization, across six registrations you'll save about four microseconds (0.7us per registration) but allocate 1KB more memory to get that.

Does that sound right?

SergeiPavlov:
Yes.

Second run of this benchmark confirms.
Probably new Guids have not so random distribution and Dictionary allocates more buckets for them.

tillig:
Hmmm. Maybe @alistairjevans feels differently, but... I... I have to admit, I kind of lean against this given the trade-off between the microseconds vs. allocations doesn't add up.

alistairjevans:
Yeah, I have to agree, the extra code complexity doesn't seem worth the tiny time gains, and we also don't understand the dictionary bucket impacts enough to know that any gain would be consistent anyway.

SergeiPavlov:
We see following hotspot on `NewGuid()` during profiling
image

So it looks not-insignificant

alistairjevans:
Was that profiling session captured on windows or Linux?

SergeiPavlov:
Linux Debian x64 .NET8

alistairjevans:
Ah; is that also where you've been running benchmarks?

SergeiPavlov:
> Ah; is that also where you've been running benchmarks?

Not exactly
I was running benchmerks on Ubuntu x64 on real PC
but the App profiling was under Docker container in the Azure cloud

May be `.NewGuid()` implementation differs in that environment

SergeiPavlov:
> Is it possible to run the benchmarks in that environment?

I did. The results on Azure are the same as on bare Linux.
`FastGuid` ~10x times faster than `Guid.NewGuid()`

A agree, it is not enough to be merge into Autofac

tillig:
Sounds good. We'll close this one out for now, but if you can show that it will be a marked improvement in some environments maybe we can revive it and target those environments.

PR #1415, opened 2024-03-18
Optimization: Memoize `RequiredMemberAttribute` search

SergeiPavlov:
We see in Profiler that `.GetCustomAttribute()` is a hotspot.
Memoizing it will improve performance.

image


SergeiPavlov:
I see this will not work because of "Unloading feature"
`CanLoadInstanceOfAssemblyAndUnloadItAfterOnActivatedInModule` test

tillig:
We have all of the reflection-related caches [centralized](https://github.com/autofac/Autofac/blob/develop/src/Autofac/Core/InternalReflectionCaches.cs) to allow for unloading assembly load contexts. I'm guessing that's why some of those tests are failing.

SergeiPavlov:
Thank you.
Refactored to use `ReflectionCacheSet`

tillig (approved):
Looks great! Will hold on releasing publicly until we figure out #1414, but you can do a comparison using the MyGet feed `https://www.myget.org/F/autofac/api/v3/index.json` - a `-beta` should show up there based on the build here.

PR #1418, opened 2024-05-23
Enable analyzers on test assemblies

tillig:
I noticed that, while we have a set of test analysis rules, we didn't actually have static analysis running on the test assemblies.

This pull request fixes that:

- Enable static analysis on test assemblies.
- Update the `Test.ruleset` to turn off rules that don't make much sense in tests.
- Fix remaining analyzer issues.

The vast majority of the issues were around things being created in tests and not disposed - `var x` instead of `using var x`. It looks a little messy but it was easy enough to fix.

The second largest number was around nested public types in test fixtures. I think this is something that was hit while cleaning up tests in Autofac.Extensions.DependencyInjection or similar - I remember a bit of churn on one of those. This fixes that. I tried to just mark things private as much as possible, though there were a couple I simply suppressed the issue on rather than try to move things around too much.

There are a few suppressions, mostly around:

- Naming of test stubs that help indicate what they're for - this sometimes conflicts with "good framework naming."
- Odd constructs like set-with-no-get properties that are specifically for testing edge cases.
- Some test fixtures that have nested test fixtures sort of BDD style.
- Some stubs or methods where disposable things are getting created but it's not really clear that we can dispose the thing being created and returned.


alistairjevans (approved):
Looks good. 👍

PR #1421, opened 2024-08-11
Correctly handle polyfilled `required` infrastructure attributes

### LINKED ISSUE #1420: Autofac does not correctly process `required` members for NetFramework (or any version of .NET < 7.0)
## Describe the Bug

Properties declared as `required`[^1] are not correctly injected with Autofac when running anything below NET7.

## Steps to Reproduce

Use the attached unit test project (multi-targeting both NetFramework 472 & NET 8.0)

```c#
using Autofac;
using Autofac.Core;

namespace AfRequiredTests {

    public class Foo {
        public required int Bar { get; set; } = -1;
    }

    public class AfRequiredTests {
        public ILifetimeScope Scope { get; }

        public AfRequiredTests() {

            cb.RegisterType<Foo>()
                .AsSelf();

            this.Scope = cb.Build();

        }

        [Fact]
        public void TestWithMissingRequiredParameter() {
            Assert.Throws<DependencyResolutionException>(
                Scope.Resolve<Foo>
            );
        }

        [Fact]
        public void TestSuppliedRequiredValueIsUsed() {
            const int customValue = 42;

            Assert.Equal(customValue, Scope.Resolve<Foo>(TypedParameter.From(customValue)).Bar);
        }

    }
}
```
[AfRequiredTests.zip](https://github.com/user-attachments/files/16571239/AfRequiredTests.zip)

## Expected Behavior

Correct injection of properties with a `RequiredMemberAttribute` in _all_ (framework) builds of autofac as described in https://autofac.readthedocs.io/en/latest/register/registration.html#required-properties

## Exception with Stack Trace

_Not applicable_

## Dependency Versions

Autofac: 8.0.0
Required: 1.0.0

### Observed bahaviour 
| req. value | NET 8 | NF 4.72 |
|--------|:--------:|:--------:|
| supplied | ✅  | ⛔ <br/> ignored |
| missing | ✅ <br/> (expected exception) |  ⛔ <br/> (no exception) |

### Suggested fix
The bug (feature? 😉 ) is quite obviously buried in https://github.com/autofac/Autofac/blob/ec4907fb6ea61015080b76bcd18cbcf63e43579e/src/Autofac/Core/Activators/Reflection/InjectableProperty.cs#L29C1-L31C7

```cs
internal class InjectableProperty
{
    ...
    public InjectableProperty(PropertyInfo prop)
    {
       ...
#if NET7_0_OR_GREATER
        IsRequired = prop.GetCustomAttribute<RequiredMemberAttribute>() is not null;
#endif
    }
   ...
}   
```
And the current code is _correct_ for NET 7+ (Since there _is_ no predefined _type_ `RequiredMemberAttribute` in previous versions and it just would not build for previous versions).[^2]

But as far as the _compiler_ is concerned, the required feature is not enforced by type ***reference*** but by type ***name***.

So that `#if ...` block should actually have a second branch:
```cs
#if NET7_0_OR_GREATER
        IsRequired = prop.GetCustomAttribute<RequiredMemberAttribute>() is not null;
#else
        IsRequired  = prop.GetCustomAttributesData()
            .Any(
                cad => cad.AttributeType.FullName == "System.Runtime.CompilerServices.RequiredMemberAttribute"
            );
#endif
```

--- 
[^1]: While NetFramework (& NET Core versions below 7) do not _inherently_ support `required` members, they are perfectly able to use this feature using polyfill style code (for example using nuget packages). All the *compiler* requires is a `LangVersion`  offering the feature and an appropriate attribute type declared in the `System.Runtime.CompilerServices` _namespace_ with the _type name_ `RequiredMemberAttribute`. Most prominent nuget package solving this issue: https://www.nuget.org/packages/Required
[^2]: I have no idea which attribute is used if you supply a custom `System.Runtime.CompilerServices.RequiredMemberAttribute` type on a NET7+ build. There _might_ be mismatch issues happening if the custom type is used and AF only checks via (framework) type reference.

DoctorVanGogh:
This fixes #1420 

---

Turned out it was more than a single place where `RequiredMemberAttribute`/`SetsRequiredMemberAttribute` were used. Centralized the logic and removed any NET7+ conditional constraints on source & tests.

Tests include a .NET6 Target Framework which has no inbuilt support for `required`. 
Added the polyfill nuget package in there to allow use of `required` and have tests subsequently passing on NET6.


alistairjevans:
If you decompile a binary with required properties, you will find that the types *do* carry the attribute, indicating that at least one property in the type is required. See spec (https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/required-members#metadata-representation).

We use that as an optimisation to branch down the more expensive code path for the type only if necessary, so I'd leave that in please.

With that knowledge I suspect the memoisation holds as-is?

DoctorVanGogh:
Ah, hadn't actually checked the types themselves with a decompiler, only the properties. That check makes sense then.

As far as I'm concerned, this PR is good to go as-is then.

The memoization still could be tweaked to not just store composite results but cache intermediate analysis as well, but if you're doing that, you could also cache per-property/constructor analysis too and just stick it all into the helper methods.
But I think that would warrant its own PR then.

src/Autofac/Util/ReflectionExtensions.cs
```diff
@@ -121,4 +121,56 @@ public static bool TryGetDeclaringProperty(this ParameterInfo pi, [NotNullWhen(r
 
         return callExpression.Constructor!;
     }
+
+    /// <summary>
+    /// Checks if <paramref name="memberInfo"/> has a <c>RequiredMemberAttribute</c>.
```
tillig, line 126:
```suggestion
    /// Checks if a provided member has a <c>RequiredMemberAttribute</c>.
```

Trying to avoid referencing parameters before they're defined in the doc.

src/Autofac/Util/ReflectionExtensions.cs
```diff
@@ -121,4 +121,56 @@ public static bool TryGetDeclaringProperty(this ParameterInfo pi, [NotNullWhen(r
...
+    /// </returns>
+    public static bool HasRequiredMemberAttribute(
+        this MemberInfo memberInfo)
+    {
+        return memberInfo.CustomAttributes.Any(
+            cad => cad.AttributeType.FullName == "System.Runtime.CompilerServices.RequiredMemberAttribute");
+    }
+
+    /// <summary>
+    /// Checks if <paramref name="constructorInfo"/> has a <c>SetsRequiredMembersAttribute</c>.
```
tillig, line 152:
```suggestion
    /// Checks if a constructor has a <c>SetsRequiredMembersAttribute</c>.
```

src/Autofac/Util/ReflectionExtensions.cs
```diff
@@ -121,4 +121,56 @@ public static bool TryGetDeclaringProperty(this ParameterInfo pi, [NotNullWhen(r
...
+    /// <summary>
+    /// Checks if <paramref name="memberInfo"/> has a <c>RequiredMemberAttribute</c>.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// On NET7+ this would <em>typically</em> be the framework supplied <c>RequiredMemberAttribute</c>, <em>but</em> internally the compiler
+    /// <em>only</em> requires an attribute with that specific type <em>name</em>, not that specific type <em>reference</em>.
+    /// </para>
+    /// <para>
+    /// This could very well be an internally defined custom polyfill attribute using that type name (for example
```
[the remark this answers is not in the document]
tillig: Looks like this also appears below.

src/Autofac/Util/ReflectionExtensions.cs
DoctorVanGogh: While I dont think xmldocs have a concept of ordering, sure ;)

src/Autofac/Util/ReflectionExtensions.cs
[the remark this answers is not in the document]
DoctorVanGogh: Sure, I'll strip this from the xmldocs. Was merely traing to retain the rationale for the non-reference checks.

Would sticking things inside a plain `// ...` code comment still be endorsing? Or should I just go with "for example imported via nuget or entirely self-supplied".

On that same note:
Two **Test** projects now have a `PackageReference` to that Required package (to get the `required` feature working in .NET6). Are those okay?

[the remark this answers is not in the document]
DoctorVanGogh: Okay, removed the mentions and added the custom polyfill in f5346a7bb0115a3a5ef9a2d83dbc01ae56d8cf1f

Had to muzzle stylecop slightly to retain the polyfill's original copyright notice & license. If you have a more "structured" process to include third party notices, feel free to tweak that yourself ;)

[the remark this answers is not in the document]
tillig: It would have been nice to maybe create these from scratch instead of copying them out of an existing library. Part of the point is to show folks how to do this without references and without extra licensing.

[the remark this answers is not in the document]
DoctorVanGogh: Well, considering the polyfill is *purely* functional and *has* to take that specific shape with *those* specific class names & preprocessor guards (okay, one could skip the nullable directives and warning toggles) and me *knowing* where I saw this, whatever I did would *always* be a "derivative" of a pre-existing version. 
And if that's the case I might as well acknowledge where it came from... I'm not too fond of not-invented-here-syndrome, and considering the pre-existing polyfill _is_ MIT licensed.... I'm even less tempted to pretend having done a clean room implementation ;)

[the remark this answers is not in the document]
tillig: @alistairjevans I'll leave this one for you. My thoughts:

- Don't keep the preprocessor directives if we don't need them.
- One class per file, even for the polyfills.
- Possibly make them public rather than internal since they're in the test project anyway.
- It may be arguably fundamentally different enough that it's unique but I'm not sure I care enough to argue it beyond that.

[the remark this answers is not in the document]
DoctorVanGogh: > * Possibly make them public rather than internal since they're in the test project anyway.

I'd strongly discourage that. (At least if the polyfill is supposed to stand as an _example_ for how to do this). Imagine the scenario of someone (re)using that example with `public` attributes. They build a `netstandard2` assembly with it. That assembly then gets consumed in a NET7+ project. Suddenly you have duplicate attribute definitions and a lot of compiler barfing.

src/Autofac/Util/ReflectionExtensions.cs
```diff
@@ -121,4 +121,54 @@ public static bool TryGetDeclaringProperty(this ParameterInfo pi, [NotNullWhen(r
...
+    /// check is done <em>only</em> via type <em>name</em>, not reference.
+    /// </para>
+    /// </remarks>
+    /// <param name="constructorInfo">Constructor to check.</param>
+    /// <returns>
+    /// <see langword="true" /> if <paramref name="constructorInfo"/> carries a <see cref="MemberInfo.CustomAttributes">CustomAttributeData</see> with
+    /// a type <em>name</em> of <c>System.Diagnostics.CodeAnalysis.SetsRequiredMembersAttribute</c>; <see langword="false" /> otherwise.
+    /// </returns>
+    public static bool HasSetsRequiredMembersAttribute(
+        this ConstructorInfo constructorInfo)
```
alistairjevans, line 169: Doesn't `ConstructorInfo` derive from `MemberInfo`? Do you need the overload?

test/RequiredPolyfill.cs
```diff
@@ -0,0 +1,48 @@
+﻿#pragma warning disable SA1636, SA1028 // File header copyright text should match, Code shopuld not contain trailing whitespace
```
alistairjevans, line 1: On the licensing note I'd actually prefer that we copy any shims we need directly from the .NET Source and preserve the MS license, that's what we've done where we have previously copied in shim classes elsewhere I think. Any shim libraries would be based on those anyway, and they're the formal 'specification' of the attribute.

As @tillig mentioned, split the attributes into different files, probably under a 'shims' folder or something?

E.g.
https://github.com/dotnet/runtime/blob/c1ab6c5b2524880ed9368841a0a5d8ead78ea09d/src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/RequiredMemberAttribute.cs

alistairjevans (changes requested):

src/Autofac/Util/ReflectionExtensions.cs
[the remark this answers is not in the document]
alistairjevans: I'm not fussed about making the shims public; they're defined by the .NET Compiler as internal, so if we're looking to create proper shims, we should do the same. See my other note about them being the 'formal' spec for the attribute. 

I would like to be clear thought that this polyfill in this test project is *absolutely* not an example of how to polyfill the required attributes. I'd really hope no-one ends up using an Autofac test project as an example of polyfilling in a distributed library.

src/Autofac/Util/ReflectionExtensions.cs
DoctorVanGogh: Checks for a different attribute: `SetsRequiredMembersAttribute`, not `RequiredMemberAttribute` (and only ever does so for constructors).

alistairjevans: Of course, my bad 🤦

test/RequiredPolyfill.cs
DoctorVanGogh: I mean, yeah, [SetsRequiredMembersAttribute](https://github.com/dotnet/runtime/blob/c1ab6c5b2524880ed9368841a0a5d8ead78ea09d/src/libraries/System.Private.CoreLib/src/System/Diagnostics/CodeAnalysis/SetsRequiredMembersAttribute.cs) is internal too, sadly [CompilerFeatureRequiredAttribute](https://github.com/dotnet/runtime/blob/c1ab6c5b2524880ed9368841a0a5d8ead78ea09d/src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/CompilerFeatureRequiredAttribute.cs) isn't.

DoctorVanGogh: So... you'd want a folder structure like
```
..
    shims
        required
            `License.md` "Files in this folder are slightly tweaked copies from https://github.com/dotnet/runtime and are licensed as ..."
            CompilerFeatureRequiredAttribute.cs
            RequiredMemberAttribute.cs
            SetsRequiredMembersAttribute.cs
```
?

alistairjevans (approved):
Nice! This looks clean, thanks for making the licensing changes.

PR #1429, opened 2024-09-20
Avoid redundant null-checking

SergeiPavlov:
`ResolutionExtensions.TryResolveService()`  already checks `context != null`
No need to check for this in any caller.
Each branching has significant overhead in modern CPUs with high pipe-lining.

Also:
* Use `ArgumentNullException.ThrowIfNull()` instead of direct `if`s .
it is more efficient as explained here: https://dunnhq.com/posts/2022/throw-helper/
(less of "cold code" would be cached in CPU I-cache)

### DIFF
--- a/src/Autofac/ResolutionExtensions.cs
+++ b/src/Autofac/ResolutionExtensions.cs
@@ -194,6 +194,10 @@ public static class ResolutionExtensions
     /// <returns>True if the service is registered.</returns>
     public static bool IsRegisteredService(this IComponentContext context, Service service)
     {
+#if NET7_0_OR_GREATER
+        ArgumentNullException.ThrowIfNull(context);
+        ArgumentNullException.ThrowIfNull(service);
+#else
         if (context == null)
         {
             throw new ArgumentNullException(nameof(context));
@@ -203,6 +207,7 @@ public static class ResolutionExtensions
         {
             throw new ArgumentNullException(nameof(service));
         }
+#endif
 
         return context.ComponentRegistry.IsRegistered(service);
     }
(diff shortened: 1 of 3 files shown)


src/Autofac/ResolutionExtensions.cs
```diff
@@ -194,6 +194,10 @@ public static bool IsRegistered(this IComponentContext context, Type serviceType
     /// <returns>True if the service is registered.</returns>
     public static bool IsRegisteredService(this IComponentContext context, Service service)
     {
+#if NET7_0_OR_GREATER
```
alistairjevans, line 197: I don't like adding a load of additional `#if` blocks like this just to optimise the null check. Can you provide a benchmark with evidence of the gain?  It would have to be a pretty big improvement to justify I think.

alistairjevans (changes requested):
The extra-null-check-removal looks fair, but the extra if-defs make it a bunch more complicated.

src/Autofac/ResolutionExtensions.cs
SergeiPavlov: > Can you provide a benchmark with evidence of the gain

It will be difficult to model the impact of CPU I-cache pollution on a real application.
Typical benchmarks contain little code and CPU cache is big enough.
And big-code benchmark would have many other, may be volatile, factors 

>I don't like adding a load of additional #if

There is other way:  to write custom `ThrowHelper` with the same functionality as `ArgumentNullException.ThrowIfNull()`

tillig: We've definitely avoided adding the ThrowIfNull thing in the past simply because the maintenance overhead has outweighed the micro-perf you might get from that helper. I don't think they do this in the core framework libraries, either.

SergeiPavlov:
Reverted `ArgumentNullException.ThrowIfNull()` `#if`a

Also:
* Removed redundant null-cheking in `RegistrationExtensions`

src/Autofac/ResolutionExtensions.cs
SergeiPavlov: > I don't think they do this in the core framework libraries

They do this from beginning of the "Core fork". e.g.
https://github.com/dotnet/runtime/blame/a0e9764708d5781168c22d86ccb7437f1e54a915/src/coreclr/src/mscorlib/src/System/Collections/Generic/Dictionary.cs#L106

alistairjevans (approved):
Thanks!

PR #1430, opened 2024-09-20
Fix boxing in `ResolveRequest.operator==()`

SergeiPavlov:
`ResolveRequest` is Value type and calling `Equals(object, object)` in `operator==()`  boxes arguments

But this operator is not used anyway.


alistairjevans (approved):
This was probably added to conform to the 'implement equals on value types' analyzer requirement, despite not being needed. Thanks!

Issue #1437, opened 2024-12-04
ALC aware lifetime scope throws SOE when disposed.


hemirunner426:
## Describe the Bug
If I register a concrete class which inherits from an abstract generic class, a stack overflow exception occurs when an ALC child lifetime scope is disposed.

## Steps to Reproduce

I've created a test console app which highlights this issue.  It can be found here:
Scoped Lifetime Bug Repro(https://github.com/hemirunner426/autofaclifetime)

The test console project more closely resembles the project I discovered this issue in.  The project I'm working on has a plugin arch to it and the plugins it loads can run code any time during the lifecycle of the application including bootstrap and shutdown.
I'm looking to ensure that those lifetime scopes are disposed of when the host shuts down.

To reproduce in the test console:
1. Start the project.
2. CTRL+C the console window to invoke IHostedService.StopAsync()
3. The SOE occurs when the child scope is disposed.

NOTE: This is also reproducible if the child scope (the ALC aware lifetime scope) is disposed at any time, not just during StopAsync().  A standard child lifetime scope does not exhibit this behavior.  Also, removing the generic declararion on the abstract class allows the ALC aware scope to dispose properly.

## Expected Behavior

The ALC child scope disposes without an exception.

## Exception with Stack Trace

```text
   System.Collections.Generic.HashSet`1<System.Reflection.Assembly>)
   at Autofac.Util.Cache.TypeAssemblyReferenceProvider.PopulateAllReferencedAssemblies(System.Type, System.Collections.Generic.HashSet`1<System.Reflection.Assembly>)
   at Autofac.Util.Cache.TypeAssemblyReferenceProvider.PopulateAllReferencedAssemblies(System.Type, System.Collections.Generic.HashSet`1<System.Reflection.Assembly>)
   at Autofac.Util.Cache.TypeAssemblyReferenceProvider.PopulateAllReferencedAssemblies(System.Type, System.Collections.Generic.HashSet`1<System.Reflection.Assembly>)
   at Autofac.Util.Cache.TypeAssemblyReferenceProvider.PopulateAllReferencedAssemblies(System.Type, System.Collections.Generic.HashSet`1<System.Reflection.Assembly>)
   at Autofac.Util.Cache.TypeAssemblyReferenceProvider.PopulateAllReferencedAssemblies(System.Type, System.Collections.Generic.HaSystem.Collections.Generic.HashSet`1<System.Reflection.Assembly>)
[...]

PR #1438, opened 2024-12-05
check if generic type assembly has been added to set

hemirunner426:
Check the holding set to see if the type assembly has already been added to avoid the stack overflow.
https://github.com/autofac/Autofac/issues/1437

### DIFF
--- a/src/Autofac/Util/Cache/TypeAssemblyReferenceProvider.cs
+++ b/src/Autofac/Util/Cache/TypeAssemblyReferenceProvider.cs
@@ -68,6 +68,11 @@ internal static class TypeAssemblyReferenceProvider
 
         foreach (var genericArgumentType in genericArguments)
         {
+            if (holdingSet.Any(a => a.DefinedTypes.Contains(genericArgumentType)))
+            {
+                continue;
+            }
+
             PopulateAllReferencedAssemblies(genericArgumentType, holdingSet);
         }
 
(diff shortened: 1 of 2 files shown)


src/Autofac/Util/Cache/TypeAssemblyReferenceProvider.cs
```diff
@@ -68,6 +68,11 @@ private static void PopulateAllReferencedAssemblies(Type inputType, HashSet<Asse
 
         foreach (var genericArgumentType in genericArguments)
         {
+            if (holdingSet.Any(a => a.DefinedTypes.Contains(genericArgumentType)))
```
alistairjevans, line 71: For performance reasons, if this only overflows in a tight loop where we're self-referencing, can we please pass into PopulateAllReferencedAssemblies a "HashSet<Type> activeWorkingSet" and insert there each time we add a type, then check that active working set instead? This should result in a simple O(1) check for the type having already been seen in this tree.

alistairjevans (changes requested):

src/Autofac/Util/Cache/TypeAssemblyReferenceProvider.cs
hemirunner426: Sure thing.  Changes have been committed.

tillig (approved):
Looks good! Thanks for helping out!

PR #1439, opened 2024-12-11
Register service with Interceptor, then register multiple decorator, auto apply Interceptor to the outermost decorator

idiotsky:
Register service with Interceptor, then register multiple decorator, auto apply Interceptor to the outermost decorator
Actually this is the default behaviors when I use autofac 4.x. after upgrade version 8.0.0, this not working any more.
I know explicit configure interceptor on the outermost decorator make it more clarity and flexibility. but I thinks we should keep the original behaviors some how, even it not the default behaviors.
This behaviors will benefit in case when add new decorator, we don't need remove the interceptor and add the interceptor to the new decorator.

### DIFF
--- a/src/Autofac/ServiceMiddlewareRegistrationExtensions.cs
+++ b/src/Autofac/ServiceMiddlewareRegistrationExtensions.cs
@@ -110,6 +110,57 @@ public static class ServiceMiddlewareRegistrationExtensions
     /// </summary>
     /// <param name="builder">The container builder.</param>
     /// <param name="serviceType">The service type.</param>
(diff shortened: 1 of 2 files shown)


idiotsky:
Just Add the unit test for code coverage and create an new pull request, any suggestion, let me know.
https://github.com/autofac/Autofac/issues/1416

tillig (approved):
Looks good - thanks for the help! We appreciate it.

idiotsky:
> Looks good - thanks for the help! We appreciate it.

You are welcome

PR #1447, opened 2025-03-06
Skip assignability check for COM objects

fl3pp:
This PR skip the assignability check of singleton registrations (`ServiceType.IsAssignableFrom(limitType)`) in case the registered instance is a COM object. This check makes it currently impossible to introduce COM object as single instances, as the .NET Runtime handles COM objects specially.

The changes are tested, but the test is not executed on non-Windows platforms due to the nature of COM. The manufactured scenario isn't a completely fair representation of the issue but demonstrates the effect of the productive changes. A full featured example would require an environment with a type library, a working registration for the runtime of the tests, etc., which was deemed too complicated for the proposed changes. 

It also looks like people ran into this issue in the past, as demonstrated by this [poor guy on a Microsoft Q&A](https://learn.microsoft.com/en-us/answers/questions/624191/autofac-throws-exception-for-microsoft-office-inte).

### DIFF
--- a/src/Autofac/Builder/RegistrationBuilder.cs
+++ b/src/Autofac/Builder/RegistrationBuilder.cs
@@ -196,7 +196,7 @@ public static class RegistrationBuilder
         }
 
         var limitType = activator.LimitType;
-        if (limitType != typeof(object))
+        if (limitType != typeof(object) && !limitType.IsCOMObject)
         {
             foreach (var ts in services)
             {
--- a/test/Autofac.Test/Autofac.Test.csproj
+++ b/test/Autofac.Test/Autofac.Test.csproj
@@ -36,6 +36,10 @@
     <None Remove="TestResults\**" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(OS)' != 'Windows_NT'">
+    <Compile Remove="ComObjectRegistrationTest.cs" />
+  </ItemGroup>
+
   <ItemGroup>
     <Compile Include="..\Shims\required\*.cs" Link="Util\shims\required\%(FileName).cs" />
     <None Include="..\Shims\required\README.md" Link="Util\shims\required\README.md" />
--- a/test/Autofac.Test/ComObjectRegistrationTest.cs
+++ b/test/Autofac.Test/ComObjectRegistrationTest.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) Autofac Project. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+namespace Autofac.Test;
+
+public class ComObjectRegistrationTest
+{
+    private const string WindowsScriptingFileSystemObjectProgramId = "Scripting.FileSystemObject";
+
+    [Fact]
+    [SuppressMessage("Interoperability", "CA1416:Validate platform compatibility", Justification = "File is excluded from non-Windows platforms.")]
+    public void AsTest()
+    {
+        var fsoType = Type.GetTypeFromProgID(WindowsScriptingFileSystemObjectProgramId);
+        var fso = Activator.CreateInstance(fsoType);
+
+        var builder = new ContainerBuilder();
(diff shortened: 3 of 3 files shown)


tillig:
Our build only runs on Linux machines and many contributors only have Mac/Linux so this is a challenge to validate.

I'm also a little worried that since - as you said - COM objects are handled a little differently, it means that on Windows machines in this edge case, there could well be other odd side effects that aren't accounted for. Just brainstorming, I don't know how this will work...

- ...as a dependency of another thing
- ...as a transient component that Autofac is tasked to create based on the type
- ...in the relationships like `IEnumerable<T>` or `Lazy<T>`

There are a lot of assumptions in the type and resolution system where we actually need to be able to work with things as known .NET types and not special cases. For example, classic .NET Remoting and WCF used a `TransparentProxy` which [was also a challenge to work with](https://autofac.readthedocs.io/en/latest/advanced/interceptors.html#enable-interception-on-types) and [which we removed back in 2020](https://github.com/autofac/Autofac.Extras.DynamicProxy/releases/tag/v6.0.0) (so I should probably update those docs... hmmm...).

Anyway, I'm not terribly comfortable with our ability to properly validate, maintain, and support this. It seems trivial ("just a one liner!") but we probably need a ton more testing around it (which we couldn't run as devs, and which wouldn't run in the build) and that makes it maybe not great.

For folks who really want to do this, I wonder if a standalone extension library would be a better option. I see the `ComponentRegistration` object is public and has a public constructor; it would be _relatively_ trivial to have a standalone library that has

- It's own `CreateRegistration()` method that allows skipping that check and possibly throws custom exceptions based on COM object registration.
- An extension to `ContainerBuilder` like `RegisterComObject` that handles correct chaining and makes it explicit about what's happening there - using that will bypass any attempt to type check, etc.
- A build that runs on a Windows box and has all the validation tests to ensure you can register it and resolve it as needed.

That'd make this an entirely opt-in feature and any folks who need it can grab the library; others wouldn't need to worry about platform compatibility issues or gotchas.

fl3pp:
Hi @tillig,

Thanks for your fast response!

I think you are absolutely right regarding further possible incompatibilities regarding COM objects and Autofac. Their .NET representation is basically a wrapper around a native object, with a middle-layer that translates between the garbage-collected  world of .NET and the reference counting of native COM called [RCW](https://learn.microsoft.com/en-us/dotnet/standard/native-interop/runtime-callable-wrapper). The behavior of these objects is inconsistent and overall not too much fun...

I also completely understand that you aren't keen on maintaining a Windows technology when your Workstations aren't Windows based.

The custom registration approach is feasible, so thanks for you suggestion! The only downside is that the registration builder interface `IRegistrationBuilder<TLimit, TActivatorData, TRegistrationStyle>` requires lots of methods from an implementation.

Most of these could likely be implemented as extension methods or interface methods (given that they mostly operate on properties that are available on the interface anyway, such ash `RegistrationData`), which would significantly simplify a custom `RegistrationBuilder` implementation.

Anyway, the currently available solutions are to either:
- Copy ~90% of the `RegistrationBuilder`s implementation
- Implement an alternative builder interface which doesn't offer all bells and whistles, even though it would be possible

I think I'll go with the latter approach, and add the missing pieces as we go to avoid too much code duplication.

For an official implementation, I guess that it would make sense to share as much of the `RegistrationBuilders` implementation as possible. Or convert many / most of the method to extension methods or interface methods (as explained above), which would make custom implementations of something like this more feasible.

Feel free to close / reject this PR if you want to. You can also let me know if you are interested in a PR that converts the `IRegistrationBuilders` methods to extension methods as much as possible (I would personally benefit from it 😄).

Thanks again for response and have a nice weekend

tillig:
Probably that abbreviated RegistrationBuilder is what you're looking for. I'm generally fine opening stuff up, but making that one public would probably be a little more API footprint than I'm willing to publicly support for this level of edge case.

Sounds like you have a way forward, so I'll close this. Good luck!

PR #1455, opened 2025-04-23
Update build to use GitHub Actions

tillig:
This switches the overall build and release process to be more modernized and use GitHub Actions and MSBuild.

Run the build with `dotnet msbuild ./default.proj`.

The original Powershell build was written before the `dotnet` command existed, when it was still `dnx`. The notion of `dotnet test` didn't quite work as expected, the MSBuild support wasn't really cross-platform... it was a long time ago. A lot has evolved since then and it's time to update.

- Removed the Powershell and AppVeyor build connections.
- Add GitHub Actions workflows for CI and publishing.
  - Pre-releases (pushes to `develop`) and releases (tags on `master` with `vX.Y.Z` format) will be published to GitHub Packages.
  - Releases (just tags on `master` with `vX.Y.Z` format) will be published to NuGet.
- Added [pre-commit linting](https://pre-commit.com/) to the build, which should also work for any developer wanting to contribute. Lots of whitespace changes due to this with trailing whitespace.
- Added `dotnet format` checking to the build. It isn't part of pre-commit because it seems to want to restore the whole solution just to check one file. We can converge that later if someone figures out how. There are a couple of minor formatting fixes included due to this.
- The build will install whatever the latest SDK versions are so we don't have to track it in `global.json`.
- Updated the way coverage is gathered to work around a known issue in the `coverlet.msbuild` package where sometimes the coverage process gets killed mid-gathering.

I'll need to update the central `CONTRIBUTING.md` to explain the changes here.

I have not been able to actually test the publishing to GitHub Packages or NuGet. It could be that we need to adjust things on the fly the first time we run it. Likely we should do that before converting other repos to this structure.


tillig:
Fixed that. I think there's a potential that the `download-artifact` task may have the wrong path based on where things will unzip, but we can figure that out on the fly.

.github/workflows/build.yml
```diff
@@ -0,0 +1,40 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Setup .NET 6
+        uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: 6.0.x
```
alistairjevans, line 17: I think you can do:

```
with:
    dotnet-version: | 
      6.0.x
      7.0.x
      8.0.x
```
?

alistairjevans (approved):
Nice!

.github/workflows/build.yml
tillig: Nice catch! Somehow I missed that.

tillig:
I'm not sure why the coverage is lower, but it's likely a difference in reporting and due to some of the minor rename/refactor items here.

I'm going to go ahead and merge this in since I want to get past the build challenges and start getting other PRs building using the new system.

PR #1458, opened 2025-05-04
Allow Keyed Composites

syko9000:
I encountered a situation where multiple implementations of a service were registered with the same keys. A composite currently only wraps all implementations of that service that do not have a key.

The workaround involves creating the composite manually, instead of resolving it.
```csharp
var comp = new MyComposite(container.ResolveKeyed<IEnumerable<I1>>("2")) as I1;
```
```csharp
var comp = container.ResolveKeyed<I1>("1");
```
Currently the above line resolves only the last instance of `I1` registered with `"1"`. I needed this to resolve a composite containing just the registrations with a specific key.

This PR allows for that functionality. All behavior remains the same unless the composite is additionally registered with `.Keyed<I1>(...)`, then the above line will resolve an appropriate composite.

Example from added unit test.
```csharp
builder.Register(ctx => new S1()).As<I1>().Keyed<I1>("1");
builder.Register(ctx => new S2()).As<I1>().Keyed<I1>("1");
builder.Register(ctx => new S3()).As<I1>().Keyed<I1>("2");
builder.Register(ctx => new S4()).As<I1>().Keyed<I1>("2");

builder.RegisterComposite<MyComposite, I1>()
    .Keyed<I1>("1")
    .Keyed<I1>("2");

var comp = container.Resolve<I1>(); // gets composite with all I1 without a key (all)
var comp1 = container.ResolveKeyed<I1>("1"); // gets composite with only I1 keyed to "1"
var comp2 = container.ResolveKeyed<I1>("2"); // gets composite with only I1 keyed to "2"
```
Still only one non-keyed service registration is allowed. Keyed registrations must be added to enable this feature.

It works by determining: the resolving service is a composite, it is resolving a parameter of an `IEnumerable<>`, `IList<>`, or `ICollection<>` of the same service type, or the service type is nested in relationship types. Then it does a keyed service resolution using the composites key, else a normal typed resolution.
```csharp
builder.RegisterComposite<MyComplexComposite, I1>()
    .Keyed<I1>("1")
    .Keyed<I1>("2");

var comp2 = container.ResolveKeyed<I1>("2"); // gets composite with only I1 keyed to "2"

private class MyComplexComposite : I1
{
    public MyComplexComposite(IEnumerable<Lazy<Func<int, Owned<Meta<I1>>>>> implementations)
    {
    }
}
```
Will resolve Lazies of a Function that takes an int argument (for demonstration only in this example) that returns an Owned Meta object for `S3` or `S4` only.

### DIFF
--- a/src/Autofac/Core/Activators/Reflection/AutowiringParameter.cs
+++ b/src/Autofac/Core/Activators/Reflection/AutowiringParameter.cs
@@ -42,7 +42,7 @@ public class AutowiringParameter : Parameter
             return type.IsGenericType && type.GenericTypeArguments.Any(genericType => genericType == serviceType || HasService(genericType, serviceType));
         }
 
-        if (context is DefaultResolveRequestContext ctx)
+        if (context is ResolveRequestContext ctx)
         {
             if ((ctx.Registration.Options & Registration.RegistrationOptions.Composite) == Registration.RegistrationOptions.Composite
                 && ctx.Service is KeyedService keyedService
--- a/src/Autofac/Core/Activators/Reflection/AutowiringParameter.cs
+++ b/src/Autofac/Core/Activators/Reflection/AutowiringParameter.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using System.Reflection;
+using Autofac.Core.Resolving.Pipeline;
 
 namespace Autofac.Core.Activators.Reflection;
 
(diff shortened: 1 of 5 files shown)


src/Autofac/Core/Activators/Reflection/AutowiringParameter.cs
```diff
@@ -34,7 +35,27 @@ public override bool CanSupplyValue(ParameterInfo pi, IComponentContext context,
...
+        {
+            return type.IsGenericType && type.GenericTypeArguments.Any(genericType => genericType == serviceType || HasService(genericType, serviceType));
+        }
+
+        if (context is ResolveRequestContext ctx)
+        {
+            if ((ctx.Registration.Options & Registration.RegistrationOptions.Composite) == Registration.RegistrationOptions.Composite
+                && ctx.Service is KeyedService keyedService
+                && pi.ParameterType.IsGenericType
+                && (pi.ParameterType.GetGenericTypeDefinition() == typeof(IEnumerable<>)
```
tillig, line 50: Check out [`CollectionRegistrationSource`](https://github.com/autofac/Autofac/blob/develop/src/Autofac/Features/Collections/CollectionRegistrationSource.cs#L97) to see how we detect generic collections. There's an [extension method `IsGenericListOrCollectionInterfaceType`](https://github.com/autofac/Autofac/blob/develop/src/Autofac/Util/InternalTypeExtensions.cs#L155) that can check but also caches the result so we don't have to look it up every time.

tillig: It may be worth looking at that for detection of things like array parameters or collection types that are derived from an open generic (like `public class MyCollection : IEnumerable<MyType>`).

tillig (changes requested):
I think there's some logic that can be pulled/adapted/reused from `CollectionRegistrationSource` to aid in perf/caching as well as collection type detection logic.

src/Autofac/Core/Activators/Reflection/AutowiringParameter.cs
syko9000: Thank you for the info. I moved to using the extension method that included the IEnumerable<> lookup. I looked through the rest of them to see if any of them already did my recursive lookup, there was not. I moved my lookup into the extension class as well, which feels better. I didn't know there was caching as I looked at the code for the first time on Saturday. I wrapped my lookup into it's own cache as well. Will look into the derived types.

tillig (approved):
This looks really great, thanks for the PR!

PR #1463, opened 2025-08-01
Make 'RequiresUnreferencedCodeAttribute' internal.

### LINKED ISSUE #1462: The 'RequiresUnreferencedCodeAttribute' should be internal
## Describe the Bug

The 'RequiresUnreferencedCodeAttribute' should be internal. When targeting .NET 4.6.2 and using the Polyfill package this causes compile errors.

## Steps to Reproduce

Create a project which targets .NET 4.6.2 and Reference Autofac and Polyfill.

```c#
using System;
using System.Diagnostics.CodeAnalysis;

[RequiresUnreferencedCode("Requires unreferenced code.")]		
public class Test
{
}
```

## Expected Behavior

The code should compile.

Autofac: 8.3.0

prochnowc:
Fixes #1462


tillig:
There is something up with the `dotnet-format` part of the build. It's a one-line change, I'm just going to force it through; but I need to make the build work before I can publish a release with it. That will not be today. It will probably be after the weekend.

PR #1475, opened 2026-02-11
AnyKey support and service key injection

### LINKED ISSUE #1471: Possible improvements to support .NET 10 AnyKey behavior
## Problem Statement

At the moment, the entirety of the support for adapting Microsoft.Extensions.DependencyInjection registrations into Autofac registrations is in the Autofac.Extensions.DependencyInjection repository. Until .NET 10, we were able to keep these things separate because all of the extensions could be done using the current public API for Autofac.

As of .NET 10 [there is more explicit desired behavior for the `AnyKey` construct](https://github.com/dotnet/runtime/blob/dcce6043428b0448f2efe7f72821ab0d19786368/src/libraries/Microsoft.Extensions.DependencyInjection.Specification.Tests/src/KeyedDependencyInjectionSpecificationTests.cs#L41) (where you can register a keyed service to respond to any key).

Converted into Autofac language, it's like

`builder.RegisterType<T>().Keyed("*")`

and that has special meaning. Sometimes it shows up in resolved results, sometimes it doesn't. Based on the table in the unit tests...

| Query                     | Keyed? | Unkeyed? | AnyKey? | null key? |
| --- | --- | --- | --- | --- |
| GetServices(Type)         | no     | yes      | no      | yes |
| GetService(Type)          | no     | yes      | no      | yes |
| GetKeyedServices(null)    | no     | yes      | no      | yes |
| GetKeyedService(null)     | no     | yes      | no      | yes |
| GetKeyedServices(AnyKey)  | yes    | no       | no      | no |
| GetKeyedService(AnyKey)   | throw  | throw    | throw   | throw |
| GetKeyedServices(key)     | yes    | no       | no      | no |
| GetKeyedService(key)      | yes    | no       | yes     | no |
 
- A null key is the same as unkeyed. This allows the KeyServices APIs to support both keyed and unkeyed.
- AnyKey is a special case of Keyed.
- AnyKey registrations are not returned with GetKeyedServices(AnyKey) and GetKeyedService(AnyKey) always throws.
- For IEnumerable, the ordering of the results are in registration order.
- For a singleton resolve, the last match wins.

The problem that's being encountered in trying to make it work in an extension pattern is that if you register something like this...

`builder.RegisterType<T>().Keyed("*")`

then when you resolve a _collection_ of services, _sometimes_ it comes back in the collection and _sometimes_ it doesn't. For example, if you do `container.ResolveKeyed<IEnumerable<T>>("key")` then the AnyKey service does _not_ come back in the collection, but if you resolve a _single_ item `container.ResolveKeyed<T>("key")` then the AnyKey service _does_ come back.

At the moment, in Autofac core, we have some shortcomings:

1. A component can be registered with multiple services, but a keyed service can only respond to a single key. Supporting the concept of AnyKey requires something more dynamic than that.
2. Whether or not a component is included in a collection is controlled at the [component level rather than the service level](https://github.com/autofac/Autofac/blob/17ccc0bb4eae00e1e409679462ef44ecfb875742/src/Autofac/Features/Collections/CollectionRegistrationSource.cs#L123) so you couldn't say "if responding to a keyed service collection request and this was registered with AnyKey, don't include it."

We need to enhance the internals of Autofac somehow to more natively support AnyKey.

## Desired Solution

Ideas include...

- A native concept of AnyKey itself, possibly with a derived version of KeyedService or a well-known key.
- A more dynamic predicate-based service so a service could match any number of criteria.
- Ability to specify "participates in collections" at the service level in addition to the registration level.

This isn't prescriptive, but the idea is that the adapter for AnyKey would be able to use more native Autofac constructs instead of trying to mask over things entirely with AEDI.

tillig:
Resolves #1471.

Adds two primary features that allow for improved support for .NET 10:

1. **Direct support for `AnyKey`**: Microsoft.Extensions.DependencyInjection added the concept of `AnyKey` back in .NET 8, but the behavior of that became far more complex in .NET 10. For example, `AnyKey` items no longer get returned when resolving collections of `AnyKey` - they only get returned in individual resolution calls. This made `AnyKey` much more complex than simply "matching any key," necessitating a more integrated solution with core Autofac.
2. **Ability to inject the service key into constructors or properties:** M.E.DI has a `[ServiceKey]` attribute it uses to inject the service key into an object. With native support for `AnyKey` this seemed like a valuable feature to also bring into Autofac.

This should all be backwards-compatible with existing Autofac keyed services - it's just new API to support these new features. It should also make the Autofac.Extensions.DependencyInjection layer a bit easier to implement since we can map the Microsoft `AnyKey` to the Autofac `AnyKey` and it should "just work."

While working in here I did a couple of one-click optimizations to remove unused usings and similar minor housekeeping things. That increased the size of the PR but seemed not-unreasonable.


tillig (commented):
I've reviewed and iterated over this, verifying that the performance impact is negligible at best, and all the new tests pass. I'm going to call this good so I can see what it will take to update A.E.DI using the new functionality.

Issue #1480, opened 2026-05-18
Performance regression in V9.1


cjlotz:
## Describe the Bug

When updating Autofac from v8.4 to v9.1 we have noticed a considerable degrade in the startup performance for our back-end .NET API server.  Startup has increased from 22 to 34 seconds.  The only change made to the code base were the following updates to Autofac in the `Directory.Packages.Props` as shown below:

<img width="1361" height="186" alt="Image" src="https://github.com/user-attachments/assets/226d0372-936f-4005-a873-ec90163dd6a1" />

We've noticed the efforts in #1478 to try and claw back some performance due to the recent changes to support AnyKey. Is there anything (besides downgrading) that we can do to get the performance back to v8.4 levels?  We don't require AnyKey support and we want to stick with the latest Autofac updates.  

Will it be possible to make AnyKey support an optional feature for those who do not require it and don't won't to incur the performance penalty for a feature they will not use?  Or does isolating/opting out of the AnyKey feature cut too deep?  Any other suggestions

## Steps to Reproduce

- Update Autofac to the latest v9.1 release
- Startup .NET back-end API server
- Notice considerable slowdown in startup performance

tillig:
There's nothing I'm aware of that would be impacting registrations, but, unfortunately, there's also nothing here that I can use to reproduce or test anything. "Start up .NET back-end API server" doesn't mean anything to me. You need to provide a reproduction in the form of something I can run (as indicated by the issue template).

Long startup times usually indicate you're registering _a loooooot_ of stuff. In a lot of cases, when I've seen that, it's people doing assembly scanning and registration, and usually not being very specific about it - like "register everything in this whole assembly" sorts of things instead of registering only what's required.

I can't make AnyKey optional, sorry. This is one of those things that the base .NET folks unfortunately have forced on us by choosing to implement that in a way that we can't easily provide by "extending" something outside core Autofac. I tried hard, believe me, but it just didn't work. That's largely going to affect resolution time, though, not registration time, and there are a lot of hot-path optimizations around skipping AnyKey when it's not being used, so if I had to guess, that's more of a red herring.

Long story short: I can't reproduce it, and there's not enough information here to know exactly what is taking longer, so there's no real way to offer any help at all. I will leave this open for about a week as "feedback required" and if I don't get anything actionable I'll close it.

[...]

PR #1495, opened 2026-07-31
fix: reduce O(types × inherited-members) registration cost from [ServiceKey] scan

npease18:
As initially reported in https://github.com/autofac/Autofac/issues/1480, a performance regression was discovered in the start time of Autofac applications with large numbers of derived types and inherited members. I was able to concretely nail this down using [this example codebase](https://github.com/npease18/example-autofac-regression) where I was able to determine that the `ServiceKey` scan introduces a `O(types × inherited-members)` regression, that can be simplified down to `O(distinct declared members)`.

Here is the result of the benchmark run in that other repo (`./scripts/benchmark.sh 9.0.0 9.1.0 9.3.1 fix`)
```
  Autofac     scan         Build() cold    vs baseline     keyed
  ----------  ---------  --------------  -------------  --------
  9.0.0       absent             0.522s          1.00x         0
  9.1.0       present            3.022s          5.79x         0
  9.3.1       present            2.902s          5.56x         0
  local       present            0.909s          1.74x         0
```

Let me know if there is any additional information I can provide or change!

### DIFF
--- a/bench/Autofac.Benchmarks/ContainerBuildInheritedMembersBenchmark.cs
+++ b/bench/Autofac.Benchmarks/ContainerBuildInheritedMembersBenchmark.cs
@@ -71,11 +71,10 @@ public class ContainerBuildInheritedMembersBenchmark
         // are not cleared on container build (their usage includes resolution), so without
         // this reset only the first invocation would scan anything and BenchmarkDotNet's
         // steady-state measurement would report a warm no-op. Clearing here - rather than in
-        // an [IterationSetup] - keeps the reset inside the measured region whatever
-        // invocation count BenchmarkDotNet picks, and avoids attaching a job attribute that
-        // would compose badly with the jobs Program.cs configures. The clear itself is
-        // proportional to the entries this benchmark created, and is negligible next to the
-        // scan it forces.
+        // an [IterationSetup], which would require [InvocationCount(1)] and yields a
+        // multimodal, high-variance distribution - keeps the reset inside the measured region
+        // whatever invocation count BenchmarkDotNet picks. The clear itself is proportional to
+        // the entries this benchmark created, and is negligible next to the scan it forces.
         ReflectionCacheSet.Shared.Clear();
 
         var builder = new ContainerBuilder();
--- a/bench/Autofac.Benchmarks/ContainerBuildInheritedMembersBenchmark.cs
+++ b/bench/Autofac.Benchmarks/ContainerBuildInheritedMembersBenchmark.cs
@@ -1,0 +1,260 @@
+﻿// Copyright (c) Autofac Project. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Autofac.Core;
+
+namespace Autofac.Benchmarks;
+
+/// <summary>
+/// Measures container build (registration) time for many <em>distinct</em> implementation
+/// types that <em>share a base class</em> exposing many inherited writable properties.
+/// </summary>
+/// <remarks>
+/// <para>
+/// <see cref="Autofac.Core.Activators.Reflection.ReflectionActivator"/> scans each
+/// implementation type for <c>[ServiceKey]</c> on constructor parameters and properties.
+/// Scanning <em>inherited</em> members from the derived type yields
+/// <see cref="System.Reflection.PropertyInfo"/> objects whose <c>ReflectedType</c> is the
+/// derived type, so a per-member attribute cache misses on every (derived type x inherited
+/// member) pair - O(types x inherited members) of attribute reflection during registration.
+/// Scanning declared-only members per type and recursing into the base type collapses that
+/// to O(distinct declared members).
+/// </para>
+/// <para>
+/// This benchmark exists because <see cref="ContainerBuildBenchmark"/> cannot detect that
+/// cost: it registers a single sealed type with no properties and no base class, so there
+/// are no inherited members to rescan, and the per-type memoization makes the scan a one-off
+/// regardless of registration count. The two <c>TypeCount</c> values are the point of this
+/// benchmark - a regression shows up as super-linear growth between them.
+/// </para>
+/// <para>
+/// No <c>[ServiceKey]</c> attribute appears anywhere below, deliberately: the scan cost is
+/// paid by every reflection-based registration whether or not the feature is used.
+/// </para>
+/// </remarks>
+public class ContainerBuildInheritedMembersBenchmark
+{
+    private const int InheritedPropertyCount = 20;
+
+    private List<Type> _implementationTypes = new();
+
+    [Params(250, 1000)]
+    public int TypeCount
+    {
+        get; set;
+    }
+
+    [GlobalSetup]
+    public void Setup()
+    {
+        _implementationTypes = DistinctDerivedTypes(TypeCount);
+
+        // Guard the fixture itself: the benchmark is only meaningful while the registered
+        // types really are distinct and really do inherit a wide set of writable properties.
+        if (_implementationTypes.Distinct().Count() != TypeCount)
+        {
+            throw new InvalidOperationException("Implementation types are not distinct.");
+        }
+
+        var writableInheritedProperties = typeof(WideBase).GetProperties().Count(property => property.CanWrite);
+        if (writableInheritedProperties != InheritedPropertyCount)
+        {
+            throw new InvalidOperationException(
+                $"Expected {InheritedPropertyCount} writable properties on {nameof(WideBase)} but found {writableInheritedProperties}.");
+        }
+    }
+
+    [Benchmark]
+    public void Build()
+    {
+        // The [ServiceKey] scan caches live in the process-wide ReflectionCacheSet.Shared and
+        // are not cleared on container build (their usage includes resolution), so without
+        // this reset only the first invocation would scan anything and BenchmarkDotNet's
+        // steady-state measurement would report a warm no-op. Clearing here - rather than in
+        // an [IterationSetup] - keeps the reset inside the measured region whatever
+        // invocation count BenchmarkDotNet picks, and avoids attaching a job attribute that
+        // would compose badly with the jobs Program.cs configures. The clear itself is
+        // proportional to the entries this benchmark created, and is negligible next to the
+        // scan it forces.
+        ReflectionCacheSet.Shared.Clear();
+
+        var builder = new ContainerBuilder();
+
+        foreach (var implementationType in _implementationTypes)
+        {
+            builder.RegisterType(implementationType);
+        }
+
+        using var container = builder.Build();
+        GC.KeepAlive(container);
+    }
+
+    /// <summary>
+    /// Produces <paramref name="count"/> distinct closed <see cref="Derived{T}"/> types, all
+    /// inheriting <see cref="WideBase"/>. The index is encoded as a fixed-depth nesting of
+    /// <see cref="Wrap{TA, TB}"/> over two markers, which yields distinct type arguments
+    /// without a hand-written type declaration per registration. Nesting depth stays at
+    /// log2(count), so type names never get deep enough for name construction to skew the
+    /// measurement.
+    /// </summary>
+    /// <param name="count">The number of distinct types to produce.</param>
+    /// <returns>The distinct closed implementation types.</returns>
+    private static List<Type> DistinctDerivedTypes(int count)
+    {
+        var bits = 1;
+        while (1 << bits < count)
+        {
+            bits++;
+        }
+
+        var types = new List<Type>(count);
+
+        for (var i = 0; i < count; i++)
+        {
+            var argument = typeof(Zero);
+
+            for (var bit = 0; bit < bits; bit++)
+            {
+                var next = ((i >> bit) & 1) == 1 ? typeof(One) : typeof(Zero);
+                argument = typeof(Wrap<,>).MakeGenericType(argument, next);
+            }
+
+            types.Add(typeof(Derived<>).MakeGenericType(argument));
+        }
+
+        return types;
+    }
+
+    /// <summary>
+    /// Stands in for a framework or domain base class with many injectable properties; every
+    /// registered type inherits these writable members.
+    /// </summary>
+    private class WideBase
+    {
+        public string? Property01
+        {
+            get; set;
+        }
+
+        public string? Property02
+        {
+            get; set;
+        }
+
+        public string? Property03
+        {
+            get; set;
+        }
+
+        public string? Property04
+        {
+            get; set;
+        }
+
+        public string? Property05
+        {
+            get; set;
+        }
+
+        public string? Property06
+        {
+            get; set;
+        }
+
+        public string? Property07
+        {
+            get; set;
+        }
+
+        public string? Property08
+        {
+            get; set;
+        }
+
+        public string? Property09
+        {
+            get; set;
+        }
+
+        public string? Property10
+        {
+            get; set;
+        }
+
+        public string? Property11
+        {
+            get; set;
+        }
+
+        public string? Property12
+        {
+            get; set;
+        }
+
+        public string? Property13
+        {
+            get; set;
+        }
+
+        public string? Property14
+        {
+            get; set;
+        }
+
+        public string? Property15
+        {
+            get; set;
+        }
+
+        public string? Property16
+        {
+            get; set;
+        }
+
+        public string? Property17
+        {
+            get; set;
+        }
+
+        public string? Property18
+        {
+            get; set;
+        }
+
+        public string? Property19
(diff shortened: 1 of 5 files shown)


tillig:
I've enabled the build to run and while I look over the code proper, it seems there are some formatting/style issues to fix to allow the build to progress.

npease18:
> I've enabled the build to run and while I look over the code proper, it seems there are some formatting/style issues to fix to allow the build to progress.

Misfire there, looks like I forgot to run the linters before I pushed up. The latest push should have the correct styling.

src/Autofac/Core/Activators/Reflection/ReflectionActivator.cs
```diff
@@ -141,7 +138,17 @@ private static bool UsesServiceKeyAttribute([DynamicallyAccessedMembers(Activato
...
         // parameter "just in case" you change your mind at runtime.
-        foreach (var constructor in implementationType.GetConstructors(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic))
+        //
+        // We scan DeclaredOnly members and recurse into the base type (memoized per
+        // type via ServiceKeyUsageByType). This is critical for performance: scanning
+        // *inherited* members via GetProperties on a derived type yields PropertyInfo
+        // objects whose ReflectedType is the derived type, so the per-member attribute
+        // cache misses on every (derived type x inherited member) pair - O(types x
+        // inherited members) of attribute reflection. Scanning DeclaredOnly anchors each
+        // member to its declaring type so the cache is shared across all derived types.
```
tillig, line 148: The rewrite finds strictly more than the old scan did — private base properties, members hidden by `new` or `override`, and base constructor parameters all newly come back `true`.

That's safe, and it's the right tradeoff for a perf fix: `_requiresServiceKeyParameter` only gates whether the key parameter is *offered*, and `KeyedServiceKeyParameter.CanSupplyValue` re-checks the attribute per parameter, so a false positive costs one unused `Parameter` and never a wrong injection.

The catch is that nothing in the code says so. This comment covers the performance rationale thoroughly but not the behavior change, and the two tests that touch widened shapes (`...ConstructorParameterOnBaseClassOnly`, `...PrivateServiceKeyPropertyOnBaseClass`) assert the non-leak rather than the widening itself — so a future reader won't infer it. Could you add a note here?

```suggestion
        // member to its declaring type so the cache is shared across all derived types.
        //
        // This also makes the scan more conservative than the one it replaces: it now
        // additionally sees private base properties, members hidden by new or override,
        // and base constructor parameters. That is safe - the flag only gates whether
        // the key parameter is offered, and KeyedServiceKeyParameter.CanSupplyValue
        // re-checks the attribute per parameter - so a false positive costs one unused
        // Parameter, never a wrong injection.
```

src/Autofac/Core/Activators/Reflection/ReflectionActivator.cs
```diff
@@ -152,17 +159,36 @@ private static bool UsesServiceKeyAttribute([DynamicallyAccessedMembers(Activato
...
+    // Memoizes the per-type result so each base type in a hierarchy is scanned once,
+    // regardless of how many derived types share it. The factory captures the annotated
+    // 'type' local (the dictionary key parameter is ignored) so the DynamicallyAccessedMembers
+    // contract flows without an unannotated-to-annotated assignment.
+#pragma warning disable S6612 // Intentionally capture the annotated 'type' local (not the unannotated lambda key) so the DynamicallyAccessedMembers contract flows and trimming stays satisfied.
+    private static bool UsesServiceKeyAttributeCached([DynamicallyAccessedMembers(ActivatorMemberTypes.ActivatedType)] Type type)
+        => ReflectionCacheSet.Shared.Internal.ServiceKeyUsageByType.GetOrAdd(
+            type,
+            _ => UsesServiceKeyAttribute(type));
+#pragma warning restore S6612
```
tillig: Since this suppression is required it'd be good to use the attribute form to match the identical case at `DefaultConstructorFinder.cs:49` — same DAM-annotated-local capture, suppressed with `[SuppressMessage]` rather than a pragma pair.

`System.Diagnostics.CodeAnalysis` is already a global using in `src/Autofac`, so no import is needed:

```suggestion
    [SuppressMessage("Major Code Smell", "S6612:The lambda parameter should be used instead of capturing arguments", Justification = "The factory deliberately reads the [DynamicallyAccessedMembers]-annotated 'type' local rather than the unannotated lambda parameter so the trimming contract flows into UsesServiceKeyAttribute.")]
    private static bool UsesServiceKeyAttributeCached([DynamicallyAccessedMembers(ActivatorMemberTypes.ActivatedType)] Type type)
        => ReflectionCacheSet.Shared.Internal.ServiceKeyUsageByType.GetOrAdd(
            type,
            _ => UsesServiceKeyAttribute(type));
```

test/Autofac.Specification.Test/Features/KeyedServiceTests.cs
```diff
@@ -1075,4 +1075,165 @@ public TVal Value
...
     }
+
+    // The tests below pin down [ServiceKey] behavior for INHERITED members. Autofac decides
+    // once per implementation type whether a keyed resolve needs to carry its key as a
+    // parameter, and it determines that by scanning constructors and properties. How that
+    // scan treats inherited members is easy to change by accident while refactoring it -
+    // reflection surfaces inherited members differently depending on whether the scan walks
+    // the type or walks the hierarchy - so each case where the two strategies can disagree
+    // gets a test.
+    [Fact]
```
tillig: This block comment should go away, and the reasoning it carries should move into the individual tests.

Tests get refactored, reordered, and sometimes sorted alphabetically within a fixture. A "the tests below..." preamble silently stops being true the moment one of these moves or a new test lands in the middle of the run — and nothing catches that. Short comments *inside* each test travel with the test.

Convention here is one or two lines, tagged with the issue number, e.g.:

```csharp
// Issue #1234: Ensure the behavior for a thing is cached properly.
```

Deleting the block (the per-test comments in the comments below carry the same information):

```suggestion
```

test/Autofac.Specification.Test/Features/KeyedServiceTests.cs
```diff
@@ -1075,4 +1075,165 @@ public TVal Value
...
+    public void ResolveKeyedServiceWithServiceKeyPropertyOnBaseClass()
+    {
+        var builder = new ContainerBuilder();
+        builder.RegisterType<DerivedFromKeyAwareBase>().Keyed<DerivedFromKeyAwareBase>("inherited-property").PropertiesAutowired();
```
tillig, line 1084:
```suggestion
    {
        // Issue #1480: A [ServiceKey] property inherited from a base class still receives the key.
        var builder = new ContainerBuilder();
```

test/Autofac.Specification.Test/Features/KeyedServiceTests.cs
```diff
@@ -1075,4 +1075,165 @@ public TVal Value
...
+        var svc = provider.ResolveKeyed<DerivedFromKeyAwareBase>("inherited-property");
+
+        Assert.Equal("inherited-property", svc.Key);
+    }
+
+    [Fact]
+    public void ResolveKeyedServiceWithServiceKeyConstructorParameterOnDeclaringType()
+    {
+        var builder = new ContainerBuilder();
+        builder.RegisterType<OwnKeyConstructorService>().Keyed<OwnKeyConstructorService>("own-parameter");
```
tillig, line 1097: This one is the control case for the inherited tests — `[ServiceKey]` declared directly on the type under test, which worked before and after the change. Worth saying so, since otherwise it reads as redundant with the existing keyed-injection tests further up the file.

```suggestion
    {
        // Issue #1480: Baseline for the inherited cases - [ServiceKey] on the type's own constructor parameter.
        var builder = new ContainerBuilder();
```

test/Autofac.Specification.Test/Features/KeyedServiceTests.cs
```diff
@@ -1075,4 +1075,165 @@ public TVal Value
...
+    }
+
+    [Fact]
+    public void ResolveKeyedServiceWithServiceKeyConstructorParameterOnBaseClassOnly()
+    {
+        // The base constructor's parameter carries [ServiceKey], but the derived constructor
+        // passes its own literal, so the service key must NOT reach the base parameter. This
+        // holds even when the scan notices the base constructor, because the key parameter
+        // only supplies values to parameters that themselves carry the attribute.
+        var builder = new ContainerBuilder();
```
tillig: Good explanation, just longer than it needs to be — and it should carry the issue number. Trimming to the essential claim:

```suggestion
    {
        // Issue #1480: The base constructor parameter carries [ServiceKey], but the derived
        // constructor passes its own literal, so the key must not reach it.
```

test/Autofac.Specification.Test/Features/KeyedServiceTests.cs
```diff
@@ -1075,4 +1075,165 @@ public TVal Value
...
+        Assert.Equal("literal-from-derived", svc.Key);
+    }
+
+    [Fact]
+    public void ResolveKeyedServiceWithPrivateServiceKeyPropertyOnBaseClass()
+    {
+        // A private property on a base class is not part of the supported property injection
+        // surface (only public setters are injected by default), so the key is not supplied
+        // to it. The resolve must still succeed.
+        var builder = new ContainerBuilder();
```
tillig: Same here — trim and tag:

```suggestion
    {
        // Issue #1480: A private [ServiceKey] property on a base class is outside the injection
        // surface (public setters only), so the resolve succeeds with the property left unset.
```

test/Autofac.Specification.Test/Features/KeyedServiceTests.cs
```diff
@@ -1075,4 +1075,165 @@ public TVal Value
...
+
+        Assert.NotNull(svc);
+        Assert.Null(svc.ReadPrivateKey());
+    }
+
+    [Fact]
+    public void ResolveKeyedServiceWithRequiredServiceKeyPropertyOnBaseClass()
+    {
+        var builder = new ContainerBuilder();
+        builder.RegisterType<DerivedFromRequiredKeyAwareBase>().Keyed<DerivedFromRequiredKeyAwareBase>("inherited-required");
```
tillig, line 1140:
```suggestion
    {
        // Issue #1480: An inherited 'required' [ServiceKey] property receives the key even
        // without PropertiesAutowired, because required members are always populated.
        var builder = new ContainerBuilder();
```

test/Autofac.Specification.Test/Features/KeyedServiceTests.cs
```diff
@@ -1075,4 +1075,165 @@ public TVal Value
...
+        var svc = provider.ResolveKeyed<DerivedFromRequiredKeyAwareBase>("inherited-required");
+
+        Assert.Equal("inherited-required", svc.Key);
+    }
+
+    [Fact]
+    public void ResolveKeyedServiceWithDeepHierarchyAndNoServiceKey()
+    {
+        var builder = new ContainerBuilder();
+        builder.RegisterType<DeepLevel3>().Keyed<DeepLevel3>("no-attribute").PropertiesAutowired();
```
tillig, line 1154: This is the one that pins the recursion terminating cleanly, which isn't obvious from the name alone:

```suggestion
    {
        // Issue #1480: A multi-level hierarchy with no [ServiceKey] anywhere resolves normally -
        // the base-type walk terminates without requiring the key parameter.
        var builder = new ContainerBuilder();
```

bench/Autofac.Benchmarks/ContainerBuildInheritedMembersBenchmark.cs
```diff
@@ -0,0 +1,260 @@
...
+    {
+        // The [ServiceKey] scan caches live in the process-wide ReflectionCacheSet.Shared and
+        // are not cleared on container build (their usage includes resolution), so without
+        // this reset only the first invocation would scan anything and BenchmarkDotNet's
+        // steady-state measurement would report a warm no-op. Clearing here - rather than in
+        // an [IterationSetup] - keeps the reset inside the measured region whatever
+        // invocation count BenchmarkDotNet picks, and avoids attaching a job attribute that
+        // would compose badly with the jobs Program.cs configures. The clear itself is
+        // proportional to the entries this benchmark created, and is negligible next to the
+        // scan it forces.
```
tillig: This justification is mostly right, but one clause isn't accurate and it'll steer future contributors away from something that actually works. The claim that an `[IterationSetup]` "would compose badly with the jobs Program.cs configures" doesn't hold — I tried it:

```csharp
[IterationSetup]
public void ClearCaches() => ReflectionCacheSet.Shared.Clear();

[Benchmark]
[InvocationCount(1)]
public void Build()
```

That builds clean and composes fine with both jobs — the attribute just folds into each job's config (`Source(InvocationCount=1, UnrollFactor=1)` and `Package-9.3.1(Arguments=..., InvocationCount=1, UnrollFactor=1)`). No conflict, no dropped job.

**Your choice of clearing in-method is still the correct one** — just for a different reason. `[IterationSetup]` requires `[InvocationCount(1)]`, so every sample pays full cold cost plus GC noise, and the distribution falls apart:

| Variant | TypeCount | Source | StdDev | RatioSD | Warnings |
|---|---|---|---|---|---|
| in-method clear (this PR) | 250 | 5.596 ms | 0.195 | 0.02 | none |
| in-method clear (this PR) | 1000 | 28.900 ms | 0.426 | 0.01 | none |
| `[IterationSetup]` | 250 | 7.941 ms | 0.589 | **0.88** | multimodal (mValue 4.35) |
| `[IterationSetup]` | 1000 | 37.261 ms | **8.676** | 0.09 | multimodal (mValue 4.67) |

A RatioSD of 0.88 makes that row useless as a regression gate. Amortizing over `UnrollFactor=16` is what buys the ~2% StdDev and the clean ratios you have now.

Could you swap the composition claim for the measurement-quality one, so the comment documents the real tradeoff?

```suggestion
        // steady-state measurement would report a warm no-op. Clearing here - rather than in
        // an [IterationSetup], which would require [InvocationCount(1)] and yields a
        // multimodal, high-variance distribution - keeps the reset inside the measured region
        // whatever invocation count BenchmarkDotNet picks. The clear itself is proportional to
        // the entries this benchmark created, and is negligible next to the scan it forces.
```

tillig (changes requested):

tillig:
Thank you for this — outside PRs are a rarity for us and genuinely appreciated, and this is a good one. You found a real, non-obvious performance bug, diagnosed it correctly, and fixed it in the right place, with benchmarks and tests to back it up. That's the whole package.

I reproduced the root cause and the improvement independently before reviewing, since a caching fix is the kind of thing worth confirming from scratch.

Reflection hands back a distinct `PropertyInfo` instance per reflecting type, so the `ServiceKeyPropertyAttributes` cache (keyed on `PropertyInfo`) missed on every `(derived type x inherited member)` pair:

```
D1 prop ReflectedType=D1 DeclaringType=WideBase
ReferenceEquals(p1,p2)=False  p1.Equals(p2)=False
distinct cache entries for the same logical property across D1/D2/Base = 3
```

Measuring distinct derived types over a 20-property base (net10.0, Release):

| types | develop | this PR |
|---|---|---|
| 250 | 53.1 ms | 43.5 ms |
| 1000 | 199.8 ms | 42.3 ms |
| 2000 | 932.8 ms | 84.7 ms |

Baseline growth is clearly super-linear (4.7x from 1000 to 2000); this PR is roughly linear. Matches the numbers in your PR description.

Also verified:

- `src/Autofac` Release across all 4 TFMs: 0 warnings, 0 errors.
- `Autofac.Test` 870/870 and `Autofac.Specification.Test` 504/504 pass.
- Both new suppressions are genuinely necessary — I removed each and confirmed the build fails (`S6612`, `IL2072`). Neither is defensive noise, which is nice to see.
- Your new tests really do guard the recursion, not just the `DeclaredOnly` switch: I applied a naive half-fix (`DeclaredOnly` with no `BaseType` walk) and 2 of the 6 fail. They're pinning the right thing.

**Everything I've left in the inline comments is cosmetic — comment wording, one suppression written as `[SuppressMessage]` instead of a pragma pair, and one sentence in a benchmark comment that I measured and found isn't accurate. No correctness concerns, and nothing that changes your approach.** The code itself is good to go.

The one item worth a real look is the comment on `ReflectionActivator.cs` line 148: the rewrite intentionally detects strictly more than the old scan did, which is both safe and the right call here, but nothing currently says so out loud. I'd just like that written down for whoever reads it next.

npease18:
Just addressed those changes via the fixup! commit. If everything looks good, I will go ahead and squash it back down!

tillig (approved):
This is great! Thanks for your help and your patience. I'll see about getting a release out with this in it shortly.

PR #1501, opened 2026-09-17
Return a configurable builder from RegisterDecorator.

idiotsky:
The RegisterDecorator overloads taking a decorator type or factory returned void and built the decorator registration internally, calling CreateRegistration() - which clones the pipeline - before returning. There was nothing to chain configuration onto, and the pipeline was frozen by the time the caller got control back.

They now return IDecoratorRegistrationBuilder<TService>, exposing ConfigurePipeline, WithMetadata, WithCondition and the registration metadata. This is deliberately a new, narrow builder rather than IRegistrationBuilder: As, Named, Keyed and the lifetime methods are declared directly on IRegistrationBuilder and are meaningless or harmful for a decorator, whose service is fixed by DecoratorService and whose lifetime is governed by the decoration machinery.

Creating the component registration is deferred to a single container build callback, so configuration applied any time before Build() is picked up. RegisterGenericDecorator is unchanged: open generic decorators build their closed registration lazily at resolve time, so there is no registration-time builder to hand back.

Middleware added through ConfigurePipeline runs on the decorator's own registration pipeline, which is what lets an integration wrap the outermost decorator instead of the component being decorated.

Source-compatible for existing callers, but binary-breaking on the strong-named assembly, so this needs to ship in a major release. The version in default.proj is left alone.

Refs #1416


tillig:
Since this is a breaking API change, I don't know that I'll accept this _immediately_. I just released a major version and avoided this (though it is a valuable change) so I may need to "save this for later."

That said, without doing a deep review - one of the things I've been trying to get to in the repo for a while is the separation of "tests through the public API" from "tests that require use of an internal API."

The "Autofac.Specification.Test" project is the "test through the public API" set of tests - these are the ones that we shouldn't change as we modify internals. "Autofac.Test" is continued testing that requires internal access. This is admittedly a work in progress - we started this separation before AI agents were really a thing and it would have caused a lot of churn. It may be something I'll look at finishing later. Anyway - we do have a decorator testing fixture [here](https://github.com/autofac/Autofac/blob/develop/test/Autofac.Specification.Test/Features/DecoratorTests.cs) and I think the tests here that verify how one might access pipeline configurations and such through this new feature would go in there rather than Autofac.Test, ideally reusing types and things in there.

idiotsky:
> Since this is a breaking API change, I don't know that I'll accept this _immediately_. I just released a major version and avoided this (though it is a valuable change) so I may need to "save this for later."
> 
> That said, without doing a deep review - one of the things I've been trying to get to in the repo for a while is the separation of "tests through the public API" from "tests that require use of an internal API."
> 
> The "Autofac.Specification.Test" project is the "test through the public API" set of tests - these are the ones that we shouldn't change as we modify internals. "Autofac.Test" is continued testing that requires internal access. This is admittedly a work in progress - we started this separation before AI agents were really a thing and it would have caused a lot of churn. It may be something I'll look at finishing later. Anyway - we do have a decorator testing fixture [here](https://github.com/autofac/Autofac/blob/develop/test/Autofac.Specification.Test/Features/DecoratorTests.cs) and I think the tests here that verify how one might access pipeline configurations and such through this new feature would go in there rather than Autofac.Test, ideally reusing types and things in there.

No rush at all — please take whatever time you need on the review, and I completely understand if this has to wait for the next major.

Thanks for the pointer on test placement. You're right, and I'd got it wrong: nothing in that file actually needed internal access — Service, IComponentRegistration, DecoratorService, IResolvePipelineBuilder, PipelinePhase and ResolveRequestContext are all public — so I've moved the whole thing into Autofac.Specification.Test/Features/DecoratorTests.cs and deleted the Autofac.Test file. The tests now reuse ImplementorA, DecoratorA and DecoratorB instead of redeclaring near-identical copies, and I dropped five that duplicated coverage already in that fixture.

Any other suggestions are very welcome — happy to keep iterating.

Issue #1500, opened 2026-09-16
Concurrent first resolution of two services of one multi-service component can duplicate the component or miss an override


tillig:
## Describe the Bug

When one component exposes several services and two of those services are resolved for the **first time** on different threads at once, the two resolutions can each query the registration source independently. Because a source mints a new component per query, the result is either two distinct components for what should be one shared component, or the wrong default for one of the services.

This is long-standing, not a regression. It is being filed as a follow-up to #1465 / #1499, which fixed the single-threaded ordering defect and removed most of the exceptions this race used to throw, but did not close the race itself.

## Why It Happens

Service info initialization is serialized per service — `GetInitializedServiceInfo` takes `Monitor.Enter(info)` for the service being resolved. But when a source produces a component exposing several services, the tracker has to reach into a *different* service's info to record that the component belongs to it too. That second info is not covered by the monitor the thread holds, so the handoff between "one thread records the component for the other service" and "the other thread drains that source" is not atomic. Miss it, and the second thread queries the source itself and mints a second component.

The obvious fix — one tracker-wide lock around source draining — does not work. Sources call the `registrationAccessor` they are handed, which is `ServiceRegistrationsFor` → `GetInitializedServiceInfo` → `Monitor.Enter(otherInfo)`. A gate held across draining would therefore want an info monitor while another thread holding an info monitor wants the gate, which is the same lock-order cycle in a new place. Closing this properly needs either a deterministic acquisition order across service infos, or a single gate hoisted above every `Monitor.Enter(info)` so that all service info initialization serializes. Both change how registration concurrency works in general and would need their own performance validation.

## Steps to Reproduce

```csharp
interface IA<T> { }
interface IB<T> { }
class AB<T> : IA<T>, IB<T> { }

var builder = new ContainerBuilder();
builder
    .RegisterGeneric(typeof(AB<>))
    .As(typeof(IA<>))
    .As(typeof(IB<>))
    .SingleInstance();

using var container = builder.Build();

// First resolution of each service, concurrently.
var first = Task.Run(() => (object)container.Resolve<IA<int>>());
var second = Task.Run(() => (object)container.Resolve<IB<int>>());
var resolved = await Task.WhenAll(first, second);

// Intermittently fails: SingleInstance is expected to share one instance across both services.
Assert.Same(resolved[0], resolved[1]);
```

[...]

PR #1504, opened 2026-09-18
Do not build a held registration's pipeline early

### LINKED ISSUE #1503: Regression in autofac 9.3.3 resolution
## Describe the Bug

Upgrading to Autofac `9.3.3` causes `System.InvalidOperationException: Component pipeline has already been built, and cannot be modified.` when resolving an open generic component registered with multiple services (e.g. `.As(typeof(IFirstService<>)).As(typeof(ISecondService<>))`) if `Autofac.
  Extensions.DependencyInjection` (or any listener that attaches to `PipelineBuilding` via `ComponentRegistryBuilder.Registered`) is used.

I believe the regressions source is in commit [`6d74fde`](https://github.com/autofac/Autofac/commit/6d74fde113aed93ec0b0118561f5fc94dd22bdd1) (PR #1499 / issue #1465). The pipeline is pre-built
 ```csharp
    if (additionalInfo.IsSourceQueued(source))
    {
        // Recording the implementation publishes it to whichever thread drains this source next,
        // and that thread resolves through it, so its pipeline has to be built first. Building
        // is idempotent, so the call AddRegistration makes later does nothing.
        if (_ephemeralServiceInfo is null)
        {
            provided.BuildResolvePipeline(this);
        }

        DeferSourceImplementation(additionalInfo, source, provided);
    }
```

and the assumption was that it is idempotent, however that doesn't seem to be the case since somewhere along the line ComponentRegistration.PipelineBuilding throws when IsPipelineBuilt == true. 
## Steps to Reproduce

The following unit test shows the regression
```c#
    [Fact]
    public void MultiServiceOpenGenericCanAttachToPipelineBuildingViaRegistryEvent()
    {
        builder
            .RegisterGeneric(typeof(BothServices<>))
            .As(typeof(IFirstService<>))
            .As(typeof(ISecondService<>));

        builder.ComponentRegistryBuilder.Registered += (sender, e) =>
        {
            e.ComponentRegistration.PipelineBuilding += (sender, pipeline) => { };
        };

        using var container = builder.Build();

        Assert.NotNull(container.Resolve<IFirstService<int>>());
        Assert.NotNull(container.Resolve<ISecondService<int>>());
    }
```

## Expected Behavior

Should resolve the services without throwing.

## Exception with Stack Trace

```text
System.InvalidOperationException: Component pipeline has already been built, and cannot be modified.
       at Autofac.Core.Registration.ComponentRegistration.add_PipelineBuilding(EventHandler`1 value)
       at Autofac.Extensions.DependencyInjection.AutofacRegistration.AddFromKeyedServiceParameterMiddleware(Object sender, ComponentRegisteredEventArgs
  e)
       at Autofac.Core.Registration.ComponentRegistryBuilder.OnRegistered(Object sender, IComponentRegistration e)
       at Autofac.Core.Registration.DefaultRegisteredServicesTracker.AddRegistration(IComponentRegistration registration, Boolean preserveDefaults,
  IRegistrationSource originatingSource)
       at Autofac.Core.Registration.DefaultRegisteredServicesTracker.PopulateServiceInfo(Service service, ServiceRegistrationInfo info, Boolean
  isScopeIsolatedService)
```

## Dependency Versions
Autofac: 9.3.3 (works on 9.3.2)

tillig:
Fixes #1503

A regression I introduced in #1499. `@hjalle`'s diagnosis is exactly right, including the line.

`AddRegistration` raises `Registered` **before** it builds the resolve pipeline, so that listeners can attach to `PipelineBuilding`:

```csharp
_registrations.Enqueue(registration);
var handler = Registered;
handler?.Invoke(this, registration);   // listeners attach PipelineBuilding here

if (originatedFromDynamicSource)
{
    registration.BuildResolvePipeline(this);
}
```

Holding an implementation for a second service built that pipeline before `AddRegistration` was even called, so by the time `Registered` fired the pipeline existed and `PipelineBuilding`'s add accessor threw. `Autofac.Extensions.DependencyInjection` attaches on every registration through `AddFromKeyedServiceParameterMiddleware`, so any ASP.NET Core app with a multi-service open generic registration hit this deterministically on first resolve. My "building is idempotent" note was true of `BuildResolvePipeline` itself and missed the ordering contract around it.

## Proposed Changes

- Stop building the pipeline when holding an implementation. The build at the end of the enclosing `AddRegistration` call happens before any service can apply what was held, which is what that service needs.
- Add the reporter's repro as a regression test.

## Why removing it is the right fix rather than reordering

The early build was guarding a concurrent window: another thread can drain the source and apply the held implementation between the hold and the pipeline build. That window is **not** new — 9.3.2 throws the same `Component pipeline has not yet been built`, because it also applied the registration to every service before building. So removing the early build restores 9.3.2's ordering exactly rather than introducing anything.

Measured over 4,000 concurrent first-resolve attempts per build:

| | 9.3.2 | 9.3.3 (regressed) | this fix |
| --- | --- | --- | --- |
| `PipelineBuilding` attach from `Registered` | ok | **throws** | **ok** |
| #1465 ordering, resolved sequentially | 0 / 200 | 200 / 200 | **200 / 200** |
| Exceptions per 4,000 attempts | 582 | 1 | **6** |
| of which `pipeline has not yet been built` | 2 | 1 | 6 |

So this trades a rare concurrent race that predates the hold mechanism — 6 in 4,000, and only when two services of one multi-service component are first resolved on different threads — for removing a deterministic failure that affects everyone using `Autofac.Extensions.DependencyInjection` with that registration shape. The #1465 fix stays intact, and total exceptions remain far below 9.3.2.

Closing that residual window properly means publishing the hold only after `AddRegistration` has raised `Registered` and built, which needs the skip decision threaded into `AddRegistration` separately from the hold record. That is a larger change than a regression fix should carry, and it belongs with #1500, which already tracks this race.

## Verification

893 + 504 tests pass on net8.0 and net10.0. Release build clean at zero warnings, `dotnet format` clean. Confirmed the new test fails without the change with the reporter's exact exception, and passes with it.


tillig:
Pushed `b89822aa`: comment style, plus a flaky test this fix exposed.

The comments now tag the issue (`// Issue #1503:`) rather than narrating it in prose, matching the rest of the suite, and lose the restated consequence and the impact note — the issue carries those.

More importantly, removing the early pipeline build made `ConcurrentFirstResolveOfBothServicesDoesNotThrowOrDeadlock` flaky at roughly one run in ten, because it asserted the resolve does not throw and losing this race can throw — the same ~6-in-4000 window the description covers. Asserting no-throw was overreach on my part; the test exists to guard against blocking on the other service's monitor. It now asserts only that both threads finish, tolerating the lost race that #1500 tracks, and is renamed to `ConcurrentFirstResolveOfBothServicesDoesNotDeadlock` to say what it actually checks.

Verified over 20 consecutive full-suite runs — 12 on net8.0, 8 on net10.0 — with no failures, against 1 in 10 before.

tillig:
Added the version bump to `9.3.4` in `04f982e3`, so this is tag-ready on merge rather than needing the usual follow-up commit on `develop`.

Confirmed the version flows through the build — `dotnet msbuild default.proj` resolves `/p:Version=9.3.4-local` locally, and `default.proj:5` is the only place the version is declared.

